PR #53, opened 2014-05-05
Support both log4net 1.2.10 (old key) and later (new key) for Castle.Services.Logging

kkozmic:
This changes the `Castle.Services.Logging.Log4netIntegration.dll` so it references two versions of log4net (yes, feel free to look at the code of how I did it, but Abandon hope all ye who enter here.

Bottom line is, this one assembly can support either old or new-key version of log4net (and possibly both at the same time if someone is insane enough to attempt that).

this requires importing the right namespace:
`using Castle.Services.Logging.Log4netIntegration;` for old-key version (so it's not a breaking change)
`using Castle.Services.Logging.Log4netIntegration.log4net2` for new-key version

Types in both namespaces are identical to make migration trivial.

Looking for some feedback, most of all - do we want to merge that? Does anyone foresee any potential problems with this change?


jonorossi:
I don't think there is enough of a push at the moment to upgrade to log4net newer than 1.2.10 since many people are moving to more modern structured logging like Serilog. I'm happy to revisit it later though. I also really don't think this is a good solution to the problem for our maintenance and because it doesn't work with Windsor's logging facility ([LoggingFacility.cs#L46](https://github.com/castleproject/Windsor/blob/master/src/Castle.Facilities.Logging/LoggingFacility.cs#L46)). It is just too much of a hack I'm not really comfortable with it.

The problem with log4net after 1.2.10 isn't just the strong name key but a breaking change in the API. The source code is compatible if you compile Castle Core against log4net, but the binaries are not compatible. 

The `XmlConfigurator::ConfigureAndWatch(FileInfo)` method changed signature between `log4net` 1.2.10 and 1.2.11. It now returns an `ICollection` even though Castle's `ExtendedLog4netFactory::.ctor` method doesn't use the return type. You'll get a `System.MissingMethodException` mixing log4net versions if you don't use the same version you compiled with, that is just how the CLR method binding works.

```
public static void ConfigureAndWatch(FileInfo configFile)
public static ICollection ConfigureAndWatch(FileInfo configFile)
```

@kkozmic I'm going to close this pull request without merging (it conflicts anyway) for now, and we can revisit it in the future. Reopen it if you disagree.

kkozmic:
+1

kkozmic:
I definitely agree that this solution while _clever_ isn't something I'd put in production (not only for the reasons you mentioned but also because maintaining it would be hell.

I really don't care now whether we stick to log4net vOld or just upgrade to vNew.

Frankly, if someone wants to keep using old log4net and we do upgrade to new one, they can keep referencing old Castle logging (and old log4net) via binding redirects. This *will* work since Castle is still signed with the same key so it's not really cutting people out (as long as they stick to the abstraction that is) if I'm not mistaken.

PR #73, opened 2014-12-16
Upgrading to log4net 2. Fixes #64

### LINKED ISSUE #64: Update log4net integration to latest
I think it's time to upgrade to latest log4net version (with the new key).

This is a breaking change, as because of the new key, the log4net libraries can't use assembly redirect.

But, since our assembly retains the same key, people can assembly redirect between our new and old versions (as long as we don't change our API, which we most likely won't)

DaveMiscampbell:
So I'm sure there'll be more things to change than this but thought it would be worth getting the ball rolling.


jonorossi:
Thanks for the effort Dave, however please take a look at #53.

DaveMiscampbell:
No worries - thanks for the reply!

Issue #77, opened 2015-01-13
ProxyGenerator cache does not take into account AdditionalAttributes


cmerat:
When generating Proxy types, the CacheKey does not take into account the AdditionalAttributes specified on the ProxyGenerationOptions which causes a cache hit to return a Proxy that does NOT have the AdditionalAttributes applied to it **if a previous Proxy class has already been generated for the same type**.

The solution to fix this is for the Equality/GetHashCode operation on ProxyGenerationOptions to take into consideration the AdditionalAttributes and cause a cache miss, forcing the generation of a new type that will have the correct attributes applied to it.

kkozmic:
Yup. This is a known... Unsupported scenario.

Why so you need this?

cmerat:
I'm using a mocking framework (https://github.com/FakeItEasy/FakeItEasy) that uses ProxyGenerator to generate its fake instances. I need to test logic based on Attributes that are applied to a type. My unit tests are failing because the first proxy type definition is sticking to all other test cases (since the AppDomain is not unloaded between unit tests). The fake class with custom attributes was the simplest way to test the algorithm.

kkozmic:
Can you post some of that code?

cmerat:
Even better, I can link to the issue that I've opened on their project: https://github.com/FakeItEasy/FakeItEasy/issues/436

It contains repro steps as well as my investigation into what is causing the issue.

kkozmic:
right.

@cmerat a workaround for time being would be to disable caching (override `ModuleScope.GetFromCache` to return `null`.

In the longer term, perhaps this would be an acceptable workaround for FakeItEasy too. I see this very much as a corner case and I'm not really convinced fixing it, and therefore taking a performance hit for all other use-cases is a worthwhile trade of.

kkozmic:
I guess it's worth noting this was also mentioned in #71

cmerat:
Would adding AdditionalProperties to the CacheKey really be that much of a performance hit? Edge case or not, the interface is essentially lying (GenerateProxy with the specified options will SOMETIMES not return a proxy class matching the options). In the end it's your project and I understand the need to decide where resources are allocated. If you're open to contributions I'm willing to adjust GetHashCode and Equals to take into consideration AdditionalProperties.

kkozmic:
Go for it

adamralph:
Great - we'll hold off on the FakeItEasy workaround then. From our POV it would definitely be better for this to be fixed in Castle.

[...]

PR #78, opened 2015-01-16
Altered Equality on ProxyGenerationOptions to consider attributes

cmerat:
Changed Equals and GetHashCode to take into consideration
AdditionalAttributes. Order of contents is not important so long as
contents match.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+indent_style = tab
--- a/src/Castle.Core.Tests/ProxyGenerationOptionsTestCase.cs
+++ b/src/Castle.Core.Tests/ProxyGenerationOptionsTestCase.cs
@@ -16,12 +16,15 @@ namespace Castle.DynamicProxy.Tests
 {
 	using System;
 	using System.Collections.Generic;
+	using System.Reflection.Emit;
 
 	using Castle.DynamicProxy.Tests.Mixins;
 
 	using CastleTests;
 
 	using NUnit.Framework;
+
+	using DescriptionAttribute = System.ComponentModel.DescriptionAttribute;
 
 	[TestFixture]
 	public class ProxyGenerationOptionsTestCase
@@ -193,6 +196,58 @@ namespace Castle.DynamicProxy.Tests
 		}
 
 		[Test]
+		public void Equals_DifferentAdditionalAttributes()
+		{
+			var builder1 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test1" });
+			var builder2 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test2" });
+
+			_options1.AdditionalAttributes.Add(builder1);
+			_options2.AdditionalAttributes.Add(builder2);
+
+			Assert.AreNotEqual(_options1, _options2);
+		}
+
+		[Test]
+		public void Equals_SameAdditionalAttributes()
+		{
+			var builder = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test" });
+			_options1.AdditionalAttributes.Add(builder);
+			_options2.AdditionalAttributes.Add(builder);
+
+			Assert.AreEqual(_options1, _options2);
+		}
+
+		[Test]
+		public void Equals_SameAdditionalAttributesDifferentOrder()
+		{
+			var builder1 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test1" });
+			var builder2 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test2" });
+
+			_options1.AdditionalAttributes.Add(builder1);
+			_options1.AdditionalAttributes.Add(builder2);
+
+			_options2.AdditionalAttributes.Add(builder2);
+			_options2.AdditionalAttributes.Add(builder1);
+
+			Assert.AreEqual(_options1, _options2);
+		}
+
+		[Test]
+		public void Equals_DifferentAdditionalAttributesDuplicateEntries()
+		{
+			var builder1 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test1" });
+			var builder2 = new CustomAttributeBuilder(typeof(DescriptionAttribute).GetConstructor(new[] { typeof(string) }), new object[] { "Test2" });
+
+			_options1.AdditionalAttributes.Add(builder1);
+			_options1.AdditionalAttributes.Add(builder1);
+
+			_options2.AdditionalAttributes.Add(builder1);
+			_options2.AdditionalAttributes.Add(builder2);
+
+			Assert.AreNotEqual(_options1, _options2);
+		}
+
+		[Test]
 		public void GetHashCode_EmptyOptions()
 		{
 			Assert.AreEqual(_options1.GetHashCode(), _options2.GetHashCode());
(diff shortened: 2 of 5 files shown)


cmerat:
For issue https://github.com/castleproject/Core/issues/77

I've only implemented for AdditionalAttributes since AttributesToAddToGeneratedTypes is marked Obsolete.

Order of attributes does not matter (since it will not matter on the generated type either).

src/Castle.Core.Tests/ProxyGenerationOptionsTestCase.cs
```diff
@@ -23,7 +24,9 @@ namespace Castle.DynamicProxy.Tests
 
 	using NUnit.Framework;
 
-	[TestFixture]
+	using DescriptionAttribute = System.ComponentModel.DescriptionAttribute;
+
+    [TestFixture]
```
blairconrad: @cmerat, there's an indentation problem here. Castle.Core uses tabs for indentation, not spaces.

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -204,7 +208,53 @@ public override int GetHashCode()
...
+        }
+
+        private static bool ListsEquivalent<T>(IList<T> listA, IList<T> listB)
+        {
+            if (listA.Count != listB.Count)
+            {
+                return false;
+            }
+
+            // order is not important, just make sure that each entry in A is also found in B
```
blairconrad: I worry about this implementation. If we're guaranteed to have only one instance in each list that's equal to one instance in the other list, we're okay, but if duplicates are allowed (and I'll say right now that I don't follow well enough to know if duplicates are possible or not), then the lists `<a, a, b>` and `<a, a, b>` will compare as equivalent, even though they are not. Perhaps more damning, the hash codes would be _not_ be equal.

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -204,7 +208,53 @@ public override int GetHashCode()
...
 		}
+
+        private static int GetContentsHashCode<T>(IList<T> list)
+        {
+            var result = 0;
+            for (var i = 0; i < list.Count; i++)
+            {
+                if (list[i] != null)
+                {
+                    // simply XOR since order does not matter
```
blairconrad: crafty!

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
cmerat: You are right, in fact if the same attribute is there an even number of times, it will effectively be XOR'd out which would result in the HashCode not taking it into account at all.

I'll adjust the algorithm to group equal HashCodes together and alter the "group hash" which will be based on the number of equal elements. This way, two types that have the same repeated attributes would only have the same hash if they had the same number (order would still not matter).

How would you feel about comparing HashCodes in Equals before doing comparisons for each object individually? If the HashCodes match, we can inspect each. If not, then we won't iterate through the array.

blairconrad: I'm less worried about `GetHashCode` not taking the value into account than the problem with `Equals`. If `GetHashCode` says more things match than we want, that's inconvenient, but not disastrous (although the algorithm could perhaps be improved by adding hash codes of the elements, rather than `xor`ing, if desired).

Sorry, but I don't follow what you propose for altering the `GetHashCode`.

As far as having `Equals` first compare hash codes, that _could_ work, but really only provides benefits if the hash code is very quick to compute and if we expect to get a lot of equality mismatches.

cmerat: An option then would be to copy the destination items to a local array and set the value to `null` once it has already been "found" with another. This would force `Equals` to have the same number of "same" objects in both lists. That as well as using `add` instead of `xor` in `GetHashCode` should satisfy the cases that you have brought up. I'll try to get the changes in later today.

blairconrad: Thanks. Then again, keep in mind that I'm just butting into the PR, and maybe the Castle guys have competing views, and I'm sure they have extra experience and opinions about their codebase.

cmerat: Your points are still valid. I'm more comfortable with these changes than my initial submission.

adamralph: FWIW, XOR'ing to generate hash codes doesn't generally give a good hash code distribution, which is important when they are used to bucket values for a dictionary, for example. The method I usually go with is described in http://stackoverflow.com/a/263416/49241. Note the info regarding XOR in that answer.

cmerat: Interesting post, I tend to use XOR'ing for hash code generation because I assumed that it would generate a good distribution. Apparently I was mistaken!

src/Castle.Core.Tests/ProxyGenerationOptionsTestCase.cs
kkozmic: :+1: please make sure you follow the formatting conventions of the codebase

cmerat: Yep, will do. Did not notice it was tabs. I'll figure out how to set my VS without changing my global settings (all of my other projects are spaces :P)

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -204,7 +208,53 @@ public override int GetHashCode()
 			result = 29*result + (Selector != null ? 1 : 0);
 			result = 29*result + MixinData.GetHashCode();
 			result = 29*result + (BaseTypeForInterfaceProxy != null ? BaseTypeForInterfaceProxy.GetHashCode() : 0);
+		    result = 29*result + GetContentsHashCode(AdditionalAttributes);
 			return result;
 		}
+
+        private static int GetContentsHashCode<T>(IList<T> list)
```
kkozmic: this might be better suited for a separate static method in a helper class, same as `ListsEquivalent`

cmerat: Are there any classes/namespaces where you normally put utility methods like this? Any suggestions?

kkozmic: https://github.com/castleproject/Core/blob/master/src/Castle.Core/Core/Internal/CollectionExtensions.cs looks like it might be a reasonable place to put it, although I wouldn't make them extension methods probably.

src/Castle.Core.Tests/ProxyGenerationOptionsTestCase.cs
adamralph: One good way to do this is to use http://editorconfig.org/

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
kkozmic: Very interesting and insightful discussion. I already learned a bunch of interesting stuff just by reading through it :)

Keep in mind the primary use of this is for https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/ModuleScope.cs#L53 so we definitely care about good distribution of hashes. Also the idea mentioned a few comments above about comparing hash in `Equals` makes little sense since the dictionary will already have called `GetHashCode` before falling back to `Equals`

kkozmic: Also, please add tests for all the scenarios discussed to prove the behaviour

cmerat:
All suggested code changes have been committed (and I've fixed the indentation issues, thanks @adamralph for the suggestion, .editorconfig works wonders!)

I've also taken the liberty of adding the .editorconfig file to the solution. For anyone else who contributes and uses different settings, this mistake should no longer occur if they have the EditorConfig plugin installed (quite painless).

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
```
blairconrad: a quibble: line 8 specifies crlf, which is Windows-style line endings, not Unix

adamralph:
Should the two new methods in `CollectionExtensions` be extension methods?

cmerat:
@adamralph as per @kkozmic's comment, I've not made them extension methods. He thought they would make more sense as regular static methods.

.editorconfig
jonorossi: Personally, I wouldn't specify `crlf` or `lf` for `end_of_line` and just leave that to the editor to autodetect from what git has checked out. However, we don't have a `.gitattributes` set up properly on this repository yet.

cmerat: Removed, now only specifies indent_style.

cmerat:
@kkozmic Is there anything else missing to push this pull request through?

kkozmic:
I'm on holidays. I'll have a look when I'm back unless someone else beats
me to it.

cmerat:
@kkozmic I hope you had a nice vacation! Have you had any time to look at this pull request?

adamralph:
:+1:

kkozmic:
yeah, I'm back. I'll look over it this week. Thanks for your patience. @jonorossi any feedback from you mate?

kkozmic:
Merged. Thanks for the effort!

adamralph:
Awesome! I take it this will go out with 3.4.0?

jonorossi:
@kkozmic sorry for not getting to have a look at this today. Just got home and was actually thinking I need to look at this on the drive 20mins ago.

Looks good to me, there is probably a micro-optimisation we could do at some point to avoid the list copy, but there shouldn't be many attributes anyway.

jonorossi:
The Silverlight 5 build broke with this change:
http://builds.castleproject.org/viewLog.html?buildId=10603&buildTypeId=bt134&tab=buildResultsDiv

kkozmic:
ah yes I see. Should be an easy fix though. Do you mind doing it? I've
blown away my Windows VM so I don't have the code on my machine at the
moment.

cheers
@k

jonorossi:
Builds are now all green. I'm happy to release a 3.3.4 or 3.4.0. I doubt we are waiting on anything in particular as unfortunately without pull requests @kkozmic and I are pretty busy at the moment.

Issue #87, opened 2015-04-15
Error creating an interface proxy in case the interface contains a method with an optional decimal parameter


fknx:
Hello,

I have found an issue in Castle.Core 3.3.3 with the DynamicProxy generation. In case I want to create an interface proxy for an interface with an **optional** decimal parameter, the proxy generation fails with an 'System.Decimal is not a supported constant type' exception. The following program can be used to reproduce the issue:

```csharp
class Program
{
  static void Main(string[] args)
  {
	 var proxyGenerator = new ProxyGenerator();
	 proxyGenerator.CreateInterfaceProxyWithoutTarget<IMyClass>(new StandardInterceptor());
  }
}

public interface IMyClass
{
  void Do(decimal d = 0m);
}
```

However, the exception only occurs with the Castle.Core build targeting the **.NET Framework 4.5**. The .NET Framework 4.0 build works fine. (I have not tested the 3.5 or Silverlight versions).

Here you can find the full stacktrace:
```
System.Reflection.Emit.TypeBuilder.SetConstantValue(ModuleBuilder module, Int32 tk, Type destType, Object value)
Castle.DynamicProxy.Generators.Emitters.MethodEmitter.DefineParameters(ParameterInfo[] parameters)
Castle.DynamicProxy.Generators.Emitters.MethodEmitter..ctor(AbstractTypeEmitter owner, String name, MethodAttributes attributes, MethodInfo methodToUseAsATemplate)
Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateMethod(String name, MethodAttributes attributes, MethodInfo methodToUseAsATemplate)
Castle.DynamicProxy.Generators.MethodGenerator.Generate(ClassEmitter class, ProxyGenerationOptions options, INamingScope namingScope)
Castle.DynamicProxy.Contributors.CompositeTypeContributor.ImplementMethod(MetaMethod method, ClassEmitter class, ProxyGenerationOptions options, OverrideMethodDelegate overrideMethod)
Castle.DynamicProxy.Contributors.CompositeTypeContributor.Generate(ClassEmitter class, ProxyGenerationOptions options)
Castle.DynamicProxy.Generators.InterfaceProxyWithoutTargetGenerator.GenerateType(String typeName, Type proxyTargetType, Type[] interfaces, INamingScope namingScope)
[...]

PR #91, opened 2015-06-09
Fix for issue #87

fknx:
I have added an `if` statement which checks whether the parameter type is a `decimal` or `decimal?` before `SetConstant`is called. I also added two simple test cases where an `Interface` and `Class` proxy is created for an `Interface` (or `Class`) containing methods with all kinds of optional parameters.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/MethodEmitter.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/MethodEmitter.cs
@@ -159,6 +159,18 @@ namespace Castle.DynamicProxy.Generators.Emitters
 #if DOTNET45
 				if (parameter.HasDefaultValue && parameter.DefaultValue != null)
 				{
+					if (parameter.ParameterType == typeof(decimal) || parameter.ParameterType == typeof(decimal?))
+					{
+						/*
+						 * A decimal value may not be passed to SetConstant(), so omit the call
+						 * in this instance.
+						 * 
+						 * See https://github.com/castleproject/Core/issues/87 and https://bit.ly/1eXx6fF 
+						 * for additional information.
+						 */
+						continue;
+					}
+					
 					parameterBuilder.SetConstant(parameter.DefaultValue);
 				}
 #endif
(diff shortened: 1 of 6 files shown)


src/Castle.Core/DynamicProxy/Generators/Emitters/MethodEmitter.cs
```diff
@@ -159,6 +159,18 @@ private void DefineParameters(ParameterInfo[] parameters)
 #if DOTNET45
 				if (parameter.HasDefaultValue && parameter.DefaultValue != null)
 				{
+					if (parameter.ParameterType == typeof(decimal) || parameter.ParameterType == typeof(decimal?))
+					{
+						/*
+						 * A decimal value may not be passed to SetConstant(), so omit the call
+						 * in this instance.
+						 * 
+						 * See https://github.com/castleproject/Core/issues/87 and https://bit.ly/1eXx6fF
```
jonorossi: I know the background on the issue, but it'd be nicer if the comment was a bit clearer that this is a defect/omission in .NET so in the future when the .NET defect is fixed someone can remove this code without worry in case our GitHub issues history isn't around anymore.

Could you also include the full URL for the MSDN link rather than a bit.ly link:
https://msdn.microsoft.com/en-au/library/system.reflection.emit.parameterbuilder.setconstant(v=vs.110).aspx

jonorossi:
Looks good, thanks. Did you try compiling this with .NET 3.5 because I suspect it'll fail because optional parameters weren't supported then?

fknx:
I was able to compile the  code using Visual Studio even with the NET35 configurations, but according to [this](http://stackoverflow.com/questions/1210679/can-you-use-optional-parameters-in-code-targeting-net-3-5) link this may not work using a command-line compiler, so I have wrapped the test code in an `#if !DOTNET35`.

Additionally, I modified the comment as you suggested.

jonorossi:
@fknx thanks

jonorossi:
The Silverlight builds failed in the same way the .NET 3.5 one would have, but don't worry I've fixed them in master.

PR #101, opened 2015-07-06
mono support

RalphBariz:
ToDo:
* test 4.0 and 4.5 build in a vs.net 2013 build environment

Done:
* modified projects for beeing able to build with mono
* removed .NET 3.5 support
* "treat wranings as errors" consequently enabled everywhere and excluded harmless warnings by pragmas
* replaced log4net, NLog and Serilog references with the respective NuGet packages
* checked all DynamicProxy tests with mono and updated the ignore attributes
* fixed ClassProxyWithMethodsWithOptionalParametersTestCase.CanCreateClassProxy
* added a bash script for executing testcases
* added a bash script for building using mdtool
* tested 4.5 build in a linux + mono 4.0.2 build environment


jonorossi:
@RalphBariz thanks for sending the changes as a PR instead of a patch file. Could you update the PR title and description to describe what you are doing here because it looks like this will break everything but Mono.

FYI, I purposely removed the Mono specific build configuration because there is no sane way for us to maintain it. Mono really needs to build with the .NET 4.5 configuration because we can't release separate binaries via NuGet, and don't want to.

RalphBariz:
* It was not meant to be pulled immedate, just wanted to create the discussion platform.
* The mono build configs I'll remove, but I need to take out "treat warnings as errors" flag from dotnet45 build configurations
* Do castle.core need to support .NET < 4.0? Otherwise I would remove also those configs and replace all dependencies with nuget packages
* For sure before pulling, I'll try out and fix the endresult inside a vs.net 2013 environment

RalphBariz:
I remove .NET 3.5 support. If you wish to get it back I'll recreate the two build configs.

jonorossi:
@RalphBariz I'm going to close this for now. There has been a lot of change since for Mono support, and this PR has line ending issues so I can't see if anything is still relevant. Please open an issue if you would like to discuss anything specific. We do have a green [master build](http://builds.castleproject.org/project.html?projectId=Core&tab=projectOverview) with Mono 4.0.2, obviously now a little outdated.

Issue #104, opened 2015-07-30
Cannot create proxy for interface with custom attribute having an enum array parameter


csharper2010:
SetUp : System.InvalidCastException : Mindestens ein Element im Quellarray konnte nicht in den Zielarraytyp umgewandelt werden.
   bei System.Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length, Boolean reliable)
   bei System.Array.CopyTo(Array array, Int32 index)
   bei Castle.DynamicProxy.Internal.AttributeUtil.ReadAttributeValue(CustomAttributeTypedArgument argument) in AttributeUtil.cs: line 97.
   bei Castle.DynamicProxy.Internal.AttributeUtil.GetCtorArguments(IList`1 constructorArguments) in AttributeUtil.cs: line 81.
   bei Castle.DynamicProxy.Internal.AttributeUtil.CreateBuilder(CustomAttributeData attribute) in AttributeUtil.cs: line 67.
   bei ...

kkozmic:
Hey, thanks for the PR.

There's some indention issue in some places. Also, could you make sure (add tests) it works for all enums, not just `int` based?

cheers
Krzysztof

jonorossi:
The pull request #105 was merged last year, this issue just didn't get closed at the time.

PR #105, opened 2015-07-30
Can create proxy for interface with custom attribute having an enum a…

csharper2010:
…rray parameter. Fix for #104

Creating a typed array in AttributeUtil.ReadAttributeValue is not needed and raises an exception when the array is of an enum type because the CustomAttributeTypedArgument.Value returns the Integer behind the enum value.

### DIFF
--- a/src/Castle.Core.Tests/CanDefineAdditionalCustomAttributes.cs
+++ b/src/Castle.Core.Tests/CanDefineAdditionalCustomAttributes.cs
@@ -76,16 +76,79 @@ namespace Castle.DynamicProxy.Tests
 		}
 	}
 
-	public enum SomeEnumForAttributeWithEnumArrayArgument
+	public enum SomeByteEnumForAttributeWithEnumArrayArgument : byte
 	{
 		Default,
 		Special
 	}
 
-	[AttributeUsage(AttributeTargets.All, Inherited = false)]
+	public enum SomeSbyteEnumForAttributeWithEnumArrayArgument : sbyte
+	{
+		Default,
+		Special
+	}
+
+	public enum SomeShortEnumForAttributeWithEnumArrayArgument : short
+	{
+		Default,
+		Special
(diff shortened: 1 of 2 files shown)


src/Castle.Core.Tests/CanDefineAdditionalCustomAttributes.cs
```diff
@@ -70,6 +76,20 @@ public AttributeWithTypeArrayArgument(params Type[] attributeTypes)
 		}
 	}
 
+    public enum SomeEnumForAttributeWithEnumArrayArgument
```
kkozmic-seek: that indention looks off

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -93,8 +93,8 @@ private static object ReadAttributeValue(CustomAttributeTypedArgument argument)
 			}
 			//special case for handling arrays in attributes
 			var arguments = GetArguments((IList<CustomAttributeTypedArgument>)value);
-			var array = Array.CreateInstance(argument.ArgumentType.GetElementType() ?? typeof(object), arguments.Length);
-			arguments.CopyTo(array, 0);
+            var array = new object[arguments.Length];
+            arguments.CopyTo(array, 0);
```
kkozmic-seek: indention is off

kkozmic:
Hey @csharper2010 thanks for fixing the indention.

Just to be sure it works in every case, it would be nice to have some extra tests covering enums based off of other types than int. The approved types for an enum are byte, sbyte, short, ushort, int, uint, long, or ulong so it would be great to have a test that covers each of them.

csharper2010:
I don't know if notifications work when adding commits... So I am adding this:

I am finished with adding attributes' constructor calls for alle the mentioned base types.

PR #113, opened 2015-10-06
Disable replicating UIPermissionAttribute

trivalik:
we could reach this by prevent replication of UIPermissionAttribute, which allow:
Substitute.ForPartsOf\<typeof(System.Windows.Forms.Control)\>();


jonorossi:
I wonder if we should change the line above the line you added instead. i.e. change ignoring `SecurityPermissionAttribute`s to all `CodeAccessSecurityAttribute`s, of which both of these classes inherit from, maybe even ignore all `SecurityAttribute`s.

trivalik:
In case you mean this:
```C#
		static AttributesToAvoidReplicating()
		{
			Add<ComImportAttribute>();
			Add<SecurityPermissionAttribute>();
#if DOTNET40
			Add<TypeIdentifierAttribute>();
#endif
        }
```

This would not fix the use of Substitute.ForPartsOf<typeof(System.Windows.Forms.Control)>();
this is verified by the test.
If this is not what you mean, please explain.

jonorossi:
@trivalik sorry for never replying here. Yes, I did mean something like that, and then to change the implementing code to use `IsAssignableFrom`.

If you still care about this could you please explain what this is actually fixing and why the `UIPermissionAttribute` can't be replicated for your use case.

trivalik:
`"Substitute.ForPartsOf()"` is part of library NSubstitute to create mockups for tests.
The problem is I want to create a Control with NSubstitute but it fails in your Core library.

I merged my branch, so it is now mergeable again, but:

I cannot build it anymore with VS2015 update 3. I get:

Failed to resolve all project references for 'Castle.Core.Tests'. The package restore result for 'C:\Users\triv\Source\Repos\Core\src\Castle.Core.Tests\Castle.Core.Tests.csproj' may be incomplete.
Restoring NuGet packages...
To prevent NuGet from restoring packages during build, open the Visual Studio Options dialog, click on the Package Manager node and uncheck 'Allow NuGet to download missing packages during build.'
Error occurred while restoring NuGet packages: Error reading 'C:\Users\triv\Source\Repos\Core\src\Castle.Core.Tests\project.json' at line 27 column 23 : TODO: unknown keyword platform
Unable to resolve Castle.Core  for .NETStandard,Version=v1.3.
NuGet Package restore failed for project Castle.Services.Logging.SerilogIntegration for 'Castle.Core '.
Unable to resolve Castle.Core  for .NETStandard,Version=v1.3.
NuGet Package restore failed for project Castle.Services.Logging.NLogIntegration for 'Castle.Core '.
NuGet package restore failed.
1>------ Build started: Project: Castle.Core, Configuration: NET40-Debug Any CPU ------
1>C:\Program Files (x86)\MSBuild\Microsoft\NuGet\Microsoft.NuGet.targets(140,5): error : Your project is not referencing the ".NETFramework,Version=v4.0,Profile=Client" framework. Add a reference to ".NETFramework,Version=v4.0,Profile=Client" in the "frameworks" section of your project.json, and then re-run NuGet restore.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

jonorossi:
@trivalik not sure, do you still have an old pre-release version of the .NET Core tools?

trivalik:
Ok, I could build it now from console. It is not possible to build with VS 2015 update 3.
Condition support from MS in project files is very bad.

warnings I got:

"C:\Users\triv\Source\Repos\Core\buildscripts\Build.proj" (RunAllTests target) (1) ->
"C:\Users\triv\Source\Repos\Core\Castle.Core.sln" (Build target) (2) ->
"C:\Users\triv\Source\Repos\Core\src\Castle.Core.Tests\Castle.Core.Tests.csproj" (default target) (4) ->
(ResolveAssemblyReferences target) ->
  C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "Serilog". Check to make sure the as
sembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\Users\triv\Source\Repos\Core\src\Castle.Core.Tests\Castle.Core.Tests.csproj]
  C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.Common.targets(1605,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "Serilog.Sinks.TextWriter". Check to
 make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\Users\triv\Source\Repos\Core\src\Castle.Core.Tests\Castle.Core.Tests.cspro
j]


"C:\Users\triv\Source\Repos\Core\buildscripts\Build.proj" (RunAllTests target) (1) ->
(_ExecNUnit target) ->
  [IL] : warning : [C:\Users\triv\Source\Repos\Core\build\NET40\NET40-Release\bin\CastleDynProxy2.dll : InvalidType::InvalidMethod][mdToken=0x6000001][offset 0x00000000] fall through end of the
method without returning [C:\Users\triv\Source\Repos\Core\buildscripts\Build.proj]
  C:\Users\triv\Source\Repos\Core\buildscripts\Build.proj(204,3): warning MSB3073: The command "C:\Users\triv\Source\Repos\Core\buildscripts/../tools/NUnit/nunit3-console.exe --noheader "C:
\Users\triv\Source\Repos\Core\build\NET40\NET40-Release\bin\Castle.Core.Tests.dll" --result=C:\Users\triv\Source\Repos\Core\buildscripts/../build/NET40/NET40-Release/bin/test-results/nunit-
results.xml;format=nunit2 --framework=v4.0 --x86" exited with code -1.

trivalik:
I used the old version 0f12b4163d8f07ea6451d9b9d9e78ba4e70d6c81 to repoduce and debug the issue. that I can provide you more information if needed.

The exception is shown on the picture.
fail

jonorossi:
@trivalik thanks for the detailed info, sorry for not coming back here after your last comment I thought you were still working on my comments from Feb 26 about changing the implementation to use `IsAssignableFrom`.

I now see why it has taken so long for this defect to be raised, Moq and NSubstitute ignore this attribute themselves, as well as some of the 42 other CAS attributes.

* https://github.com/moq/moq4/blob/8772a49d4565a03d8455685f5633ffed80bb2cc9/Source/Proxy/CastleProxyFactory.cs#L62-L71
* https://github.com/nsubstitute/NSubstitute/blob/f3608f0a55cabcc78c50f33466d867b537e19405/Source/NSubstitute/Proxies/CastleDynamicProxy/CastleDynamicProxyFactory.cs#L99-L110

To clear up that mess let's change the implementation to ignore all attributes that inherit from [`System.Security.Permissions.SecurityAttribute`](https://msdn.microsoft.com/en-us/library/system.security.permissions.codeaccesssecurityattribute(v=vs.110).aspx) as I suggested on Feb 26, otherwise we'll be endlessly adding yet another one of these CAS attributes.

You'll also need to wrap your `UIPermissionAttribute` unit test with `#if FEATURE_SECURITY_PERMISSIONS` otherwise it'll fail on .NET Core which doesn't have WinForms.

jonorossi:
Thanks @trivalik for the pull request, however I've just implemented a fix for this in #221 with my requested changes.

trivalik:
Fine, the only point which I miss is the aim of this change. That we can create a mock of "Control". Please think about adding my test, to make this sure for future.

jonorossi:
@trivalik we currently don't even reference `System.Windows.Forms` in our unit test project, seems a tad overkill to reference it since WinForms isn't going to change.

trivalik:
@jonorossi you don't have to reference it, dynamic load is enough for test
```C#
        [Test]
        public void DYNPROXY_99_ClassProxyFromControl()
        {
            Assembly assembly = Assembly.Load("System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089");
            var typeControl = assembly.GetType("System.Windows.Forms.Control");
            Assert.IsNotNull(typeControl);
            object control = generator.CreateClassProxy(typeControl);
            Assert.IsNotNull(control);
        }
```

jonorossi:
@trivalik I've added a belts and braces unit test for WinForms. Dynamically loading the assembly is still depending on it.

jonorossi:
Disabled this WinForms unit test on Mono so we don't have to install a X and a graphical environment.

PR #115, opened 2015-10-20
InvocationHelper performance fix

pecanw:
I have found that (with extensive usage of dynamic proxies) the readers-writers lock (implemented using SlimUpgradeableReadLockHolder / ReaderWriterLockSlim) on Castle.DynamicProxy.Internal.InvocationHelper.GetMethodOnType causes serious performance issues. 

As the Dictionary.TryGetValue is thread safe (according to the MSDN documentation) I propose to use a simple double-checked locking pattern. 

I have written a simple performance test which proved at least 5 times improvement.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
@@ -27,7 +27,7 @@ namespace Castle.DynamicProxy.Internal
 		private static readonly Dictionary<KeyValuePair<MethodInfo, Type>, MethodInfo> cache =
 			new Dictionary<KeyValuePair<MethodInfo, Type>, MethodInfo>();
 
-		private static readonly Lock @lock = Lock.Create();
+		private static readonly object @lock = new object();
 
 		public static MethodInfo GetMethodOnObject(object target, MethodInfo proxiedMethod)
 		{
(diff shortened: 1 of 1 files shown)


hammett:
@pecanw this is a great addition! But i dont see any test... Do you mind also including a MT test that ensures it wont break in concurrent access? I'm not sure Dictionary.TryGetValue is thread safe, the dictionary's bucket array may expand/contract. The concurrentdictionary is safe, but I dont think that's the one we use.

jonorossi:
I agree with @hammett, `Dictionary<K,V>.TryGetValue` is only thread safe if it isn't be modified while reading, calling `Add` on another thread can cause the reorganizing of hash buckets which shouldn't corrupt the dictionary because you're added a write lock, but will result in bad reads.

> A Dictionary<TKey, TValue> can support multiple readers concurrently, as long as the collection is not modified. Even so, enumerating through a collection is intrinsically not a thread-safe procedure. In the rare case where an enumeration contends with write accesses, the collection must be locked during the entire enumeration. To allow the collection to be accessed by multiple threads for reading and writing, you must implement your own synchronization.
> (https://msdn.microsoft.com/en-us/library/xfhwa508(v=vs.110).aspx)

I'm going to close this pull request as it currently stands, however there are other options that we should look at:
* The obvious one is to change the cache to a `ConcurrentDictionary<K,V>` which uses `Monitor` which many report is faster than `ReaderWriterLockSlim` and internally it will acquire a lock per hash bucket rather than serializing all writes. We still support .NET 3.5 so we'd need to keep the old code for it.
* If a plain critical section via `Monitor` turns out to be faster than `ReaderWriterLockSlim` which I don't imagine it will, we could switch the implementation of `SlimUpgradeableReadLockHolder` to use `Monitor` [like we used to for Silverlight](https://github.com/castleproject/Core/blob/v3.3.3/src/Castle.Core/Core/Internal/MonitorLockHolder.cs). This isn't the only cache in DynamicProxy that uses the `Lock` class, so we should be looking at this for all usages.
* Or go with a lock free cache and use `Interlocked.Exchange`.

PR #119, opened 2015-12-04
Port DictionaryAdapter and tests to .Net Core

jeremymeng:
DictionaryAdapter Xml feature and tests are excluded as they depend on Xml features that are not available on .Net Core.

- [X] All tests are passing. ~~Currently 1 test is failing.  They need further investigation to either fix or disable for .Net Core~~
    - [X] __Fixed__ ~~Some failed due to inaccessible members~~
    - [X] __Fixed__ ~~Tests for dictionary proxy with prefixed keys~~
    - [X] __Fixed__ ~~Resource tests failed. One is getting resource strings from .resx, the other is testing an embedded .txt file.~~
        - `AssemblyResourceFactoryTestCase.CreateWithAbsolutePath`
    - [X] __Disabled__ as it requires `PEVerify`. ~~`BaseTestCaseTestCase.TearDown_FindsVerificationErrors`~~
    - [X] __Disabled__ TraceLogger listener tests are configured in `App.Config`.  ~~Need to find out how to do that for .Net Core.~~
        - `TraceLoggerTests.FallUpToDefaultSource`
        - `TraceLoggerTests.FallUpToShorterSourceName`
        - `TraceLoggerTests.TracingErrorInformation`
        - `TraceLoggerTests.WritingToLoggerByType`

@jonorossi


jeremymeng:
@jonorossi I need some help from you on the test failures, especially the first two types of failures.  Thanks!

jeremymeng:
@jonorossi 

I tried a hacky way (https://github.com/jeremymeng/Core/commit/0b4687cc8bf754dff6dd736bd9c467f6ffece93f) to add trace listeners in the code since `System.Configuration` is not supported.  But it  does not match the expectation in `TraceLogger.cs`, specifically in `Initialize()` as it relies on configured trace sources: `var foundSource = new TraceSource("Default", defaultLevel);` at https://github.com/jeremymeng/Core/blob/0b4687cc8bf754dff6dd736bd9c467f6ffece93f/src/Castle.Core/Core/Logging/TraceLogger.cs#L153.  So I disabled the tests instead.

Now with only one test failure left.  I will check whether embedding .txt file is supported on .Net Core.  if you think it is not worth it I can just disable it.

Thanks!

README.md
```diff
@@ -42,13 +42,19 @@ Symbol                              | NET35              | NET40              |
 ----------------------------------- | ------------------ | ------------------ | ------------------ | ------------------
 `FEATURE_APPDOMAIN`                 | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
 `FEATURE_ASSEMBLYBUILDER_SAVE`      | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
+`FEATURE_BINDINGLIST`               | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
+`FEATURE_CONFIGURATION`             | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
```
jonorossi: There is already a `FEATURE_SYSTEM_CONFIGURATION` symbol.

README.md
```diff
@@ -68,17 +74,23 @@ Symbol                              | NET35              | NET40              |
...
 * `FEATURE_ASSEMBLYBUILDER_SAVE` - enabled support for saving the dynamically generated proxy assembly.
+* `FEATURE_BINDINGLIST` - enables support features that make use of System.ComponentModel.BindingList.
+* `FEATURE_CONFIGURATION` - enables System.Configuration features.
+* `FEATURE_DICTIONARYADAPTER_XML` - enable DictionaryAdapter Xml features.
 * `FEATURE_EMIT_CUSTOMMODIFIERS` - enables emitting optional and required custom modifiers defined on parameters including return parameters. It seems like a defect in corefx not to expose these methods because they are still implemented.
 * `FEATURE_EVENTLOG` - provides a diagnostics logger using the Windows Event Log.
 * `FEATURE_GAC` - enables support for obtaining assemblies using an assembly long form name.
 * `FEATURE_GET_REFERENCED_ASSEMBLIES` - enables code that takes advantage of System.Reflection.Assembly.GetReferencedAssemblies().
+* `FEATURE_ISUPPORTINITIALIZE` - enables support for features that make use of System.ComponentModel.ISupportInitialize.
+* `FEATURE_LISTSORT` - enables support for features that make use of System.ComponentModel.ListSortDescription.
```
jonorossi: `LE` comes before `LI`, it is right above in the table.

src/Castle.Core.Tests/BaseTestCaseTestCase.cs
```diff
@@ -50,7 +50,7 @@ public void TearDown_DoesNotSaveAnything_IfNoProxyGenerated()
 		}
 
 		[Test]
-#if SILVERLIGHT
+#if !FEATURE_ASSEMBLYBUILDER_SAVE
 		[Ignore("Cannot do in Silverlight")]
```
jonorossi: The ignore description is no longer correct, maybe just remove the description.

jonorossi:
@jeremymeng could you provide some detail about the missing XML features in .NET Core that DictionaryAdapter is using.

src/Castle.Core.Tests/Core.Tests/Resources/AssemblyResourceFactoryTestCase.cs
```diff
@@ -55,7 +55,12 @@ public void CanHandleBundleResource()
 		{
 			IResource resource =
 				new AssemblyBundleResource(
-					new CustomUri("assembly://" + AssemblyName + "/CastleTests.Core.Tests.Resources.MoreRes.TestRes/content1"));
+#if !FEATURE_NETCORE_RESOURCE
+					new CustomUri("assembly://" + AssemblyName + "/CastleTests.Core.Tests.Resources.MoreRes.TestRes/content1")
+#else
+					new CustomUri("assembly://" + AssemblyName + "/Castle.Core.Tests.Core.Tests.Resources.MoreRes.TestRes/content1")
```
jonorossi: I thought you could set the name of the resource in the project.json file?

src/Castle.Core.Tests/project.json
```diff
@@ -19,17 +22,17 @@
...
+    "define": [ "FEATURE_XUNITNET", "FEATURE_NETCORE_RESOURCE" ]
   },
   "resource": [
-    "Core.Tests/Resources/MoreRes/TestRes.resx"
+    "Core.Tests/Resources/MoreRes/TestRes.resx",
+    "Core.Tests/Resources/file1.txt"
   ],
   "exclude": [
-    "Components.DictionaryAdapter.Tests/**/*",
+    "AssemblyInfo.cs",
```
jonorossi: Excluding AssemblyInfo.cs means the generated file doesn't get included, we need a solution to that.

src/Castle.Core/Compatibility/NetCoreComponentModelExtensions.cs
```diff
@@ -12,11 +12,21 @@
...
 // limitations under the License.
 
-#if FEATURE_NETCORE_CONVERTER_API
+#if FEATURE_NETCORE_COMPONENTMODEL_API
 
-namespace System
+namespace System.ComponentModel
 {
-    public delegate TOutput Converter<TInput, TOutput>(TInput input);
+	public interface IDataErrorInfo
```
jonorossi: I'm not that comfortable publicly exposing this interface in Castle.Core especially under the System.ComponentModel namespace. We don't own it and nothing else on .NET Core is going to be able to use ours so I see no point in having it.

Outside of our unit tests there are only 3 usages of the interface, but possibly many more of the members. I'm not sure exactly what to do as I don't use DA. Do you know how much work it would be to conditionally compile out the interface usage but leave the implementation so consumers can call the properties directly?

My preference with .NET Core is to exclude any functionality we can't implementation rather than put more stop gap solutions into place, we can always expose more functionality later but much harder to remove hacky things like our own custom IDataErrorInfo interface.

Thoughts?

src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterFactory.cs
```diff
@@ -171,7 +175,8 @@ private object InternalGetAdapter(Type type, IDictionary dictionary, PropertyDes
 		}
 
 		#region Type Builders
-	
+
+#if FEATURE_APPDOMAIN
```
jonorossi: Since .NET 4.5 has `AssemblyBuilder.DefineDynamicAssembly` and `AssemblyBuilder.DefineDynamicModule` this should be `FEATURE_LEGACY_REFLECTION_API` instead so we can easily drop the old code once we no longer support pre-.NET4.5.

src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterFactory.cs
```diff
@@ -440,22 +468,21 @@ private static void PreparePropertyMethod(PropertyDescriptor descriptor, ILGener
 						.AddBehaviors(typeBehaviors.OfType<IDictionaryInitializer>    ().Cast<IDictionaryBehavior>());
 #endif
 
-			CollectProperties(type, property =>
+			CollectProperties(type, (property, reflectedType) =>
```
jonorossi: Did this break because I made changes to `InterfaceAttributeUtil`? Or something .NET Core related?

src/Castle.Core.Tests/TestDictionaryValidator.cs
```diff
@@ -47,11 +47,11 @@ public bool IsValid(IDictionaryAdapter dictionaryAdapter)
 		public string Validate(IDictionaryAdapter dictionaryAdapter)
 		{
 			List<String> errors = new List<string>();
-			var globalRules = AttributesUtil.GetTypeAttributes<IValidationRule>(dictionaryAdapter.Meta.Type);
+			var globalRules = AttributesUtil.GetTypeAttributes<ValidationRuleAttribute>(dictionaryAdapter.Meta.Type);
```
jonorossi: Why did these unit tests need to change? I know I made changes to AttributesUtil. I just want to make sure we haven't broken functionality.

src/Castle.Core/project.json
```diff
@@ -29,10 +32,9 @@
   },
   "compilationOptions": {
     "warningsAsErrors": true,
-    "define": [ "FEATURE_NETCORE_REFLECTION_API", "FEATURE_NETCORE_CONVERTER_API" ]
+    "define": [ "FEATURE_NETCORE_REFLECTION_API", "FEATURE_NETCORE_COMPONENTMODEL_API" ]
   },
   "exclude": [
-    "Components.DictionaryAdapter/**/*"
```
jonorossi: `exclude` can go now that it is empty.

jeremymeng:
@jonorossi Thanks for the feedback! I will look into it today or tomorrow.

src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterFactory.cs
jeremymeng: `PropertyInfo.ReflectedType` is gone from .Net Core.  The recommended change is to use the reference to the object being reflected to get this value.

src/Castle.Core.Tests/TestDictionaryValidator.cs
jeremymeng: Yes it's related to your change.  Before the change, the extension method 
```csharp
public static T[] GetAttributes<T>(this ICustomAttributeProvider member)
```
is on `ICustomAttributeProvider`, which is not available in .Net Core.  `ICustomAttributeProvider.GetCustomAttributes()` allows attributes that are not derived from `Attribute` (not CLS-Compliant).  Do you know whether this feature is needed?

After the change, the original test throws

```
Test Name:	DictionaryAdapterFactoryTestCase.CanValidateAndObtainDataErrorInformation
Test FullName:	Castle.Components.DictionaryAdapter.Tests.DictionaryAdapterFactoryTestCase.CanValidateAndObtainDataErrorInformation
Test Source:	E:\github\jeremymeng\Core\src\Castle.Core.Tests\Components.DictionaryAdapter.Tests\DictionaryAdapterFactoryTestCase.cs : line 1130
Test Outcome:	Failed
Test Duration:	0:00:00.028

Result StackTrace:	
at System.Attribute.GetCustomAttributes(MemberInfo element, Type type, Boolean inherit)
   at Castle.Core.Internal.AttributesUtil.GetAttributes[T](MemberInfo member) in E:\github\jeremymeng\Core\src\Castle.Core\Core\Internal\AttributesUtil.cs:line 69
   at Castle.Components.DictionaryAdapter.Tests.TestDictionaryValidator.Validate(IDictionaryAdapter dictionaryAdapter) in E:\github\jeremymeng\Core\src\Castle.Core.Tests\TestDictionaryValidator.cs:line 52
   at Castle.Components.DictionaryAdapter.Tests.TestDictionaryValidator.IsValid(IDictionaryAdapter dictionaryAdapter) in E:\github\jeremymeng\Core\src\Castle.Core.Tests\TestDictionaryValidator.cs:line 44
   at Castle.Components.DictionaryAdapter.DictionaryAdapterBase.<get_IsValid>b__87_0(IDictionaryValidator v) in E:\github\jeremymeng\Core\src\Castle.Core\Components.DictionaryAdapter\DictionaryAdapterBase.Validate.cs:line 33
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Castle.Components.DictionaryAdapter.DictionaryAdapterBase.get_IsValid() in E:\github\jeremymeng\Core\src\Castle.Core\Components.DictionaryAdapter\DictionaryAdapterBase.Validate.cs:line 33
   at Castle.Components.DictionaryAdapter.Tests.DictionaryAdapterFactoryTestCase.CanValidateAndObtainDataErrorInformation() in E:\github\jeremymeng\Core\src\Castle.Core.Tests\Components.DictionaryAdapter.Tests\DictionaryAdapterFactoryTestCase.cs:line 1134
Result Message:	Type passed in must be derived from System.Attribute or System.Attribute itself.
```

jeremymeng:
@jonorossi The change to conditionally compile on `IDataErrorInfo` isn't too hard (https://github.com/jeremymeng/Core/commit/a6ad98b5056e4f64176adfe888e62556ce036b8f).  After the change, a lot of the *Validate* types won't make sense any more on .Net Core, as they depend on `IDataErrorInfo`.  I don't know the usage of these any better than you do.

jeremymeng:
> @jeremymeng could you provide some detail about the missing XML features in .NET Core that DictionaryAdapter is using.

The major blocker here is the Xslt related stuffs (`XsltContext`)

src/Castle.Core.Tests/project.json
jeremymeng: > Excluding AssemblyInfo.cs means the generated file doesn't get included, we need a solution to that.

Is `[ClsCompliant(true)]` necessary? I think it caused build errors so I excluded it.

src/Castle.Core.Tests/Core.Tests/Resources/AssemblyResourceFactoryTestCase.cs
jeremymeng: > I thought you could set the name of the resource in the project.json file?

Yes, named resources are better here.  Fixed.

jeremymeng:
@jonorossi I've pushed the IDataErrorInfo change.  The Xslt dependency seems non-trivial to exclude.  Please let me know what you think.

src/Castle.Core.Tests/TestDictionaryValidator.cs
jonorossi: I can't think of a reason it would be needed, and since it is in the unit tests (not the library) I don't think there is any point spending more time trying to discover if it is.

src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterFactory.cs
jonorossi: Ah ReflectedType, yep I remember reading that is gone. Thanks.

jonorossi:
@jeremymeng thanks for getting the feedback addressed. I've just done another review pass and this is what is left before we can merge:

* Remove unneeded `FEATURE_NETCORE_RESOURCE` from README and project.json.
* Remove unneeded `FEATURE_NETCORE_COMPONENTMODEL_API` from README and project.json.
* Change exception message in `DictionaryValidateGroup.IsValid` to "IDataErrorInfo is not supported on this runtime." or something rather than mentioning .NET Core because other targets can use the same code in the future.
* We need to work out something for excluding AssemblyInfo.cs in project.json, we've always had CLS compliant defined. I'd be happy to merge the PR without it excluded (like it is in the netcore branch) with the branch not compiling and then we can work out what to do next.

jeremymeng:
@jonorossi sounds great!  I've pushed the suggested changes.

The ClsCompliant issue might be related to the newly introduced XUnit shim.

jonorossi:
Thanks, merged.

I think our next step should be getting the build going:
http://builds.castleproject.org/project.html?projectId=CoreNetCore&tab=projectOverview

Looking at the CLS compliant errors from it seems unrelated to xUnit.net, this is the first compiler warning:
> 'ProxyGenerator' cannot be marked as CLS-compliant because the assembly does not have a CLSCompliant attribute

Also the change you made to change the unit tests to using `ValidationRuleAttribute` rather than the interface has broken the .NET 3.5 unit tests, I haven't had a look at why that might be.

jeremymeng:
@jonorossi Thanks!

> Looking at the CLS compliant errors from it seems unrelated to xUnit.net, this is the first compiler warning:

This is a different issue than the one in the test project.  I believe that both AssemblyInfo.cs files (in Castle.Core and Castle.Core.Tests projects) are generated by the non-CoreClr build so in dnx only build they don't exist yet.  One possible solution is to check in them.



> Also the change you made to change the unit tests to using ValidationRuleAttribute rather than the interface has broken the .NET 3.5 unit tests, I haven't had a look at why that might be.

I will take a look tomorrow.

jonorossi:
@jeremymeng ah yes you are right, I remember realising that a few weeks back and started looking at how we'd run MSBuild to generate those files before running dnu. They can't be committed because they have the version number in them.

jeremymeng:
> They can't be committed because they have the version number in them.

If the files remain relatively stable (only change when version explicitly being bumped up), IMHO it is not that bad to check them in.

jonorossi:
The problem is they don't remain stable across build configurations, I just mentioned version because it was the easiest to explain. The assembly info [changes for each build configuration](https://github.com/castleproject/Core/blob/821047561ff957dac13e0e6d655395d0e0bae105/buildscripts/Castle.Common.Targets#L174-L184) because it changes description (the build configuration is in there so you know how the assembly you've got was built), and a bunch of other security attributes.

If we commit the files then the files cannot ever be generated otherwise it'll show up in Git with unstaged changes pending when you are working on a different build configuration. I'm happy to look at committing it with conditional compilation and we get rid of the MSBuild logic.

Since this issue is merged, lets start a new issue.

jeremymeng:
ah, I now see how the file is generated.

> I'm happy to look at committing it with conditional compilation (i.e. extending our CommonAssemblyInfo.cs)

I believe it is only the `[ClsCompliant(true)]` entry is needed to fix the build break.  Maybe just add that to CommonAssemblyInfo.cs?  Setting it to false would cause the same build errors anyway.

Issue #123, opened 2016-01-04
Failure in netcore build


jeremymeng:
http://builds.castleproject.org/project.html?projectId=CoreNetCore&tab=projectOverview

The issue is that `[ClsCompliant(true)]` is in the `AssemblyInfo.cs` which is generated by the MSBuild build.  In netcore build the file is missing.

PR #125, opened 2016-01-06
Move [ClsCompliant(...)] to CommonAssemblyInfo.cs

jeremymeng:
@jonorossi 

This fixes the failures in clean netcore build. (#123)


jonorossi:
1. When do we need `ASSEMBLY_CLSCOMPLIANT_FALSE`?
2. The unit test project also needs to use CommonAssemblyInfo.cs now too
3. What about the other fields in the assembly info file? Did you just want to do this one for now to get things building?

jeremymeng:
> When do we need ASSEMBLY_CLSCOMPLIANT_FALSE?

The original MSBuild logic allows setting `[ClsCompliant(false)]` so I made configurable too.  If there's no need I will remove it and just leave it always true.


> The unit test project also needs to use CommonAssemblyInfo.cs now too

`[ClrCompliant(true)]` would cause many build error for the unit test project.  I looked at the non-CoreClr test project and found that it doesn't seem to have either CommonAssemblyInfo.cs or AssemblyInfo.cs.  This probably explains why I only see these errors in dnx builds.

image


> What about the other fields in the assembly info file? Did you just want to do this one for now to get things building?

Yes.  I wanted to unblock the dnx build first.

jeremymeng:
@jonorossi The MSBuild-based unit test project does not compile *AssemblyInfo.cs".  Can you please suggest what to do for the other two bullet points?  Thanks!

jeremymeng:
@jonorossi friendly ping...

adamralph:
Is this the last thing that needs doing to complete #90?

jeremymeng:
This PR will allow building a NuGet package that supports .NET Core. There are other things pending, for example, to strongname sign or not, waiting for the dotnet/cli release in RC2, etc. However the main porting is done from my perspective.  @jonorossi might have more thoughts.

adamralph:
@jeremymeng that's great news. Thanks for the update.

bob-devereux-zocdoc:
Any progress on this?  Would love to keep using CastleCore and Moq on .Net core.

PR #132, opened 2016-02-26
upgrade to rc1-update1

rjdevereux:
I think you just beat me to this, but thought I'd send it in anyway to show my support.

### DIFF
--- a/buildscripts/Build.proj
+++ b/buildscripts/Build.proj
@@ -197,10 +197,6 @@ limitations under the License.
 
 	</Target>
 
-
-	<!-- Use TeamCity's when running on the build server -->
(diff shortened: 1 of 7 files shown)


jonorossi:
@rjdevereux yep, @jeremymeng beat you to it, however what is that `launchSettings.json` file for, we don't have one of those.

rjdevereux:
I'm not actually sure what created that file.  I thought it was from either running tests via VS or DNX when I ran the tests, but deleting it and running the tests both ways did not recreate it.

Also, my VS 2015 wasn't happy before unless I added ""projects": [ "src"]," to global.json.  But again removing it now and retesting and everything seems fine.

jonorossi:
Hopefully things don't continue to be finicky, glad to know you sorted out that other file. I'll closing this PR. Thanks for contributing even if it didn't get merged, feel free to jump on another issue I've logged.

Issue #133, opened 2016-02-26
.NET Core - remove AssemblyInfo generation


jonorossi:
In our MSBuild scripts we currently generate AssemblyInfo.cs files and they are ignored in Git. Let's get rid of the generation completely and bolster up CommonAssemblyInfo.cs with conditional compilation to do our bidding.

blairconrad:
I can take a peek at this.

blairconrad:
@jonorossi, I'm looking at the complicated (and so far to me, confusing) generation here, and beginning to have some questions. I'm not held up yet, but some answers may help me go in the right direction. Sorry to be a bother.

What's the goal here? I assume the problem is that we're not running the MsBuild on the NETCORE build, and we'd like a way to have the contents of the AssemblyInfo.cs available to it.

Is there anything else? You specifically against generated files (or generated `AssemblyInfo.cs`es) or is the idea just to get the proper information to the built assemblies?

One reason I ask is that I don't see a way to provide the version numbers without either generating a file to include. I just don't know how else to supply the version numbers from `Settings.csproj`, TeamCity's `Build_number`, or the `Configuration` (feeds into `AssemblyTitle`)

Followup questions:

1. Do you object to restoring _tracked_ project-specific `AssemblyInfo.cs` files if it turns out to be useful?
2. do you know why the SilverLight projects have a different name? This isn't super difficult to achieve. I was just curious.
3. many of the values put in the current `AssemblyInfo.cs` can be overridden at build time. I'm not sure if this is just by convention, or if it's essential. The TeamCity builds don't take advantage, as far as I can tell, except for supplying `Configuration`. For example, `ComVisible` could be supplied (usually `false`, or `AssemblyVersion_Major` (usually same as `Project_Major`). Do you want to keep this flexibility, or could we just hardcode the reasonable defaults?

jeremymeng:
@blairconrad I think SiverLight support is going to be dropped so you shouldn't need to worry about it any more.

Some more points for discussion.
- Is it necessary to have version number and build configuration in `AssemblyTitle` and `AssemblyDescription`? if not it would simplify the problem a little bit.

- Most of the assembly information can be achieved using `project.json` if they are static values so maybe the NETCORE project doesn't need `AssemblyInfo.cs`?  The version number needs to be changed every time in `project.json` when bumping the versions though. This is related to #134 if all projects are migrated to use `project.json`-based projects.

blairconrad:
[...]

PR #136, opened 2016-02-26
Enable signing.

### LINKED ISSUE #124: .NET Core - strong name assemblies
@jonorossi what do you think about having both signed and unsigned nupkgs so that people can choose what they want?

For illustration I added a signed project in https://github.com/castleproject/Core/compare/netcore...jeremymeng:signed?expand=1.  (This requires RC1 and later.)  The project would generate `Castle.Core.Signed.3.3.4.nupkg` with the following contents
```
Archive:  e:/github/jeremymeng/Core/src/Castle.Core.Signed/bin/Release/Castle.Core.Signed.3.3.4.nupkg
  Length     Date   Time    Name
 --------    ----   ----    ----
      286  01-04-16 11:34   _rels/.rels
     5273  01-04-16 11:34   Castle.Core.Signed.nuspec
   417280  01-04-16 11:34   lib/net35/Castle.Core.Signed.dll
   321708  01-04-16 11:34   lib/net35/Castle.Core.Signed.xml
   421376  01-04-16 11:34   lib/net40/Castle.Core.Signed.dll
   322077  01-04-16 11:34   lib/net40/Castle.Core.Signed.xml
   421888  01-04-16 11:34   lib/net45/Castle.Core.Signed.dll
   322077  01-04-16 11:34   lib/net45/Castle.Core.Signed.xml
   420352  01-04-16 11:34   lib/net46/Castle.Core.Signed.dll
   322077  01-04-16 11:34   lib/net46/Castle.Core.Signed.xml
   258560  01-04-16 11:34   lib/dotnet5.5/Castle.Core.Signed.dll
   298678  01-04-16 11:34   lib/dotnet5.5/Castle.Core.Signed.xml
      409  01-04-16 11:34   [Content_Types].xml
 --------                   -------
  3532041                   13 files
```

or name it `Castle.Core.StrongName.3.3.4.nupkg` if that sounds better.

jeremymeng:
As a result, several tests failed.  They are all accessing internals.  Is it expected since strong name signing test assemblies is not supported on .Net Core?

Fixes #124.


jonorossi:
I thought @Eilon said that the ASP.NET team also strong named their unit tests because they needed to be able to access the internals of the main assembly. This also works with strong names on the desktop runtime with both strong named.

jeremymeng:
I was referring to the strongname signing the dynamic assemblies. The disabled tests in these PR don't use dynamic assemblies, do they? I will look into it deeper.



On Fri, Feb 26, 2016 at 3:01 AM -0800, "Jonathon Rossi" <notifications@github.com<mailto:notifications@github.com>> wrote:


I thought @Eilon<https://github.com/Eilon> said that the ASP.NET team also strong named their unit tests because they needed to be able to access the internals of the main assembly. This also works with strong names on the desktop runtime with both strong named.

-

jeremymeng:
@jonorossi I verified it is indeed the DynamicProxyGenAssembly2. It is not strong name signed on .Net Core. For example, the following line failed, and you can see the public key token is null for the assembly.  If you agree, then this PR is good to merge.

image

jonorossi:
@jeremymeng Oh, didn't realise you meant `DynamicProxyGenAssembly2`, yep that is why `Castle.Core.dll` has an embedded resource ([DynProxy.snk](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/DynProxy.snk)) which we use to strong name the dynamic assembly if the assembly we are proxying is also strong named.

The project.json file doesn't specify the embedded resource, however the code isn't [throwing an exception to tell you that](https://github.com/castleproject/Core/blob/netcore/src/Castle.Core/DynamicProxy/ModuleScope.cs#L188) so must have conditionally compiled that strong naming code out because we didn't think we needed any strong name support at the time or maybe the BCL API is missing.

I've got to go out in a few minutes but take a look at the code that calls `ModuleScope.GetKeyPair` with the .NET 4.5 code and you should be able to find what we aren't calling anymore. If you can't find what's wrong just give me a yell and I'll take a look.

jeremymeng:
Thanks for the pointer @jonorossi.  I think I have seen that code before.  Sorry I didn't make it clearer, I was actually talking about strong name APIs (e.g., `StrongNameKeyPair`) are not supported in .NET Core (the real .NET Core, not Castle.Core .NET Core build :)), which is probably why the code was [conditional on `FEATURE_STRONGNAME`.](https://github.com/castleproject/Core/blob/netcore/src/Castle.Core/DynamicProxy/ModuleScope.cs#L387-L396)

jonorossi:
@jeremymeng I thought that might have been what you meant just didn't want to make the jump there without clarifying. I conditionally compiled all the strong name code out because at the time Microsoft's position was that strong names were dead for .NET Core and only corefx will have to deal with them. Do you know if there is a corefx/coreclr issue for exposing this API so we can strong name our dynamic assembly with Reflection.Emit? If this isn't going to be possible it's going to annoy a lot of people, I know heaps of people mock internal classes which they have no intention of exposing as public just to mock in a unit test.

jeremymeng:
@jonorossi I agree it is annoying without the support.  In fact the ASP.Net team was unable to run some of their tests with signed binaries due to this.  I am not aware of any issue for exposing StrongName APIs.  Could you please open one (issues from customers weight more)? I guess there's reasoning behind the removal. It would still be good to have a discussion and see whether there are any solutions.

jeremymeng:
Thanks to the help from @jkotas!  Although strong name signing APIs are not available on CoreCLR. For this particular scenario (InternalsVisibleTo dynamic RefEmit in-memory assembly), full strong name signing is not needed.  Setting public key is enough.

src/Castle.Core/Core/Internal/InternalsVisible.cs
```diff
@@ -21,21 +21,27 @@ public class InternalsVisible
...
 		///   Constant to use when making assembly internals visible to proxy types generated by DynamicProxy. Required when proxying internal types.
 		///   <c>[assembly: InternalsVisibleTo(CoreInternalsVisible.ToDynamicProxyGenAssembly2)]</c>
 		/// </summary>
 		public const string ToDynamicProxyGenAssembly2 =
-#if FEATURE_STRONGNAME
 			"DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7";
-#else
-			"DynamicProxyGenAssembly2";
-#endif
+
```
jeremymeng: Another way is to programmatically read the public key from the embedded DynProxy.snk.  However since the key remain relatively stable, and if it changes, these constants need to be updated too.  So I don't feel too much value in reading public key from the embedded snk file.

Any suggestion for the proper place for this field?  It is not too bad to have these public keys to stay together though.

jeremymeng:
Bummer.  Now tests are passing, but the SerilogIntegration now cause build failure with this PR because the beta Serilog binary is not strong name signed yet.

jonorossi:
Looks good to me, leave it hardcoded in a field as it'll be faster and the key is never going to change, not even sure why we changed it a decade ago for DP v2. If we are doing that then there isn't a lot of point keeping the embedded resource then?

If Serilog isn't actually strong named then does Serilog have a bug? Nick definitely made the change in https://github.com/serilog/serilog/pull/627 and our build server is using Serilog 2.0.0-beta-505 which was released 25th February.

jonorossi:
Also what is left using `FEATURE_STRONGNAME` as it no longer makes sense with your change in this PR.

jeremymeng:
> If we are doing that then there isn't a lot of point keeping the embedded resource then?

Right, if we don't read from the file.

> If Serilog isn't actually strong named then does Serilog have a bug?

Looks like it is indeed strong named. But somehow I got a build error last night complaining about public key=null.  It was late so I haven't looked further.  Will check again today.

jeremymeng:
> Also what is left using FEATURE_STRONGNAME as it no longer makes sense with your change in this PR.

Not needed anymore, unless we need to save strong named assemblies to disk.

jeremymeng:
I pushed a change to remove `FEATURE_STRONGNAME`.

jonorossi:
@jeremymeng looking great, `FEATURE_ASSEMBLYBUILDER_SAVE` is definitely the one to use around those tests which you've done. Everything is ready to merge except you've excluded `AssemblyInfo.cs` in Castle.Core.Tests, is there a reason for that?

jeremymeng:
> you've excluded AssemblyInfo.cs in Castle.Core.Tests, is there a reason for that?

Because including it would break tests which expect version of "0.0.0.0" ([one example](https://github.com/castleproject/Core/blob/master/src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs#L264)) as it override the version specified in project.json. The NET35/NET40/NET45 projects don't include AssemblyInfo.cs either.

jonorossi:
Great, we'll sort out AssemblyInfo in #133.

jonorossi:
Many thanks for getting that sorted, and to @jkotas for the assistance with getting it working on .NET Core.

Issue #137, opened 2016-02-27
Update Serilog library version to 2.0 and fix failing test


jeremymeng:
As the .NET Core port requires v2.0, it make sense to have the same version for Desktop. 

(I wonder why the dependencies are not managed by NuGet.)

There's [a breaking change](https://github.com/serilog/serilog/pull/571) in v2.0 

which would break one Serilog integration test: https://github.com/castleproject/Core/blob/v3.3.3/src/Castle.Core.Tests/SerilogIntegration/SerilogTests.cs#L163-L183

jonorossi:
@jeremymeng our vendored dependencies pre-date NuGet, and the early versions of NuGet sucked real bad so we didn't want to add that pain. I'd like to start the discussion on #134 which would turn that around and make everything more consistent to how the .NET Core stuff works.

jonorossi:
#134 has been closed because project.json won't survive, however we need to still look at upgrading the dependencies used by the .NET Framework builds.

jeremymeng:
The current build/project setup makes updating dependencies harder.  How about switching to NuGet dependencies?  Will probably need to split Castle.Core project into three projects: net35, net40, and net45.

jonorossi:
@jeremymeng yep I agree the current extended MSBuild set up makes it hard to integrate NuGet. We used to have `-vs2005`, `-vs2008` and `-vs2010` project files to account for the incompatible Visual Studio file formats, I don't think we want to go back there, it was very painful when you add just a single new file.

I think we should just wait until the new Visual Studio tooling comes which integrates .NET Core and retargeting support, then we'll only need a single place for dependencies to be defined. With that in mind let's continue with #188.

jonorossi:
Serilog has been upgraded, closing.

Issue #128, opened 2016-02-25
.NET Core - upgrade log4net, NLog, Serilog logging integration libraries


jonorossi:
These are still yet to be touched.

PR the `netcore` branch.

jeremymeng:
log4net porting PR: https://github.com/apache/log4net/pull/16

jonorossi:
Thanks for the links, I had been following those. As you probably saw Serilog 2.0.0 alphas have been out for a while now.

jeremymeng:
I am working on Serilog integration.

jonorossi:
FYI, NLog released alpha1 on the January 14: https://www.nuget.org/packages/NLog/4.4.0-alpha1.

jeremymeng:
Thanks @jonorossi I didn't know about that.  I gave it a try but it looks like they don't support `dotnet5.4` yet.  I will see what I can do.

jeremymeng:
I am on NLog integration.  There's a alpha-2 package that supports dotnet5.4 now.  However it failed the one and only NLog integration unit test. The same kind of failures are also seen in NLog unit tests (Dnx).

hikalkan:
I created a nuget package to integrate Castle.LoggingFacility package with Microsoft.Extensions.Logging package. See: https://github.com/volosoft/castle-logging-ms-adapter

jonorossi:
log4net, NLog and Serilog have been upgraded (#199) to the newest versions.

hikalkan:
Thank you for that. When can this be released?

jonorossi:
@hikalkan I'll have a beta release out in the next couple of days, just a few more things to sort out, RTM should hopefully follow fairly quicker afterwards.

hikalkan:
Thank you very much for informing. This was one of the most asked questions to me since my framework (https://github.com/aspnetboilerplate/aspnetboilerplate) depends on castle logging facility.

PR #138, opened 2016-02-27
Add Serilog integation project for .NET Core.

jeremymeng:
For part of #128.

One test is failing due to a Serilog v2 breaking change.  I disabled it and opened an issue #137.  There are [suggestions about improving Serilog integration](https://github.com/serilog/serilog/issues/561#issuecomment-189552136). I don't feel it belongs to this branch.

The PEVerify related change could have gone to another PR but since I need the newer version to run NET45 and NET40 tests using VS 2015 I just included it here.  I could do a separate PR if desired.

### DIFF
--- a/Castle.Core-NetCore.sln
+++ b/Castle.Core-NetCore.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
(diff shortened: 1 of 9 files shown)


jonorossi:
Looks good, I've created #142 to rework out SerilogFactory since it needs quite a bit of change.

PR #140, opened 2016-02-27
Add NETCORE build target

### LINKED ISSUE #130: .NET Core - unit test runner returning exit code 3
Build 23 failed:

> Tests passed: 266, ignored: 1; exit code 3 (new) 	
http://builds.castleproject.org/viewType.html?buildTypeId=CoreNetCore_MasterNetCore

### LINKED ISSUE #139: Create build target for NETCORE
From https://github.com/castleproject/Core/issues/130#issuecomment-189626632.

Will allow us to run

```cmd
build.cmd NETCORE
```

I made up the label NETCORE. Can easily be changed.

We'd also want to update the build configuration to call this target instead of its current 5 steps.

blairconrad:
Fixes #139.
Fixes #130. (Once we have the build server use `build.cmd NETCORE`.)

### DIFF
--- a/buildscripts/Build-NetCore.ps1
+++ b/buildscripts/Build-NetCore.ps1
@@ -1,0 +1,39 @@
+$frameworkVersion = "1.0.0-rc1-update1"
+$frameworkArchitecture = "x86"
+
+$dnxHome = Join-Path (Get-Location) ".dnx"
+$dnxPath = Join-Path $dnxHome "bin"
+
+Write-Host "Downloading dnvm to: $dnxPath"
+if (!(Test-Path $dnxPath)) { md $dnxPath | Out-Null }
+
+$dnvmPs1Path = Join-Path $dnxPath "dnvm.ps1"
+$dnvmCmdPath = Join-Path $dnxPath "dnvm.cmd"
+
+$wc = New-Object System.Net.WebClient
+$wc.DownloadFile('https://raw.githubusercontent.com/aspnet/Home/dev/dnvm.ps1', $dnvmPs1Path)
+$wc.DownloadFile('https://raw.githubusercontent.com/aspnet/Home/dev/dnvm.cmd', $dnvmCmdPath)
+
+Write-Host "Downloading dnx"
+
+$env:PATH = ("$dnxPath;" + $env:PATH)
+
+dnvm install $frameworkVersion -runtime coreclr -arch $frameworkArchitecture
+
+dnx --version
+
+Write-Host "Downloading packages"
+
+dnu restore
+
+Write-Host "Building"
+
+$env:Path = ((Join-Path $dnxHome "runtimes\dnx-coreclr-win-$frameworkArchitecture.$frameworkVersion\bin") + ";" + $env:Path)
+
+dnu build src/Castle.Core src/Castle.Core.Tests --configuration Release --out build/NETCORE
+
+Write-Host "Running tests"
+
+dnx -p src/Castle.Core.Tests test
+
+exit $LASTEXITCODE
--- a/buildscripts/Build-NetCore.ps1
+++ b/buildscripts/Build-NetCore.ps1
@@ -34,6 +34,7 @@ dnu build src/Castle.Core src/Castle.Core.Tests --configuration Release --out bu
 
 Write-Host "Running tests"
 
-dnx -p src/Castle.Core.Tests test
+cd src/Castle.Core.Tests
+dnx test
 
 exit $LASTEXITCODE
--- a/buildscripts/build.cmd
+++ b/buildscripts/build.cmd
@@ -19,6 +19,13 @@ IF NOT EXIST %~dp0..\Settings.proj GOTO msbuild_not_configured
 REM Set Framework version based on passed in parameter
 IF "%1" == "" goto no_nothing
 
+REM nesting the netcore build in an IF body prevents reporting of the error level
+IF /i "%1" NEQ "NETCORE" goto not_netcore
+
+PowerShell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Unrestricted -Command %~dp0Build-NetCore.ps1
+EXIT /B %ERRORLEVEL%
+
+:not_netcore
 IF /i "%1" == "NET40" (SET FrameworkVersion=v4.0)
 IF /i "%1" == "NET40" (SET BuildConfigKey=NET40)
 
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ artifacts/
 src/**/obj/
 src/**/bin/
 project.lock.json
+/.dnx/


buildscripts/Build-NetCore.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+Write-Host "Building"
+
+$env:Path = ((Join-Path $dnxHome "runtimes\dnx-coreclr-win-x64.1.0.0-rc1-update1\bin") + ";" + $env:Path)
+
+dnu build src/Castle.Core src/Castle.Core.Tests --configuration Release --out build/NETCORE
+
+Write-Host "Running tests"
+
+dnx -p src/Castle.Core.Tests test
```
blairconrad: I worry a little about this, as I'm not getting a summary like @jeremymeng suggests in https://github.com/castleproject/Core/issues/130#issuecomment-189781359, but that could just be because of whatever needs to be done for #130, which I intend to return to. 
Experts, is line 34 correct?

jeremymeng:
The resource factory test failure could be due to dnx bug, as a workaround for now, could we cd into the test project folder then run `dnx test`?

blairconrad:
Yeah, I was going to give that a try. I was off on another project for today...

blairconrad:
The tricky part is that I'm not getting those failures all the time, so it's a little tough to see whether the error not appearing is just luck.

**Although** I just gave it a quick try and did not see the errors show up. And dnx exits with error code 0 instead of 1. This may indeed be the workaround for #130. Nice job.

I'd be inclined to follow that line up there, but to keep the line as is here, since I consider this issue to be focused on creating a build script that mirrors (except for framework version, I suppose) what TeamCity is doing now.

buildscripts/Build-NetCore.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+$wc = New-Object System.Net.WebClient
+$wc.DownloadFile('https://raw.githubusercontent.com/aspnet/Home/dev/dnvm.ps1', $dnvmPs1Path)
+$wc.DownloadFile('https://raw.githubusercontent.com/aspnet/Home/dev/dnvm.cmd', $dnvmCmdPath)
+
+Write-Host "Downloading dnx"
+
+$env:PATH = ("$dnxPath;" + $env:PATH)
+
+dnvm upgrade -r coreclr
```
jonorossi: This is different to what I threw into TeamCity, if a new version gets released won't the new one get downloaded and the line below which has a hardcoded path will no longer work? I assume the new DNX doesn't work like beta8 where I just installed it into the same directory as dnvm.

buildscripts/Build-NetCore.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+Write-Host "Downloading packages"
+
+dnu restore
+
+Write-Host "Building"
+
+$env:Path = ((Join-Path $dnxHome "runtimes\dnx-coreclr-win-x64.1.0.0-rc1-update1\bin") + ";" + $env:Path)
+
+dnu build src/Castle.Core src/Castle.Core.Tests --configuration Release --out build/NETCORE
```
jonorossi: I know this is what I already had, but does DNX have a feature to build using the global.json file so we don't have to add each logging service here too?

jeremymeng:
Instead of `dnvm upgrade`, the version installed and used probably should match what's in `project.json`.  I am not sure about `global.json` (had thought that it was used by VS only).

jonorossi:
@jeremymeng the DNX version is definitely in global.json not project.json. Maybe `dnvm upgrade` reads global.json and we don't need to specify it because it gets the right version not the latest.

jeremymeng:
No, dnvm manages the dnx runtime installations for the current user. It does not interact with projects.  `global.json` [is for VS](http://docs.asp.net/en/latest/conceptual-overview/understanding-aspnet5-apps.html#the-global-json-file).

buildscripts/Build-NetCore.ps1
blairconrad: Of course. I will investigate.

blairconrad:
I've (at home - rushing to try it out over breakfast) an amendment that will suck the framework version out of `project.json` (using `Select-String`, so not particularly pretty) and will use the found value for downloading dnx and for building. It seems to work quite well.

kenegozi:
Nice trick. Not a PS expert here at all so grain of salt is required, but
there should be a ConvertFrom-Json that should let you do the lookup a
little bit easier.

blairconrad:
@kenegozi I bet it's not easier (of course, that depends on one's attitude toward regular expressions), but would be safer!

I'd considered it, but `ConvertFrom-Json` didn't show up until PowerShell 3.0, and I'm not sure how prevalent it is (fun fact: I don't have it at work), so I shied away, opting for the universal `Select-String`.

Happy to change if everyone thinks I'm being overly cautious.

kenegozi:
easier to maintain and reason with future errors (especially considering
how .net core's concepts are still a moving target)

I did not know it was a PS3 only. PS already limit builds to windows,
narrowing it further is not making it better.


Bottom line, Select-string is fine by me for this (perhaps with a good
error message pointing to "oh gosh maybe format changed go read XYZ"),




Ken Egozi.
http://kenegozi.com/ <http://kenegozi.com/blog>

jeremymeng:
Thinking about it more, for now it might be fine to just use hard-coded `1.0.0-rc1-update1`, since this is the last RC1 version.  In RC2 the runtimes will come in NuGet packages, with their versions specified in `project.json`.

blairconrad:
@kenegozi said

> perhaps with a good
error message pointing to "oh gosh maybe format changed go read XYZ"

I like the idea. Would work for unfound version, but it may not be possible to detect false positives…

@jeremymeng, given what you say about the runtime delivery mechanism changing, what you say makes sense. Hardcoding seems no worse, and when things stop working with RC2, it'll be more obvious what's wrong than having to decode where we're loading the version from.

jonorossi:
I didn't expect you'd actually read `1.0.0-rc1-update1` from global.json, I was only querying why it wasn't included in your copy of the PS script as I needed it when I did the original ones.

Happy for it to be hardcoded, I just don't want a new version being released and it automatically being feed into our next build. Eventually we'll likely need to drop PS anyway so we can test on our Linux build agent.

jonorossi:
FYI our build server is running PowerShell 3.0, however I do know a lot of people with Windows 7 machines only have 2.0.

blairconrad:
> I didn't expect you'd actually read 1.0.0-rc1-update1 from global.json, I was only querying why it wasn't included in your copy of the PS script

Oh. Sorry. You probably shouldn't leave dangerous phrases like

> the DNX version is definitely in global.json

lying around then. :stuck_out_tongue_winking_eye: 

hardcodedness coming up real soon now.

buildscripts/build.cmd
```diff
@@ -19,6 +19,13 @@ IF NOT EXIST %~dp0..\Settings.proj GOTO msbuild_not_configured
 REM Set Framework version based on passed in parameter
 IF "%1" == "" goto no_nothing
 
+REM nesting the netcore build in an IF body prevents reporting of the error level
```
blairconrad: … and if I told you how long it took me to figure _that_ out, you wouldn't believe me.

jonorossi:
Ah, this quote was directed at Jeremy as he said the DNX version was in project.json which it isn't, well at least in the version we are using.
> @jeremymeng the DNX version is definitely in global.json not project.json

So are we ready to merge with your last change?

I noticed you added `-ExecutionPolicy Unrestricted` to the PowerShell line, is that needed?

buildscripts/build.cmd
jonorossi: LOL, don't worry I'd believe you, batch files are so painful.

blairconrad:
> So are we ready to merge with your last change?

I think so.

> I noticed you added -ExecutionPolicy Unrestricted to the PowerShell line, is that needed?

Unclear. It came along with the samples I was reading on reporting error codes from PowerShell, back when I was going crazy with the IFs. It seemed harmless, so I left it. I'm happy to remove it.

Also, you may have noticed that I bundled two fixes in the one PR. Not sure what the policy is on that. Some projects are okay with it, but some aren't. They are separate commits, so I can easily separate them…

jonorossi:
Merged, both fixes are closely related so happy to take them in the same PR. I've updated the build configuration before merging.

blairconrad:
Cool.
Oh. You _disabled_ the build steps! That's clever. I wouldn't've known that was a thing!

blairconrad:
Great heavenly day! It seems to have worked!

jeremymeng:
> Jeremy as he said the DNX version was in project.json

oh, I didn't mean literally matches.  was more about `rc1-update1` <=> package version `-beta-23516`.

Thanks for the work @blairconrad.

blairconrad:
All: Thanks for doing most of the work! I stole your ideas and glued them together, and it seemed to work!

blairconrad:
@jeremymeng, a pleasure, but I did very little. 80% of my time was spent getting error codes out of the verdammt powershell script.

jonorossi:
@jeremymeng ah, it doesn't matter now but what package would you look take that from out of the project.json?

@blairconrad doesn't really matter what it builds on you are moving things along. Thanks.

jeremymeng:
> it doesn't matter now but what package would you look take that from out of the project.json?

I wasn't thinking about programmatically getting the runtime version from project.json.  I like the hardcoded versions as we have to explicitly change it when upgrading to a newer version. By matching I mean project.json has beta8 (23409) then build server should use beta8, project.json has rc1 (rc1-final or beta-23516) then build server should use rc1-update1.  Anyway there won't be any future dnx runtimes so it doesn't matter.

kenegozi:
The stuff that is being going on on this group in the past few days is an
insanely great OSS story. Thanks a million to everyone who is getting
actively involved.

Ken Egozi.
http://kenegozi.com/ <http://kenegozi.com/blog>

PR #144, opened 2016-02-29
Save and restore Console.Out and Console.Error.

### LINKED ISSUE #143: ConsoleLogger tests change Console.Out and Console.Error, but do not restore them.
As a result when `dnx test` we don't see any output after these tests are executed.

jeremymeng:
Save and restore Console.Out and Console.Error.

Fixes #143


jonorossi:
Thanks

Issue #146, opened 2016-03-03
NLogTests.Should_preserve_correct_callsite_information failed after upgrading to NLog v4.4.0-alpha3


jeremymeng:
Failing for both NET45 and .NET Core.  Looks like a breaking change?

```
CastleTests.NLogIntegration.NLogTests.Should_preserve_correct_callsite_information:
  Expected: String containing "NLogTests.WriteLogMessage"
  But was:  "14:03:39 MyLoggerName Castle.Services.Logging.NLogIntegration.NLogLogger.Debug something"
```

jonorossi:
That is the only NLog unit test so we need to get it fixed. Maybe the CoreCLR doesn't respect `[MethodImpl(MethodImplOptions.NoInlining)]`?

jeremymeng:
My debugging session showed that [NLog skip the first stack, then skip frames from NLog assembly](https://github.com/NLog/NLog/blob/d87bc70f176356c6f5bcb48f598e1b67fb7f0b61/src/NLog/LoggerImpl.cs#L121).  The second frame is indeed NLogLogger.Debug().  

Looks like the relevant code was [just rewritten](https://github.com/NLog/NLog/commit/086f09b86f22fa86a4cc54cdc337a4a1c387cdd2).  I will run a quick test to see if the test passes using NLog prior to the rewrite.

jeremymeng:
Looks like it is a regression in NLog.4.3.0-beta1.  The test passed with NLog.4.3.0-alpha3.

Add @304NotModified. Is this a bug in NLog?  The test is https://github.com/castleproject/Core/blob/netcore/src/Castle.Core.Tests/NLogIntegration/NLogTests.cs.

304NotModified:
Thanks for testing /reporting. This has been rewritten indeed, so it looks like a bug.

jonorossi:
@jeremymeng thanks for tracking it down.

304NotModified:
This has been fixed in NLog 4.3 beta 3 & NLog 4.4 alpha 4

jeremymeng:
NLog 4.4 alpha 4 package has dependencies on several RC2 packages and doesn't seem to work in Castle.Core project.

PR #147, opened 2016-03-03
Add NLog Integration project for .NET Core

jeremymeng:
For part of #128.

The only test failed after upgrading to NLog v4.4.0-alpha3 for both Full .NET Framework and .NET Core.  It is tracked by #146.


jonorossi:
Thanks, merged.

PR #148, opened 2016-03-04
Add notification for virtual but final methods

axelheer:
If a class implements an interface method, the .net runtime marks this method as virtual but final, which means it's not overridable. If such a method should be part of a class (not interface) proxy, we have a problem: since it's marked as final, it's not "accepted" for our proxy. BUT, since it's marked as virtual, the configured proxy generation hook doesn't get any notification.


jonorossi:
@axelheer thanks for your contribution, this sounds like a good change to me, I can't think of any reason we wouldn't want to notify the hook.

PR #149, opened 2016-03-07
Fix for optional, nullable enum method parameters

### LINKED ISSUE #141: Can't create a DynamicProxy for an interface containing method with optional, nullable enum parameter using net45 library
The problem seems to stem from how `ParameterBuilder` handles default values for nullable enums. The value reported by `ParameterInfo.DefaultValue` for such a parameter is an `int`, but trying to use `ParmeterBuilder.SetConstant` for a parameter of a nullable enum type doesn't work (and, as far as I can tell, this is a bug in `ParameterBuilder.SetConstant` itself).

As a larger question, why does DynamicProxy even try to set default values for parameters on the generated class? Normally, default values are inserted by the compiler at the call site - default parameters are mostly compile-time, call-site syntactic sugar. In fact, it's possible for the interface and the implementation to disagree on what the default parameter value is and even on whether a particular parameter is optional or not. The compiler will use whichever default values correspond to the method on the most-specific type that the compiler knows statically. If you're calling through the concrete implementation class, it will use the defaults defined for the concrete method. If you're calling through the interface, it will use the defaults defined for the interface method.

But when using DynamicProxy, the compiler can't see the generated class. Call sites are always going to be invoking through the interface or base class, so it shouldn't be necessary to copy the default parameter values down to the parameters on the generated class. The only place where it would even be visible would be via runtime reflection on the generated type... which, while possible, doesn't seem like something that should be relied upon (especially since the net40-client and earlier libraries doesn't behave this way).

Given that libraries compiled for net40-client and earlier don't exhibit this behavior, my workaround is to just run against one of those. But this net45-specific codepath already caused problems with optional decimals in issue #87. Maybe the call to `SetConstant` should be wrapped in a try/catch and exceptions should be ignored. Or maybe the whole behavior should be removed... it seems weird that the net40-client and net45 libraries behave differently.

Here's a test case:

    [Test]
    public void CanCreateProxyForInterfaceWithNullableEnumMember()
    {
        Assert.DoesNotThrow(() =>
        {
            proxyGenerator.CreateInterfaceProxyWithTarget<IContainsMethodWithDefaultEnumParameter>(
                new ConcreteContainsMethodWithDefaultEnumParameter());
        });
    }

    public class ConcreteContainsMethodWithDefaultEnumParameter : IContainsMethodWithDefaultEnumParameter
    {
        public void Do(ConsoleColor? value = ConsoleColor.Cyan)
        {
        }
    }

    public interface IContainsMethodWithDefaultEnumParameter
    {
        void Do(ConsoleColor? value = ConsoleColor.Cyan);
    }

And here's the exception:

    System.ArgumentException: Constant does not match the defined type.
       at System.Reflection.Emit.TypeBuilder.SetConstantValue(ModuleBuilder module, Int32 tk, Type destType, Object value)
       at Castle.DynamicProxy.Generators.Emitters.MethodEmitter.DefineParameters(ParameterInfo[] parameters)
       at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateMethod(String name, MethodAttributes attributes, MethodInfo methodToUseAsATemplate)
       at Castle.DynamicProxy.Generators.MethodGenerator.Generate(ClassEmitter class, ProxyGenerationOptions options, INamingScope namingScope)
       at Castle.DynamicProxy.Contributors.CompositeTypeContributor.ImplementMethod(MetaMethod method, ClassEmitter class, ProxyGenerationOptions options, OverrideMethodDelegate overrideMethod)
       at Castle.DynamicProxy.Contributors.CompositeTypeContributor.Generate(ClassEmitter class, ProxyGenerationOptions options)
       at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.GenerateType(String typeName, Type proxyTargetType, Type[] interfaces, INamingScope namingScope)
       at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.<>c__DisplayClass1.<GenerateCode>b__0(String n, INamingScope s)
       at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, Object target, ProxyGenerationOptions options, IInterceptor[] interceptors)
       at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithTarget[TInterface](TInterface target, IInterceptor[] interceptors)
       at ReflectionEmitLibrary.ReflectionEmitTest.<>c.<CanCreateProxyForInterfaceWithNullableEnumMember>b__0_0() in C:\src\Sandbox\ReflectionEmitLibrary\ReflectionEmitTest.cs:line 18
       at NUnit.Framework.Constraints.ThrowsConstraint.VoidInvocationDescriptor.Invoke()
       at NUnit.Framework.Constraints.ThrowsConstraint.ExceptionInterceptor.Intercept(Object invocation)

balefrost:
Fixes #141 

I only have VS2015 on my computer, and I ran into a few difficulties:

1. Needed to add `C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\x64` to `PeVerifyProbingPaths` to avoid test failures.
1. VS wanted to migrate app.config values to Settings.Settings. That is, VS removed the `<userSettings>` element from app.config completely; instead, it inlines that value into the `Settings.settings` and `Settings.Designer.cs` files (duplicated in both places). I'm not sure how user settings were handled in previous versions of Visual Studio, but I guess it now wants to bake the default values into the DLLs. 
1. Couldn't build the `SL40` or `SL50` configurations from Visual Studio, either in `Castle.Core.sln` or `Castle.Core-SL.sln`. Was able to build and run tests from the command-line. Didn't investigate this too much, but I'm guessing that the command-line build does things slightly differently from the VS build.
1. Was not able to test on Mono. I would appreciate if somebody with an established Mono environment would run the unit tests.

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ClassProxyWithDefaultValuesTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ClassProxyWithDefaultValuesTestCase.cs
@@ -34,14 +34,13 @@ namespace Castle.DynamicProxy.Tests
 		// https://github.com/mono/mono/blob/master/mcs/class/corlib/System.Reflection.Emit/ParameterBuilder.cs#L101
 		[Ignore("System.ArgumentException : Constant does not match the defined type.")]
 #endif
-		public void MethodParameterWithDefaultValue_DefaultValueIsSetOnProxiedMethodAsWell()
+		public void MethodParameterWithDefaultValue_DefaultValueIsNotSetOnProxiedMethod()
 		{
 			var proxiedType = generator.CreateClassProxy<ClassWithMethodWithParameterWithDefaultValue>().GetType();
 
 			var parameter = proxiedType.GetMethod("Method").GetParameters().Single(paramInfo => paramInfo.Name == "value");
 
-			Assert.True(parameter.HasDefaultValue);
-			Assert.AreEqual(3, parameter.DefaultValue);
+			Assert.False(parameter.HasDefaultValue);
 		}
 
 		[Test]
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/Classes/ClassWithMethodsWithAllKindsOfOptionalParameters.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/Classes/ClassWithMethodsWithAllKindsOfOptionalParameters.cs
@@ -15,6 +15,7 @@
 #if !DOTNET35 && !SILVERLIGHT
 namespace CastleTests.DynamicProxy.Tests.Classes
 {
+	using System;
 	using CastleTests.DynamicProxy.Tests.Interfaces;
 
 	public class ClassWithMethodsWithAllKindsOfOptionalParameters : InterfaceWithMethodsWithAllKindsOfOptionalParameters
@@ -236,6 +237,12 @@ namespace CastleTests.DynamicProxy.Tests.Classes
 		}
 
 		public virtual void MethodWithOptionalNonDefaultNullableDecimalParameter(decimal? d = 3m)
+		{
+		}
+		public void MethodWithOptionalNullableEnumParameter(ConsoleColor? c = null)
+		{
+		}
+		public void MethodWithOptionalNonDefaultNullableEnumParameter(ConsoleColor? c = ConsoleColor.Cyan)
 		{
 		}
 	}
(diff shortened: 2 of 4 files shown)


jonorossi:
Thanks @balefrost.

1. We've just recently added (248d9257c72ac42446c33829372d6359afdf4293) the v10.0 directory to the peverify probing paths in the netcore branch. I also got rid of the .settings file and just made it hardcoded also in the netcore branch so let's just wait until that branch merged back.
2. One of the reasons I dislike those generated .settings files because of this sort of tooling pain, as per the first item I've got rid of it all in the netcore branch.
3. Don't worry about Silverlight, it is on the chopping block for v4. Yep, there are differences because the VS tooling is painful.
4. I've set up your PR on our build server: http://builds.castleproject.org/project.html?projectId=CorePr149

Two unit tests failed on Mono. The unit test that was ignored with "Expected: False  But was: True" as the reason is still working different on Mono, and now that the default value setting code is gone the other unit test that threw on Mono now does the same. Can you restore the ignores for both tests with the expected false but true reason. Thanks.

balefrost:
Sorry for the delay; I've been a bit busy.

I put the `#if __MonoCS__` check back. It looks like everything builds on the build server.

jonorossi:
Thanks

PR #151, opened 2016-03-09
Replace generated AssemblyInfo files with CommonAssemblyInfo

### LINKED ISSUE #133: .NET Core - remove AssemblyInfo generation

jonorossi:
Fixes #133.


buildscripts/CommonAssemblyInfo.cs
```diff
@@ -12,12 +12,41 @@
...
 [assembly: AssemblyCompany("Castle Project")]
-[assembly: AssemblyCopyright("Copyright (c) 2004-2014 Castle Project - http://www.castleproject.org")]
-#if ASSEMBLY_CLSCOMPLIANT_FALSE
-[assembly: System.CLSCompliant(false)]
-#else
-[assembly: System.CLSCompliant(true)]
+[assembly: AssemblyCopyright("Copyright (c) 2004-2016 Castle Project - http://www.castleproject.org")]
+
+[assembly: AssemblyVersion("3.3.0")]
+[assembly: AssemblyFileVersion("3.3.4")]
```
blairconrad: Looks good. For ease of future maintenance, I'd consider

1. pulling version numbers out of Settings.proj and having the Get-NuGetVariables read from this file, and
2. maybe making the major/minor/update components into consts so they can be defined once

These things are definitely "bonus".

One thing we do lose, though, is the BuildName in the `AssemblyInformationalVersion`. Comes from TeamCity's Build_Number, I think. This was one of the things that I wasn't sure whether it was important to you.

buildscripts/CommonAssemblyInfo.cs
```diff
@@ -12,12 +12,41 @@
...
+	#if DOTNET45 // order is important
+	"Castle Core for .NET Framework 4.5"
+	#elif DOTNET40
+	"Castle Core for .NET Framework 4.0"
+	#elif DOTNET35
+	"Castle Core for .NET Framework 3.5"
+	#else
+	"Castle Core for .NET Core 1.0"
+	#endif
+)]
```
blairconrad: All DLLs will have same name, including the logging plugins.

(Also no "compact framework" for 4.0, which doesn't worry me a bit.)

blairconrad:
Generally looks good me, @jonorossi. The generation (and the output) is much simplified. I'm sorry that my dithering forced _you_ to put more time into it, but I was unsure which of the differing bits were important. :disappointed:

blairconrad:
Oh, is the version number still required in caslte.core/project.json?

buildscripts/CommonAssemblyInfo.cs
jonorossi: Logged #152 to sort out the consolidation.

buildscripts/CommonAssemblyInfo.cs
jonorossi: Yep, one of the reasons I got rid of the dot between "Castle" and "Core", I'm setting `AssemblyProduct` and leaving `AssemblyTitle` empty. Being pragmatic the version resource is less important to me than the .NET module name in the assembly, and the file name.

Yep, don't really care that it is sort of meant to be for the client profile, client profile is so dead by now anyway.

jonorossi:
@blairconrad don't worry, I didn't tag this one up-for-grabs because I knew it was going to be a bit hairy especially that we also had a pile of inconsistent `SecurityAssemblyInfo.cs` files littered around too.

I'll merge this then and we can sort of getting consolidated version numbers in #152.

Issue #152, opened 2016-03-09
.NET Core - consolidate version numbers


jonorossi:
Now that we've got hardcoded version numbers in 5 new `project.json` files, and now also in `CommonAssemblyInfo.cs` in addition to Settings.proj let sort something out.

Can we just have all version numbers stored in Git as `0.0.0` and have the CI build replace the version number in the files like it currently does for our current nuspecs. If so we could get the build number into the `AssemblyInformationalVersion` again.

blairconrad:
Do the netcore projects require the version in `project.json`, or could they read the versions from a `CommonAssemblyInfo.cs` file like I'm used to in old-style .NET?

If the latter, I've an idea in mind that I'd flesh out and present when not at work…

jonorossi:
I suspect the DNX tools require it in there to build the NuGet package.

/cc @jeremymeng

jonorossi:
If we get rid of the csproj files in #134, then I'm pretty sure we don't need the version attributes in `CommonAssemblyInfo.cs`.

blairconrad:
Re the dnx NuGet building, "darn".
Yes, if the versions are required in the `project.json` files, then getting out of `CommonAssemblyInfo.cs` makes sense. I'd just thought we could have a cleaner solution with them _in_ `CommonAssemblyInfo.cs` if we didn't need in `project.json` **and** if you wanted to continue synchronizing versions across the projects.

jonorossi:
Yes, we'll keep all the packages versioned in lock-step otherwise it will be a nightmare and achieve nothing, they are only separate packages to separate out the dependencies because each assembly is tiny. If we do get rid of csproj it would leave us with versions only in project.json files because Settings.proj would go, then maybe a little script could update them all at once.

blairconrad:
@jonorossi, we currently build the nupkg by supplying an explicit version number, as you noted above. So mightn't we be able set the `project.json` versions all to 0.0.0 (for now? for a while)? The attributes in `CommonAssemblyInfo.cs` are respected. I'm working on something, if you're open to this.

jonorossi:
If we went with something like #153 I've got a bunch of comments on implementation, however it feels like we are pushing against the tooling pretty hard to do something like that.

[...]

PR #153, opened 2016-03-10
Consolidate version numbers

blairconrad:
For discussion, here's _an_ approach that consolidates version numbers (#152).
It restores file generation, just to write in the TeamCity `build_number`, as I could not figure out how to supply it.
Worse, in order to support a "clone and build in the IDE" experience, it generates the file in a pre-build step.

I'm not sure how well-regarded any of this is, but present it for consideration/discussion.

**Update - wildly different implementation added**


Castle.Core.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+VisualStudioVersion = 12.0.21005.1
```
blairconrad: drat, thought I got rid of that change.

buildscripts/AssemblyMetadata.cs
```diff
@@ -0,0 +1,7 @@
+internal static class AssemblyMetadata
+{
+	public const string MajorVersion = "3";
+	public const string MinorVersion = "3";
+	public const string BuildVersion = "4";
+	public const string BuildNumber = "0";
+}
```
blairconrad: This would be the new place to set the version number. `Get-NuGetVars.ps1` has been changed to parse this instead of `Settings.proj` (which could maybe lose the version numbers?)

blairconrad:
@jonorossi. I took the liberty of putting the functionality in `Get-NuGetVars.ps1`, since it was called by TeamCity anyhow, and did a lot of the work already.
Then I felt bad about the name, so changed to `Set-VersionInfo.ps1`, as it's the best I could think of. If you'd rather the original name (or another), or any other changes, I'm happy to comply.

But I'm hopeful that this is closer to what you want, so I'm removing the "DO NOT MERGE".

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+environment. Then constructs several flavor version strings and
+uses them, along with the current year to
+
+  - update buildscripts/CommonAssemblyInfo.cs,
+  - update all project.json files, and
+  - emit TeamCity directives so it can know about the values
+
+#>
+
+[xml] $settings = Get-Content 'Settings.proj'
```
blairconrad: As you can see, I didn't move the canonical version number components from `Settings.proj`. If you've a preferred place, I could do so. I considered the `AssemblyFileVersion` portion of `CommonAssemblyInfo.cs`, but thought that editing that value near all the `0.0.0`s might be confusing.

jonorossi:
> Get-NuGetVars.ps1, since it was called by TeamCity anyhow, and did a lot of the work already.

Currently TeamCity is only running that script for the "Pack" configuration, however we can change that.

Looking at the diff now.

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+$version_group = $settings.Project.PropertyGroup[1]
+
+$major = $version_group.Project_Major
+$minor = $version_group.Project_Minor
+$build = $version_group.Project_Build
```
jonorossi: I know the old script named this "build", but I think should should align with semver and refer to it as "patch" so it isn't confusing with "build number"

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+if ( $env:release_build -eq "true" )
+{
+    $version_suffix = ""
+}
+else
+{
+    $version_suffix = "-ci{0:00000}" -f [int]$env:build_number
+}
+
+$assembly_version = "$major.$minor.0";
```
jonorossi: I wonder if making the assembly version "M.0.0.0" is a better idea, otherwise it becomes a breaking change for the .NET Framework since it is strong named. Realised this when I saw Serilog does that.

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+$assembly_file_version = "$major.$minor.$build"
+$assembly_informational_version = "$assembly_file_version$version_suffix"
+$year = (Get-Date).Year
+
+Write-Host "##teamcity[setParameter name='ReleaseVersion' value='$assembly_informational_version']"
+Write-Host "##teamcity[setParameter name='CurrentYear' value='$year']"
+
+$common_assembly_info_file = "BuildScripts\CommonAssemblyInfo.cs"
```
jonorossi: "buildscripts" directory name is lowercase.

buildscripts/Set-VersionInfo.ps1
jonorossi: I was thinking of moving this into TeamCity in the "Build number format", but happy for it to continue using Settings.proj.

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+$content = $content.Replace("AssemblyVersion(`"0.0.0", "AssemblyVersion(`"$assembly_version")
+$content = $content.Replace("AssemblyFileVersion(`"0.0.0", "AssemblyFileVersion(`"$assembly_file_version")
+$content = $content.Replace("AssemblyInformationalVersion(`"0.0.0", "AssemblyInformationalVersion(`"$assembly_informational_version")
+
+$content | Set-Content -Encoding UTF8 $common_assembly_info_file
+
+# Change version in every project.json
+# Inspired by https://github.com/c3-ls/CastleWindsor-MsExtensionsIntegration/blob/d2c5d361d25fc3260ca8fcaa36a4f34fa81ad7b9/build-set-version.ps1
+Get-ChildItem -Filter project.json -Recurse | ForEach-Object {
+    $json = (Get-Content -Raw -Path $_.FullName | ConvertFrom-Json)
```
jonorossi: I'll need to make sure we are running PowerShell 3.0+ for `ConvertFrom-Json`.

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,58 @@
...
+$content = $content.Replace('Copyright (c) 2004-$CurrentYear$', "Copyright (c) 2004-$year")
+$content = $content.Replace("AssemblyVersion(`"0.0.0", "AssemblyVersion(`"$assembly_version")
+$content = $content.Replace("AssemblyFileVersion(`"0.0.0", "AssemblyFileVersion(`"$assembly_file_version")
+$content = $content.Replace("AssemblyInformationalVersion(`"0.0.0", "AssemblyInformationalVersion(`"$assembly_informational_version")
+
+$content | Set-Content -Encoding UTF8 $common_assembly_info_file
+
+# Change version in every project.json
+# Inspired by https://github.com/c3-ls/CastleWindsor-MsExtensionsIntegration/blob/d2c5d361d25fc3260ca8fcaa36a4f34fa81ad7b9/build-set-version.ps1
+Get-ChildItem -Filter project.json -Recurse | ForEach-Object {
```
jonorossi: `ForEach-Object`, I haven't seen that commandlet used like that in ages :smile:, I always use the alias: `%`. I don't mind either way.

jonorossi:
@blairconrad looks great. I really didn't like the initial implementation, sorry if it sounded like I was telling you what to do. If you can get the comments fixed up, I'll take another pass and I think we can get this merged. Another step closer to not needing the MSBuild build scripts! :tada:

~~Any thoughts on how the Mono build will be able to run this?~~ Actually it doesn't need to because we don't allow any Mono conditionals in shipping code and we don't package it, and only need the configuration to run the unit tests on Linux.

buildscripts/Set-VersionInfo.ps1
blairconrad: Can do. It seems we're dropping the build number from the `AssemblyFileInfo` as well (it was there before 55e152f49fc564581114fc0fa5908790c1bbc80d). I'll put that back too, if you've no objection, so it will be "M.m.p.b"

buildscripts/Set-VersionInfo.ps1
blairconrad: how embarrassing!

buildscripts/Set-VersionInfo.ps1
blairconrad: I tend to be a little more formal in the scripts. On the command, I'd write `%` as well. Also `ls`.

buildscripts/Set-VersionInfo.ps1
blairconrad: From http://builds.castleproject.org/agentDetails.html?id=5&tab=agentParameters&kind=conf:

image

buildscripts/Set-VersionInfo.ps1
blairconrad: done

blairconrad:
> sorry if it sounded like I was telling you what to do

@jonorossi, it's your project. _Tell me what to do_. The only reason I'd be upset due to you telling me what to do is that I feel bad for not providing enough value out of the gate—there's no animosity toward you at all. I'd prefer to get things right the first time, mostly because it would save you time and effort, but of paramount concern is that Castle.Core get solutions that meet its needs and match its style. I'm not here to get _my take_ on a solution merged, but to get the best practical solution merged.

blairconrad:
Oh, drat. Some of the failure messages have the versions baked in. I'm getting a failure on the NETCORE build. I didn't in NET40 last night, but perhaps I missed something. Investigating.

blairconrad:
Okay. I was thrown because I'd thought that the `CommonAssemblyInfo.cs` was included by the test project (it's not), and also, I didn't expect the modified project.json to produce a version like "3.3.4.31" from `    "version":  "3.3.4.31-ci00031"`. Who knew?

Excluding the tests' `project.json`s from the version-setting.

jonorossi:
@blairconrad sorry for taking so long to get back to you.

I've set up a [modified build configuration on TeamCity](http://builds.castleproject.org/project.html?projectId=CorePr153) with the additional build step and everything is working, however...

The release version used to be `M.m.p-suffix` which was semver compatible:
```
Write-Host "##teamcity[setParameter name='ReleaseVersion' value='$major.$minor.$build$version_suffix']"
```

I think we should just make things consistent and remove the 4th number from all versions because it'll cause us problems when we want to use this number for the package version. For CI builds users can see the build number in the suffix, and release builds are a one of a kind so build number isn't important. This is what we currently have // with comments:
```
[assembly: AssemblyVersion("3.0.0.0")]
[assembly: AssemblyFileVersion("3.3.4.3")] // lets make this 3.3.4
[assembly: AssemblyInformationalVersion("3.3.4.3-ci00003")] // lets make this 3.3.4-ci00003
```

Probably also worth making the replace of these attributes include the end of the attribute to make it clear.

buildscripts/Set-VersionInfo.ps1
```diff
@@ -0,0 +1,61 @@
...
+$content = (Get-Content -Raw -Path $common_assembly_info_file)
+
+$content = $content.Replace('Copyright (c) 2004-$CurrentYear$', "Copyright (c) 2004-$year")
+$content = $content.Replace("AssemblyVersion(`"0.0.0", "AssemblyVersion(`"$assembly_version")
+$content = $content.Replace("AssemblyFileVersion(`"0.0.0", "AssemblyFileVersion(`"$assembly_file_version")
+$content = $content.Replace("AssemblyInformationalVersion(`"0.0.0", "AssemblyInformationalVersion(`"$assembly_informational_version")
+
+$content | Set-Content -Encoding UTF8 $common_assembly_info_file
+
+# Change version in every project.json, except the projects, because
```
jonorossi: Did you mean "except the tests"?

blairconrad: I DID! Oh, man.

blairconrad:
@jonorossi, regarding https://github.com/castleproject/Core/pull/153/files/46d198ce300b2251231c6fd8903f20ed5b9890d2#issuecomment-199823637, specifically 

> The release version used to be M.m.p-suffix

Yes. I'm sorry to have accidentally changed this. I am really bringing more detriment than benefit to this issue. I will make the changes you outline for the attributes.

And just to be explicit, after that the nuget package version will be the same as the `AssemblyInformationalVersion`, no? e.g. "3.3.4-ci00003" for non-release builds, "3.3.4" for release builds.

blairconrad:
Oh, one thing struck me and I wanted to clarify. You said

> [assembly: AssemblyVersion("3.0.0.0")]
> [assembly: AssemblyFileVersion("3.3.4.3")] // lets make this 3.3.4
> [assembly: AssemblyInformationalVersion("3.3.4.3-ci00003")] // lets make this 3.3.4-ci00003

So you want a 4-component `AssemblyVersion`, but 3-component `AssemblyFileVersion` (3-compnent `AssemblyInformationalVersion` before the optional "-ci" part)?

jonorossi:
> I am really bringing more detriment than benefit to this issue.

Definitely not, you are learning how it works so you can take over :wink:.

> And just to be explicit, after that the nuget package version will be the same as the AssemblyInformationalVersion, no? e.g. "3.3.4-ci00003" for non-release builds, "3.3.4" for release builds.

Yes, exactly.

> So you want a 4-component AssemblyVersion, but 3-component AssemblyFileVersion (and 3-compnent AssemblyInformationalVersion before the optional "-ci" part)?

Sorry, I copied the attributes from dotPeek. Make AssemblyVersion 3-part too like in CommonAssemblyInfo.cs.

blairconrad:
Okay. I think we're in better shape now. From the CI server:

image
(specifically the [artifacts for the 4.0 build](http://builds.castleproject.org/viewLog.html?buildId=13275&tab=artifacts&buildTypeId=CorePr153_MasterNet40))

The file version shows as 3.3.4.0, but it's set to 3.3.4 in the source. I swear.

There was no `Pack` in the new configuration, but from [the build logs](http://builds.castleproject.org/viewLog.html?buildId=13275&buildTypeId=CorePr153_MasterNet40&tab=buildLog#_state=43&focus=49):

```
##teamcity[setParameter name='ReleaseVersion' value='3.3.4-ci00003']
```

Also corrected "projects" → "tests" as you noted and expanded the pattern/replacement so that includes the up to the ending `)` of the attributes. If you prefer, I could expand to include from the `[` to the `]` on the affected lines, but you only mentioned the beginning, so I only expanded in one direction. If you want me to act on all of `[assembly: AssemblyVersion("0.0.0")]`, say the word.

jonorossi:
Looks great, lets get this sucker merged.

blairconrad:
Thanks for your patience. So embarrassed about the careless mistakes.

jonorossi:
> Thanks for your patience. So embarrassed about the careless mistakes.

No problem. Don't be, I've made mistakes both here with the Castle Project and other open source projects that seem careless after the fact.

Issue #155, opened 2016-03-11
.NET Core - compile errors in Mono build


jeremymeng:
```
Compatibility/CustomAttributeExtensions.cs(32,37): error CS0030: Cannot convert type `System.Attribute[]' to `System.Collections.Generic.IEnumerable<T>'
Compatibility/CustomAttributeExtensions.cs(37,37): error CS0030: Cannot convert type `System.Attribute[]' to `System.Collections.Generic.IEnumerable<T>'
Core/Internal/AttributesUtil.cs(47,47): error CS0030: Cannot convert type `object[]' to `System.Collections.Generic.IEnumerable<T>'
Core/Internal/AttributesUtil.cs(49,46): error CS0030: Cannot convert type `object[]' to `System.Collections.Generic.IEnumerable<T>'
Core/Internal/AttributesUtil.cs(71,35): error CS0030: Cannot convert type `object[]' to `System.Collections.Generic.IEnumerable<T>'
Core/Internal/AttributesUtil.cs(73,34): error CS0030: Cannot convert type `object[]' to `System.Collections.Generic.IEnumerable<T>'
```

It seems that Mono compiler doesn't support array => IEnumerable<T> conversion.  I am working on a solution of `yield return` each of the array elements.

jeremymeng:
@jonorossi does the Linux build server build all three configurations (NET35/NET40/NET45)?  I am seeing many warnings-as-error for the NET40 configuration when building using Mono.  However I don't see any when building NET35 and NET45.  I am investigating but if you have any prior experience to share, that would be great.

jeremymeng:
My mcs version is `Mono C# compiler version 4.2.2.0`.

jeremymeng:
Okay, looking into `Castle.Core.Tests.csproj`, I found that only NET40 configurations have 
```
    <WarningLevel>4</WarningLevel>
```

while NET35 and NET45 don't have it.  It is strange that windows build passed.  Could be the difference between the two compilers.  Would you mind if I remove the above from NET40 configurations?

jonorossi:
1. The Linux build is only the NET45 configuration, it runs `xbuild /p:Configuration=NET45-Release /t:RunAllTests buildscripts/Build.proj`.
2. Happy for you to change the unit test project's NET40 configuration to match the others by removing that element so it defaults back to whatever is the default. I'm not to concerned about warnings in the unit tests. I have encountered csc and mcs sometimes with different or extra warnings.
[...]

Issue #94, opened 2015-06-23
Mono: Fix remaining ignored unit tests


jonorossi:
See parent issue: #79

These are the remaining ~~136~~ 117 (1391 - 1274) unit tests that I've ignored. Feel free to send in a PR for any of them on the ~~`mono-support`~~ `master` branch.

I've permanently ignored a few unit tests using the `[Platform(Exclude = "mono", Reason = "...")]` syntax, and am using `#if __MonoCS__  [Ignore("...")]  #endif` for those still to fix.
## DynamicProxy
- ~~`Assertion at sgen-alloc.c:460, condition '*p == NULL' not met`~~
  - ~~Fixed in https://bugzilla.xamarin.com/show_bug.cgi?id=28182~~
  - ~~It isn't in the [4.0.2 release notes](http://www.mono-project.com/docs/about-mono/releases/4.0.2/) however it is fixed in 4.0.2 in https://github.com/mono/mono/commit/ef16e5a45c0e8fa6d9305aaa885c53ad008f10b2~~
- ~~`System.Type[] doesn't implement interface Castle.DynamicProxy.IInvocation`~~
  - ~~Fixed in 4.0.2, caused by the same defect as the `Assertion at sgen-alloc.c:460, condition '*p == NULL' not met` runtime crash~~
- ~~`Assertion at reflection.c:2745, condition 'field' not met`~~
  - #72
- `System.InvalidCastException : Cannot cast from source type to destination type.`
  - src/Castle.Core.Tests/GenericClassProxyTestCase.cs
- `System.ArgumentException : Constant does not match the defined type.`
  - src/Castle.Core.Tests/DynamicProxy.Tests/ClassProxyWithDefaultValuesTestCase.cs
  - src/Castle.Core.Tests/DynamicProxy.Tests/ClassProxyWithMethodsWithOptionalParametersTestCase.cs
  - src/Castle.Core.Tests/DynamicProxy.Tests/InterfaceProxyWithMethodsWithOptionalParametersTestCase.cs
- ~~`NUnit.Framework.AssertionException was expected`~~
  - ~~src/Castle.Core.Tests/BaseTestCaseTestCase.cs... was caused because peverify isn't available on mono, permanently disabled the unit test~~
- `Expected: Castle.DynamicProxy.Generators.GeneratorException, But was: System.ArgumentNullException`
  - src/Castle.Core.Tests/BasicClassProxyTestCase.cs
[...]

PR #156, opened 2016-03-12
Replace most __MonoCS__ usage and fix mono build errors

jeremymeng:
For #155.

There are still a couple of __MonoCS__ usage left and they seem necessary.

- Two places - to avoid loading [adodb.dll](https://github.com/jeremymeng/Core/blob/master/src/Castle.Core.Tests/ComInterfacesTests.cs#L15) and [RhinoMocks](https://github.com/jeremymeng/Core/blob/master/src/Castle.Core.Tests/RhinoMocksTestCase.cs#L15).
- Four places for fixing build errors with Mono compiler in https://github.com/jeremymeng/Core/blob/mono/src/Castle.Core/Core/Internal/AttributesUtil.cs. As the errors are Mono only and in all three build configurations.

- [ ] Three tests failed. To be determined whether to disabled them.


jonorossi:
Sorry if I was unclear, I didn't mean to just replace the conditional compilation of those mono-ignored unit tests with `Assert.Ignore()` with the same reason. You'll see at the top of #94 that I was using the `[Platform(Exclude = "mono", Reason = "...")]` syntax for permanently ignored investigated failing tests (only two cases so far, peverify doesn't exist and bugs in tracing configuration not matching Windows). I was using the `#if __MonoCS__ [Ignore("...")] #endif` syntax for those that needed to be investigated so we could get a green build and knew if anything else failed.

Even if we don't track down a mono issue we need to make sure reach one is ignored with a decent reason explaining why it doesn't work on mono knowing that it shouldn't cause too many problems. I was actually thinking of including a "Known Mono Defects" section in the release notes (or our web site) using the reason specified in the attribute.

It'd be great if you could fix the compilation issues in this PR, then we can step through the ignored tests in small batches and assign some decent reasons for leaving them ignored. We don't have to fix the Mono problems, just get a decent reason in there that someone reading a "Known Mono Defects" list would have a good place to start if that is what they are doing.

Thoughts?

jeremymeng:
@jonorossi I understand and agree that the tests ignored still need investigation.  My change just moves from ignoring at compile time to ignoring at run-time, with the benefit of removing the `__MonoCS__`.   Another goal of doing this is to allow experimenting building tests using dnx then running tests using Mono.  Although some `__MonoCS__` still left.

I will submit a PR against the master branch for build errors unrelated to .NET Core work, and will update this one for #155 where array failed to be converted to IEnumerable<T>.

jeremymeng:
Replaced by PR #161

PR #159, opened 2016-03-14
Fix compile errors in Mono net40 build

### LINKED ISSUE #158: Compilation error in Mono NET40 build
There are two issues in Mono NET40 build in master branch

1. lib/net40/Serilog.FullNetF*x*.dll does not match the csproject where the file name is Serilog.FullNetF*X*.dll.  On Linux the build failed to resolve the dll as file path is case sensitive.
1. NET40 builds uses `<WarningLevel>4</WarningLevel>` while NET35 and NET45.  This setting cause more warnings to be treated as errors than default settings.  For example, warning for unused variables are treated as errors.

jeremymeng:
Fixes #158.


jonorossi:
We should leave the file name casing the same as the upstream project and fix the csproj file.

https://github.com/serilog/serilog/tree/master/src/Serilog.FullNetFx

jeremymeng:
Yes, that is better.  PR updated.

jonorossi:
Merged, thanks.

PR #161, opened 2016-03-15
Fix build errors in Mono build.

### LINKED ISSUE #155: .NET Core - compile errors in Mono build

jeremymeng:
in Mono, *.GetCustomerAttributes() returns an object array which cannot
be converted to IEnumerable<T>.  So changing to yield return each
element from the array.  In Attributesutil.GetAttributes() I had to use
__MonoCS__ because on Windows TypeInfo.GetCustomAttributes() could
return IEnumerable<Attribute>, and Attribute cannot be casted to T where
T : class.

Fixes #155 

Still needs the PR against master branch for fixes to other build errors.

### DIFF
--- a/src/Castle.Core/Compatibility/CustomAttributeExtensions.cs
+++ b/src/Castle.Core/Compatibility/CustomAttributeExtensions.cs
@@ -29,12 +29,18 @@ namespace System.Reflection
 	{
 		public static IEnumerable<T> GetCustomAttributes<T>(this Assembly element) where T : Attribute
 		{
-			return (IEnumerable<T>)Attribute.GetCustomAttributes(element, typeof(T));
+			foreach (T a in Attribute.GetCustomAttributes(element, typeof(T)))
+			{
+				yield return a;
+			}
 		}
 
 		public static IEnumerable<T> GetCustomAttributes<T>(this MemberInfo element, bool inherit) where T : Attribute
 		{
-			return (IEnumerable<T>)Attribute.GetCustomAttributes(element, typeof(T), inherit);
+			foreach (T a in Attribute.GetCustomAttributes(element, typeof(T), inherit))
+			{
+				yield return a;
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


jonorossi:
@jeremymeng sorry for the delay in getting back to you on this PR. I can't accept this PR, we can't add usage of `#if __MonoCS__` into the shipping code after all the work to remove all usage of it. If there is no way to work around the Mono defect by changing our code for all platforms then I'd prefer to leave the Mono build broken.

I found similar bug reports for erroneous CS0030 compiler errors for Mono, but none open for this specific case or since 2014:
* https://bugzilla.xamarin.com/show_bug.cgi?id=18474
* https://bugzilla.xamarin.com/show_bug.cgi?id=17469
* https://bugzilla.xamarin.com/show_bug.cgi?id=11241

I'll close this PR for now.

PR #164, opened 2016-03-22
Fix build errors in Mono build.

jeremymeng:
Mono compiler issues CS0030 error when an array (of object or Attribute)
is casted to IEnumerable<T>.  So change to yield return each element
from the array.

Fixes #155.

PR #166, opened 2016-03-23
Fix attribute cast for netcore

blairconrad:
Restores netcore build after #164. The downside is that I don't have a mono build environment set up, so _may_ be undoing @jeremymeng's fix. I don't think so, though.
I'll get a mono environment sometime…


blairconrad:
Haha. Oops.

PR #167, opened 2016-03-23
Cast attribute type using as

blairconrad:
Restores netcore build after #164.

### DIFF
--- a/src/Castle.Core/Core/Internal/AttributesUtil.cs
+++ b/src/Castle.Core/Core/Internal/AttributesUtil.cs
@@ -30,7 +30,7 @@ namespace Castle.Core.Internal
 		/// </summary>
 		/// <param name="type">The type.</param>
 		/// <returns>The type attribute.</returns>
-		public static T GetAttribute<T>(this Type type) where T : class
+		public static T GetAttribute<T>(this Type type) where T : Attribute
 		{
 			return GetAttributes<T>(type).FirstOrDefault();
 		}
@@ -40,21 +40,11 @@ namespace Castle.Core.Internal
 		/// </summary>
 		/// <param name="type">The type.</param>
 		/// <returns>The type attributes.</returns>
-		public static IEnumerable<T> GetAttributes<T>(this Type type) where T : class
+		public static IEnumerable<T> GetAttributes<T>(this Type type) where T : Attribute
 		{
-			if (typeof(T) != typeof(object))
+			foreach (T a in type.GetTypeInfo().GetCustomAttributes(typeof(T), false))
 			{
-				foreach (var a in type.GetTypeInfo().GetCustomAttributes(typeof(T), false))
-				{
-					yield return (T)a;
-				}
-			}
-			else
-			{
-				foreach (var a in type.GetTypeInfo().GetCustomAttributes(false))
-				{
-					yield return (T)a;
-				}
+				yield return a;
 			}
 		}
 
@@ -63,7 +53,7 @@ namespace Castle.Core.Internal
 		/// </summary>
 		/// <param name="member">The member.</param>
 		/// <returns>The member attribute.</returns>
-		public static T GetAttribute<T>(this MemberInfo member) where T : class
+		public static T GetAttribute<T>(this MemberInfo member) where T : Attribute
 		{
 			return GetAttributes<T>(member).FirstOrDefault();
 		}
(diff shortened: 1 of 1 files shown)


blairconrad:
The downside is that I don't have a mono build environment set up, so _may_ be undoing @jeremymeng's fix. I don't think so, though.
I'll get a mono environment sometime…

jonorossi:
This doesn't seem like the right fix for `CS0030: Cannot convert type 'System.Attribute' to 'T'`, I doubt any of the consumers of that utility method expect nulls in the sequence.

Isn't the problem that `MemberInfo.GetCustomAttributes` on .NET Framework returns `object[]`, while the [extensions methods on .NET Core](https://github.com/dotnet/corefx/blob/39471b56bd0241e60a33b4179a04187b96481547/src/System.Reflection.Extensions/ref/System.Reflection.Extensions.cs#L25-L42) return `IEnumerable<Attribute>`? I think we ran into something similar somewhere else and decided that there shouldn't be any attributes that don't inherit from `System.Attribute` so that is why .NET Core returns a more specific type now.

Could we change this line like so (and the other methods with generic parameters for an attribute type)? If so, we might even be able to change the code back to the one liner we had before rather than looping and casting in here.
```
public static IEnumerable<T> GetAttributes<T>(this Type type) where T : class
public static IEnumerable<T> GetAttributes<T>(this Type type) where T : Attribute
```

blairconrad:
> I doubt any of the consumers of that utility method expect nulls in the sequence.

I doubt it too, but this change won't introduce nulls. If `GetCustomAttributes` (on `type.GetTypeInfo()` and on `member`) aren't returning nulls and are obeying the contract (returning only attributes assignable to `T` in the case that `typeof(T)` is a parameter), then the `as` cast will do the same thing as the hard `(T)a` would.

I had just been following along with the branch and loop approach that had been used before, but much prefer the `where T: Attribute` idea. At work, so can't try it out, but will give it a go tonight if nobody else has by then (I will not be offended if someone else fixes).

While we're on the topic, the methods seem to be lightly used. Of the top of your head, do you know whether they intended for public consumption, or only Castle.Core internal use?

jonorossi:
Ah yes of course, it is specifying the type.

I hate everything in the `Castle.Core.Internal` namespace that is public, logically it makes no sense. I know that Windsor uses some of the stuff but doesn't seem to use `AttributesUtil`, happy for it to be made internal.

Over time post-4.0 I'd actually like for Castle Core to loose some weight, it has collected a lot of stuff over the years which was used by multiple projects when everything was built from a single trunk but either doesn't belong in Castle Core anymore or is obsolete.

jeremymeng:
I am very sorry for not testing the netcore build.

> Could we change this line like so (and the other methods with generic parameters for an attribute type)?

[The other file has this](https://github.com/castleproject/Core/pull/164/files#diff-ecbf2ec4fe0377831d024bbc9e64fd2b), but also caused build errors on Mono so I changed it to yield return too.

Changing from `T : class` to `T : Attribute` should get rid of the cast and the errors in netcore build.  I think yield return is still needed for Mono build.

jeremymeng:
I submitted https://github.com/castleproject/Core/pull/168.  Please have a look.

blairconrad:
Oh, shoot. @jeremymeng, I was just pushing another version, while not looking here. Ah well, @jonorossi is spoiled for choice.

I didn't make any methods internal, as it looked to me like Windsor was using them. At least in [ Castle.Windsor.Diagnostics.Helpers.GetLifestyleDescriptionLong](https://github.com/castleproject/Windsor/blob/20daf3296f5eb9009d03e12ab4a870268d4b2a0f/src/Castle.Windsor/Windsor/Diagnostics/Helpers/DescriptionUtil.cs#L62). Maybe others. I'd be happy to make the change later, if you prefer.

And as @jeremymeng said, the foreach (or alternatively a `Cast<T>()`) is needed to make everyone happy. Not knowing your feelings about excessive Linq, I opted not to make extra changes to the code.

jeremymeng:
Oops I should have add you, @blairconrad so maybe you could've got a notification. Sorry.

> I think we ran into something similar somewhere else and decided that there shouldn't be any attributes that don't inherit from System.Attribute

I found [the test change I did earlier](https://github.com/castleproject/Core/commit/659ef82b9a4d10513fad072f5bae83d075c854f6). I had to pass in an `Attribute` type because the `GetCustomAttributes` extension method in .NET Core does not take arbitrary types like previous `GetCustomAttributes` methods do.

jeremymeng:
@blairconrad I'd vote for your PR since the `object` case is no longer valid.

blairconrad:
@jeremymeng said

> I'd vote for your PR 

Me too, but I thought it impolite to say. :wink: Okay. It's much further into Thursday than it should be when I'm still awake. See you in a few hours.

jeremymeng:
@blairconrad before you go :smile: , did you not have to change other `T : class` in the same file?  I remember they were causing problems for me.

jonorossi:
Thanks guys. I'll get it merged.

> I didn't make any methods internal, as it looked to me like Windsor was using them.

Oh these are extension methods, forgot that since this file doesn't match the xxxExtensions convention. In a future release we can trim these methods down once we get Windsor off them, I suspect the changes we've already made here will probably break Windsor anyway, but that is why this is v4.

blairconrad:
Thanks, @jonorossi, @jeremymeng.

PR #170, opened 2016-03-26
Improve mono build experience on Windows

### LINKED ISSUE #169: Improve Mono on Windows build
This may be completely selfish, as I'm not sure if "Mono on Windows" is something that we really care about. I only bring it up because I installed Mono. On Windows. In order to make sure that .NET Core changes don't break Mono. And I find building/testing inconvenient.

First, the `build.cmd` script doesn't support Mono. `build MONO` fails miserably. I think this is an easy fix, if you're interested.

Second, if I run the [command specified in the README](https://github.com/castleproject/Core#mono):
```
xbuild /p:Configuration=NET45-Release /t:RunAllTests buildscripts/Build.proj
```

then the compilation works fine, but the test-running falls down because the value of `$(TestAssemblies)` is lost between the `BuildProject` target and the `_ExecTestRunner` target. The property still has the correct value in `RunAllTests`, though, which invokes `_ExecTestRunner` via a `CallTarget`. `RunAllTests` depends on `BuildProject`.

I do not understand the property scoping rules or why `xbuild` would behave differently in this regard when on Windows versus on Linux (or when compared to MSBuild), but changing the `CallTarget`s to `DependsOnTargets`es seems to do the trick.
Unless you completely hate this approach (or the whole issue), I'll send a PR for consideration.

blairconrad:
Fixes #169

### DIFF
--- a/buildscripts/build.cmd
+++ b/buildscripts/build.cmd
@@ -19,13 +19,17 @@ IF NOT EXIST %~dp0..\Settings.proj GOTO msbuild_not_configured
 REM Set Framework version based on passed in parameter
 IF "%1" == "" goto no_nothing
 
-REM nesting the netcore build in an IF body prevents reporting of the error level
+REM nesting the build command in an IF body prevents reporting of the error level
 IF /i "%1" NEQ "NETCORE" goto not_netcore
-
 PowerShell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Unrestricted -Command %~dp0Build-NetCore.ps1
 EXIT /B %ERRORLEVEL%
 
 :not_netcore
+IF /i "%1" NEQ "MONO" goto not_mono
+xbuild /p:Configuration=NET45-Release /t:RunAllTests buildscripts/Build.proj
+EXIT /B %ERRORLEVEL%
+
+:not_mono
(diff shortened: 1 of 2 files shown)


buildscripts/build.cmd
```diff
@@ -19,13 +19,17 @@ IF NOT EXIST %~dp0..\Settings.proj GOTO msbuild_not_configured
...
+REM nesting the build command in an IF body prevents reporting of the error level
 IF /i "%1" NEQ "NETCORE" goto not_netcore
-
 PowerShell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Unrestricted -Command %~dp0Build-NetCore.ps1
 EXIT /B %ERRORLEVEL%
 
 :not_netcore
+IF /i "%1" NEQ "MONO" goto not_mono
+xbuild /p:Configuration=NET45-Release /t:RunAllTests buildscripts/Build.proj
+EXIT /B %ERRORLEVEL%
```
blairconrad: I know, my answer to everything is to add a new label with an explicit command, but xbuild doesn't support `/m` or `/ToolsVersion`. By the time I replaced the command and the two switches, the `build.cmd` was even messier. I could go that route if you prefer.

blairconrad:
Oh, thanks. I nearly fell over from the lack of resistance.

PR #172, opened 2016-04-01
Move framework-specific properties into their target framework section.

jeremymeng:
Since more target frameworks could be added in the future.

### DIFF
--- a/src/Castle.Core.Tests/project.json
+++ b/src/Castle.Core.Tests/project.json
@@ -9,22 +9,27 @@
     "Castle.Core": { "target": "project", "version" : "" },
     "Castle.Services.Logging.SerilogIntegration": { "target": "project", "version" : "" },
     "Castle.Services.Logging.NLogIntegration": { "target": "project", "version": ""},
-    "Microsoft.CSharp": "4.0.0",
     "Serilog.Sinks.TextWriter": "2.0.0-beta-505",
-    "System.Console": "4.0.0-beta-23516",
-    "System.Diagnostics.TraceSource": "4.0.0-beta-23516",
-    "xunit": "2.1.0",
-    "xunit.runner.dnx": "2.1.0-rc1-build204"
+    "xunit": "2.1.0"
   },
   "commands": {
     "test": "xunit.runner.dnx"
   },
   "frameworks": {
-    "dnxcore50": { }
+    "dnxcore50": {
+      "compilationOptions": {
+        "define": [ "FEATURE_XUNITNET", "FEATURE_TEST_SERILOGINTEGRATION" ]
+      },
+      "dependencies": {
+        "Microsoft.CSharp": "4.0.0",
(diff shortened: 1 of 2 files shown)


src/Castle.Core.Tests/project.json
```diff
@@ -9,22 +9,27 @@
...
-    "Microsoft.CSharp": "4.0.0",
     "Serilog.Sinks.TextWriter": "2.0.0-beta-505",
-    "System.Console": "4.0.0-beta-23516",
-    "System.Diagnostics.TraceSource": "4.0.0-beta-23516",
-    "xunit": "2.1.0",
-    "xunit.runner.dnx": "2.1.0-rc1-build204"
+    "xunit": "2.1.0"
   },
   "commands": {
     "test": "xunit.runner.dnx"
```
jeremymeng: In RC2 there will be a "testRunner" property: [an example](https://github.com/aspnet/Hosting/blob/6c1247b30cb005aec4091904ecefe8a990296ae0/test/Microsoft.AspNetCore.Hosting.Tests/project.json).

src/Castle.Core/project.json
```diff
@@ -6,36 +6,40 @@
...
+      }
+    }
   },
   "compilationOptions": {
     "warningsAsErrors": true,
-    "define": [ "FEATURE_NETCORE_REFLECTION_API" ],
     "keyFile": "../../buildscripts/CastleKey.snk"
   },
   "compileFiles": [
-    "Properties/InternalsVisibleToTests.cs",
```
jeremymeng: Redundant.

jonorossi:
@jeremymeng thanks, sorry for the delay in looking at this.

jonorossi:
This broke the .NET Core build:

```
error CS8021: No value for RuntimeMetadataVersion found. No assembly containing System.Object was found nor was a value for RuntimeMetadataVersion specified through options.
...
src\Castle.Core.Tests\InternalsVisibleTo.cs(18,48): DNXCore,Version=v5.0 error CS0518: Predefined type 'System.String' is not defined or imported
...
```

jeremymeng:
@jonorossi Hmm the .NET Core build was passing in my AppVeyor build https://ci.appveyor.com/project/jeremymeng/core/build/4.0.0-alpha.18.  Though there are several merged PR's in upstream netcore branch since then.

I am looking into it (http://builds.castleproject.org/viewLog.html?buildId=13358&buildTypeId=CoreNetCore_MasterNetCore&tab=buildLog#_state=24,33&focus=813)

jeremymeng:
The problem is that `ConvertTo-Json` cmdlet by default only serializes two levels deep.  I submitted a fix in https://github.com/castleproject/Core/pull/173.

jonorossi:
Wow, had no idea `ConvertTo-Json` would only serialize two levels deep by default, that really seem silly. Good catch, it worked on my machine which is why I went to clear the cache, was completely lost by the error message.

PR #173, opened 2016-04-06
Set level of contained object to 99 for ConvertTo-Json.

jeremymeng:
The default value of 2 isn't enough to serialize project.json's object representation.  An argument of 99 should be enough.


jonorossi:
Thanks

Issue #100, opened 2015-07-06
Drop support for Silverlight


jonorossi:
With Microsoft's support for Silverlight declining and just general use for new work close to non-existent we need to consider whether the effort to continue keeping it supported in Castle Core is worthwhile. Many of us have wanted to drop it completely (both v4 and v5, v3 support was dropped long ago).

Microsoft has just announced that Microsoft Edge will not run any sort of built-in Silverlight like they are doing with Adobe Flash:

> Support for ActiveX has been discontinued in Microsoft Edge, and that includes removing support for Silverlight.  The reasons for this have been discussed in previous blogs and include the emergence of viable and secure media solutions based on HTML5 extensions.  Microsoft continues to support Silverlight, and Silverlight out-of-browser apps can continue to use it.  Silverlight will also continue to be supported in Internet Explorer 11, so sites continue to have Silverlight options in Windows 10.
> (http://blogs.windows.com/msedgedev/2015/07/02/moving-to-html5-premium-media/)

> Microsoft Edge will support native PDF rendering and Adobe Flash as built-in features rather than external add-ons.
> (http://blogs.windows.com/msedgedev/2015/05/06/a-break-from-the-past-part-2-saying-goodbye-to-activex-vbscript-attachevent/)

This issue is a discussion thread to get people's opinions on dropping all Silverlight support, we want to make an informed decision, a decision hasn't been made yet so we want to hear feedback as we prepare for the next major release.

/cc (lets start with those listed as organisation members of the mocking frameworks)
Moq: @kalebpederson @kzu @tkellogg
FakeItEasy: @adamralph @blairconrad @patrik-hagne @philippdolder
NSubstitute: @AnthonyEgerton @dtchepak

kkozmic:
I'm all for dropping it

jonorossi:
@kkozmic don't worry, I don't think anyone wasn't already aware :wink:

blairconrad:
@jonorossi, thanks for thinking about us mocking frameworks.
You're within your rights to drop whatever you want whenever you want, and I really appreciate that you thought of us.
One thing to remember is that even if we want to keep using the Silverlight Castle.Core libraries, there's always version 3.3.3 to fall back on (and _our_ clients could always fall back on whatever version we last offered SL support in, assuming we remove it).

[...]

PR #176, opened 2016-04-19
Support SL5 (.net 4) for referencing by .NET APTCA assembly

k0st1x:
we have a .net 4 assembly with an APTCA attribute and it referencing a "Moq.Silverlight" assembly (we should reference exactly SL version).

while Castle.Core & Moq.Silverlight moved from SL4 to SL5, that assemblies currently are "security critical" http://stackoverflow.com/questions/12092435/system-methodaccessexception-attempt-by-security-transparent-method-to-access-s/13583246#13583246 (sl4 is for .net2, sl5 is for .net4).

please please please, access this small (but useful for us) pull request : )

### DIFF
--- a/src/Castle.Core/SecurityAssemblyInfo.cs
+++ b/src/Castle.Core/SecurityAssemblyInfo.cs
@@ -1,5 +1,7 @@
 ﻿// Sets up assembly level security settings
-#if ! SILVERLIGHT
+#if SILVERLIGHT
+[assembly: System.Security.SecurityTransparent]
+#else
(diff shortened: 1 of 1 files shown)


jonorossi:
@k0st1x thanks for the pull request, however we made the decision fairly recently in #100 that we will be dropping all support for Silverlight in Castle Core v4.0, the removal work is already done in the `netcore` branch and will be merged to master soon. We've kept Silverlight going for way longer than we've wanted to, sorry to hear you are still maintaining a Silverlight based project, however feel free to maintain your own Silverlight fork.

PR #181, opened 2016-06-13
Enable xmlDoc generation for .NET Core.

jeremymeng:
- Enable xml doc file generation.
- Suppress CS1591 as other build configurations do.
- Conditionally compile StrongNamedModuleDirectory and WeakNamedModuleDirectory as they are only used when there is support to save dynamic assemblies to disk. Their xml doc comments contain cref's to `SaveAssembly()`, which failed to resolve because that method and its overloads are not available in .NET Core build.


jonorossi:
Thanks

PR #188, opened 2016-06-29
Update dependencies to the same versions as in .NET Core project

jeremymeng:
Stop building SerilogIntegration project for NET40 because Serilog v2.0 only support net45.

For updating part of #137.


src/Castle.Services.Logging.SerilogIntegration/Castle.Services.Logging.SerilogIntegration.csproj
```diff
@@ -58,22 +58,12 @@
     <ErrorReport>prompt</ErrorReport>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'NET40-Release|AnyCPU'">
```
jonorossi: Could you please remove the .NET 4.0 and 3.5 sections completely, not sure why VS doesn't do that for you.

jonorossi:
You also missed updating `Castle.Core-Serilog.nuspec`.

src/Castle.Services.Logging.SerilogIntegration/Castle.Services.Logging.SerilogIntegration.csproj
jeremymeng: I manually edited those, trying to mimic the old NET35 one. I will remove them.

jeremymeng:
> You also missed updating Castle.Core-Serilog.nuspec.

Thanks, I updated it.  I also added the packing of .NET Core output for Serilog and NLog integrations.  However local NETCORE build doesn't output to `build\NETCORE` folder.  Do you have an extra steps on build server to do that?  How can I easily testing nuget packing locally?

jonorossi:
@jeremymeng unfortunately the NuGet packing uses the TeamCity "NuGet Pack" build step so not easily runnable outside TeamCity, however here are the steps. I don't really like the whole thing, however I think we should just leave it as is until the new .NET tooling supports multi-targeting since we can't currently use the .NET CLI packing.

The .NET Core build is publishing the following files:
```
src/Castle.Core/bin/Release/netstandard1.3
```

The Pack build config pulls those files and puts them into the build\NETCORE directory as expected by the nuspec file:
```
** => .\build\NETCORE\Release\pkg
```

jonorossi:
@jeremymeng did you forget to push your nuspec change?

jeremymeng:
@jonorossi I was going to yesterday but realized that I also need to add dependencies to the integration nuspec files, which I am working on now.

jeremymeng:
@jonorossi I also added packing of NETCore binaries for NLog/SeriLog integration.  Please help review since I didn't test locally.

jonorossi:
@jeremymeng I've merged my own PR (#199) including these changes, I had to make the following changes:

* The .NET Core build was using NLog 4.4.0-beta12, not beta2 as you added, I've now updated both to beta13
* You missed removing .NET 4.0 from the Serilog nuspec
* Serilog.Sinks.TextWriter doesn't need to be packaged, it is only used by unit tests
* I've upgraded log4net to 2.0.5 (1.2.15 upstream)

Many thanks, you made it much easier.

PR #192, opened 2016-07-06
Fix `netstandard` target framework moniker in Castle.Core.nuspec file

### LINKED ISSUE #191: .net core netstandard1.6
I'm currently upgrading a library from .net core RC1 to .net core rtm.

When using the new version of moq I get the error:
The dependency Castle.Core 4.0.0-alpha001 does not support framework .NETStandard,Version-v1.6

I didn't see an issue when I searched so I decided to open one.
https://github.com/moq/moq4/issues/168

odinserj:
Previous one prevents from installing a package through NuGet, and [NuGet Gallery](https://www.nuget.org/packages/Castle.Core/4.0.0-alpha001) displays this target as "Unsupported 0.0". Correct TFM is `netstandard1.3`, as in other places. Fixes #191.

### DIFF
--- a/buildscripts/Castle.Core.nuspec
+++ b/buildscripts/Castle.Core.nuspec
@@ -17,7 +17,7 @@
       <group targetFramework="net35" />
       <group targetFramework="net40" />
       <group targetFramework="net45" />
-      <group targetFramework=".NETStandard1.3">
+      <group targetFramework="netstandard1.3">
         <dependency id="System.AppContext" version="4.1.0" />
         <dependency id="System.Collections.Specialized" version="4.0.1" />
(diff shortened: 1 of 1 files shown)


jonorossi:
As per #191 closing.

jonorossi:
@jeremymeng I assume this change isn't needed? If it is it'll need to be changed in all 4 nuspecs.

jeremymeng:
No. @jonorossi is right that the issue was caused by the older version of nuget.exe.  I believe that `.NETStandard1.3` is equivalent to `netstandard1.3`.  If you try `dotnet pack` on a `project.json` project targeting netstandard1.3, you will see the former is used in the generated nuspec.

Issue #184, opened 2016-06-28
 BasicClassProxyTestCase.ProxyForNonPublicClass test failure after upgrading to .NET Core RTM


jeremymeng:
```
Assert.Throws() Failure
Expected: typeof(Castle.DynamicProxy.Generators.GeneratorException)
Actual:   typeof(System.ArgumentNullException): Value cannot be null.
Parameter name: classToProxy
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, Object[] constructorArguments, IInterceptor[] interceptors) in D:\work\d545f799076bb8eb\src\Castle.Core\DynamicProxy\ProxyGenerator.cs:line 1424
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, IInterceptor[] interceptors) in D:\work\d545f799076bb8eb\src\Castle.Core\DynamicProxy\ProxyGenerator.cs:line 1334
   at CastleTests.BasicClassProxyTestCase.<>c__DisplayClass2_0.<ProxyForNonPublicClass>b__0() in D:\work\d545f799076bb8eb\src\Castle.Core.Tests\BasicClassProxyTestCase.cs:line 90
```

jeremymeng:
Most likely because the type `System.AppDomainInitializerInfo` is gone in .NET Core. I changed to use another internal type `System.StubHelpers.StubHelpers`, which gave me the same `GeneratorException` type, but with a slightly different exception message.  The assembly name in the message is `System.Private.CoreLib` instead of `mscorlib`: `...because assembly System.Private.CoreLib is strong-named.`.  On .NET Core 1.0 RTM mscorlib just forwards types to System.Pivate.CoreLib.

Ideally the test shouldn't depend on external things that is subject to change.  However, I could see that adding an assembly to test this scenario seems an overkill. We could change the verification to just `string.StartsWith()`, or a little more involved using Regex.

@jonorossi what's your opinion?

kzu:
For tracking purposes, linking https://github.com/moq/moq4/pull/225 from here.

jonorossi:
@jeremymeng change to using `StartsWith` sounds fine to me.

[...]

PR #195, opened 2016-07-08
Fix test failure NET35 build

jeremymeng:
Previous fix does not work in NET35 build because the type System.StubHelpers.StubHelpers isn't available on v3.5.  This change uses an internal type from NLog.

This also avoids the issue of different assembly names due to type-forwarding on .NET Core.


jeremymeng:
For #184

jonorossi:
I've merged #197 instead.

Issue #127, opened 2016-02-25
.NET Core - sort out NUnit shim over xUnit.net


jonorossi:
1. Does the NUnitLite that recently shipped provide what we need for .NET Core?
1. Are we staying with NUnit or moving the whole project to xUnit.net?
1. If we are staying with NUnit we don't need the shim anymore, if we are going to move to xUnit.net, then maybe the shim is a good transition tool to reduce code churn.

jeremymeng:
RE: NUnitLite v3

Currently there's no VS integrated runner or Dnx runner support.  A Dnx console project must be created to run the tests.  See [this for an example](https://github.com/StackExchange/StackExchange.Redis/blob/master/StackExchange.Redis.Tests/Program.cs). Unfortunately it doesn't seem to support `-nologo`, `-noshadow`, or `-xml` options.

jonorossi:
The NUnit team have an issue logged (https://github.com/nunit/nunit/issues/927) to add a DNX runner, however I suspect with the state of flux of DNX/CLI especially around commands they'll be waiting for it to stabilise.

I'm happy to leave this issue to later (even post-4.0 if needed) since the shim I made is working pretty well for now. @jeremymeng do you see any problem with us releasing 4.0 without this resolved?

jeremymeng:
> do you see any problem with us releasing 4.0 without this resolved?

No. It should be good as long as there are test coverages.

jonorossi:
Looks like many people are making good use of the new `dotnet-test-nunit` package, and it should be [updated to RTM bits within a day](https://github.com/nunit/nunit/issues/575#issuecomment-229175578).

http://www.alteridem.net/2016/06/18/nunit-3-testing-net-core-rc2/

jeremymeng:
@jonorossi switching to `dotnet-test-nunit` is super easy. It just worked!  Upgrading desktop projects to NUnit v3 is a little more involved.  Let me know if you want to change back to NUnit.  I have a commit https://github.com/jeremymeng/Core/commit/e08847ad8d8f5601cf42374c76f20cde3cedbe69 and could submit a PR. I haven't removed the xunit shim.

jeremymeng:
Haven't tested on Mono yet.

jeremymeng:
Tests run and pass on Mono.  Only difference is that now the following three `[Explicit]` tests are marked as ignored by NUnit3.
```
[...]

PR #198, opened 2016-07-08
Update to NUnit v3.4.0

jeremymeng:
The .NET Core test project also switches to use dotnet-test-nunit. There
are some breaking changes in NUnit v3. Notable ones:

- [TestFixtureSetUp] => [OneTimeSetUp].
- No longer set working directory to the directory containing the test
  assembly.
- Require v3 runner which has different command-line options.

Also copy file1.txt file to output directory for .NET Core test project.
The full .NET framework test project does the same.

For #127


jonorossi:
@jeremymeng I've reimplemented these changes and merged them in #202. This PR was out of date because you included changes to NLog binaries in it that were also in the other logging PR.

Other changes I made include:
* Kept NUnit binaries in the directory it was in before rather than moving it to a directory with a version in the name
* You had some single dash rather than double dash command args to nunit3-console
* Only including the binaries that were part of the console runner package, rather than all those extra files you had like NUnit's own unit tests and test files, and other stuff
* Didn't find a need to change BasePEVerifyTestCase to use TestContext
* Suppressed that warning as error from the obsolete attribute because ReSharper doesn't support the new attribute yet
* Fixed some MSBuild warnings with assembly references

jonorossi:
Many thanks for doing the hard work getting the PR together.

Issue #207, opened 2016-07-21
Set-VersionInfo.ps1 Not Invoked


chrislarabell:
When trying to add the pre-release to a project, I received this error:

Assembly 'Castle.Facilities.Logging' with identity 'Castle.Facilities.Logging, Version=3.3.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc' uses 'Castle.Core, Version=3.3.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc' which has a higher version than referenced assembly 'Castle.Core' with identity 'Castle.Core, Version=0.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc'

chrislarabell:
Added change to build script #208

jonorossi:
@chrislarabell thanks, however this has already been fixed in #206.

We don't want `build.cmd` running `Set-VersionInfo.ps1` because it modifies source files which developers shouldn't commit and would just get annoying, this script is run by our CI server, it just happened I stuffed up the .NET 4.5 TeamCity configuration when setting it up after merging the .NET Core branch.

PR #208, opened 2016-07-21
Updated to invoke the Set-VersionInfo.ps1 script.

chrislarabell:
Added the Set-VersionInfo.ps1 invocation in the build.cmd script.


jonorossi:
Closed as per #207.

Issue #200, opened 2016-07-11
Upgrade NLog to 4.5 RTM for .NET Core support


jonorossi:
We are currently using NLog 4.4.0-beta13 to get .NET Core support, we need to upgrade to an RTM version once it is released.

jonorossi:
The main NLog .NET Core issue for 4.4 that hasn't been closed is https://github.com/NLog/NLog/issues/641.

jonorossi:
New main issue for .NET Core support in NLog: https://github.com/NLog/NLog/issues/1559

jonorossi:
The NLog guys have gone quiet since July so it might be time to drop NLog 4.4 beta support with .NET Core support from Castle Core 4.0 and revert back to NLog 4.3.x. It'll unblock us from not being able to release a 4.0 final.

Thoughts?

jonorossi:
Looks like NLog guys have bumped their betas support .NET Standard from 4.4.x to 5.0.x.

DixonDs:
Are there any estimates when Castle.Core 4.0 RTM will be released? We have a couple of libraries depended on it, which we keep as prerelease mainly because they have prerelease dependency Castle.Core

jonorossi:
@DixonDs are you using NLog? This issue is about supporting NLog on .NET Core.

DixonDs:
@jonorossi Sorry, that was probably a wrong issue to ask that question. I've just got an impression that it is the only outstanding issue that holding Castle.Core 4.0 release. Do you happen to have gitter.im channel?

jonorossi:
@DixonDs no problem, nope no active chat room. I just added a log4net issue into the milestone too. I've just been swamped for the last few weeks so pull requests on these last two remaining issues would allow me to cut the release.

thomaslevesque:
Hi @jonorossi, do you plan to release the logging integration packages along with the main library for 4.0? From what I understand, it might be a while before NLog is ready for .NET Core, and none of the integration projects have been migrated to .NET Core yet. It might be a good idea to release Castle.Core 4.0 alone for now, and release the integration packages later. What do you think?

(If I seem eager to see Castle.Core 4.0 released, it's because I am :wink:... it's about the only thing keeping us from releasing FakeItEasy 3.0)

jonorossi:
@thomaslevesque yes, logging integration packages will be released as they have been in the betas. As per my comment on Nov 15th, this issue is only still in the milestone until we revert to a stable NLog for .NET 4.x.

[...]

PR #219, opened 2016-10-13
Fix attribute comparison for ProxyGenerator cache

### LINKED ISSUE #77: ProxyGenerator cache does not take into account AdditionalAttributes

thomaslevesque:
Fixes #77 

At this point, this PR is mostly a proof of concept; I didn't try to preserve the existing API for backwards compatibility, I just changed the type of `AdditionalAttributes` from `CustomAttributeBuilder` to a new `CustomAttributeInfo` class. This class has the same constructors as `CustomAttributeBuilder`, and overrides `Equals` and `GetHashCode`. I also introduce a `CustomAttributeInfo.FromExpression` method to make it easier to define attributes:

    var info = CustomAttributeInfo.FromExpression(
        () => new MyAttribute("foo") { Value = 42, Type = typeof(int) });
    options.AdditionalAttributes.Add(info);


jonorossi:
@thomaslevesque thanks, it looks good, sorry for the slow response to review this. I do like FromExpression, much better than the now obsolete code that pulls apart an actual Attribute instance to try rebuilding it.

Now to discuss whether this should be a breaking change. Looking at the 3 main mocking libraries only FakeItEasy uses AdditionalAttributes. You guys expose `CustomAttributeBuilder` via
[`IFakeOptions.WithAdditionalAttributes`](https://github.com/FakeItEasy/FakeItEasy/blob/dc218f8a2676c1f9864d4ce2d36d3424bd86d48b/src/FakeItEasy/Creation/IFakeOptions.cs#L38), are you happy for that to be a breaking change to FakeItEasy and are you comfortable exposing a DynamicProxy type on your `IFakeOptions` interface? If you are ILMerging you'll also have to expose that type.

`CustomAttributeInfo` works, however this all makes me wonder whether we should just side step this and inspect `CustomAttributeBuilder.m_con` (`ctor` on mono) and `CustomAttributeBuilder.m_blob` (`data` on mono) using reflection. It'll be a faster comparison, however the IL blob will preserve property order from the code which is probably want we want here anyway. We already use reflection to access many [public members of the BCL types](https://github.com/castleproject/Core/tree/master/src/Castle.Core/DynamicProxy/Tokens). Both options aren't ideal, would love to hear your (and anyone else's) thoughts.

thomaslevesque:
> are you happy for that to be a breaking change to FakeItEasy

I don't mind the breaking change. It's not a very widely used feature, so it shouldn't affect many users, and the fix will be pretty easy anyway. @adamralph, @blairconrad, what do you think?

> and are you comfortable exposing a DynamicProxy type on your IFakeOptions interface? If you are ILMerging you'll also have to expose that type.

That's a good question, I hadn't thought about that. We already expose part of the DynamicProxy API as a result of using ILMerge, but not directly as part of the FakeItEasy API. I'd rather avoid that if possible, but I'm not sure how. I guess we could create yet another `CustomAttributeBuilder`-like type that would be specific to FakeItEasy, but it's not a very appealing option. Or we could expose only a method that accepts an `Expression<Func<Attribute>>`.



> CustomAttributeInfo works, however this all makes me wonder whether we should just side step this and inspect CustomAttributeBuilder.m_con (ctor on mono) and CustomAttributeBuilder.m_blob (data on mono) using reflection. It'll be a faster comparison, however the IL blob will preserve property order from the code which is probably want we want here anyway. We already use reflection to access many public members of the BCL types. Both options aren't ideal, would love to hear your (and anyone else's) thoughts.

I guess that would work too, and it would have the benefit of avoiding the breaking change entirely. Of course, as always when using reflection on private members,  there's the risk of the internals being changed in a BCL update, but I don't think it's very likely.

thomaslevesque:
Slightly off-topic, but related: while working on this I noticed the obsolete `AttributesToAddToGeneratedTypes` property. Do you intend to remove it in 4.0?

blairconrad:
I haven't looked at the changes in depth, but in general it looks good to me.
I love `FromExpression` (surprised no test coverage for it, though).

> I don't mind the breaking change. It's not a very widely used feature, so it shouldn't affect many users, and the fix will be pretty easy anyway. @adamralph, @blairconrad, what do you think?

Well worth it, in my opinion. I'd go for the breaking change.

> Or we could expose only a method that accepts an `Expression<Func<Attribute>>`

Yes! This. 😻

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,238 @@
...
+			foreach (var kvp in x)
+			{
+				object value;
+				if (!y.TryGetValue(kvp.Key, out value))
+					return false;
+				if (!Equals(kvp.Value, value))
+					return false;
+			}
+			return true;
+		}
```
blairconrad: returns false if x has members not in y, but not the other way 'round, no?

jonorossi:
> Slightly off-topic, but related: while working on this I noticed the obsolete `AttributesToAddToGeneratedTypes` property. Do you intend to remove it in 4.0?

Yep, it was marked obsolete in 2009 so it can go, I actually wrote about it in my last comment then removed the mention to avoid making things more complicated for now. If it gets in your way then remove it, otherwise lets do it separate and update our changelog.

> I guess that would work too, and it would have the benefit of avoiding the breaking change entirely. Of course, as always when using reflection on private members, there's the risk of the internals being changed in a BCL update, but I don't think it's very likely.

We can add a `FromExpression` method either way, the remaining question is do we add `CustomAttributeInfo` or inspect the privates of `CustomAttributeBuilder`. I agree that the privates are unlikely to change (however different between the CLR and Mono) which is why I would accept the risk, however after another think/look at DP there is no precedent for accessing privates that are not DP's own or generated by DP (so under its control). I'm heavily leaning towards your `CustomAttributeInfo` to avoid adding tech debt that someone else has to sort out in the future, and we are about to release a major version anyway so it's a perfect time.

I think we should turn your PoC into a PR ready to merge. i.e. some unit tests, update the changelog and docs.

jonorossi:
> If it gets in your way then remove it, otherwise lets do it separate and update our changelog.

Actually, let's remove `AttributesToAddToGeneratedTypes` in this PR too, two birds one stone thingy.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
jonorossi: I think most of the comparisons can be short-circuited by checking the length of all the collections before even checking any contents, that would fix this problem.

thomaslevesque:
Thanks for the feedback guys! I'll make the requested changes.

> I love FromExpression (surprised no test coverage for it, though).

I'll add some tests, but I didn't want to waste time doing it until the idea was approved :wink:

blairconrad:
> We can add a `FromExpression` method either way,

I was also thinking that we _could_ do the same for FakeItEasy even before this PR makes it into a prime-time Castle.Core. It would mean duplicate work that would later be ripped out, but could provide a gentler introduction to the new way to specify attributes. Our clients could have a chance to convert before we make a breaking change.

jonorossi:
> I was also thinking that we could do the same for FakeItEasy even before this PR makes it into a prime-time Castle.Core.

Up to you, however I plan to release 4.0.0-beta2 after we get this merged. Need to sort out #200, then plan to ship a final release very soon.

thomaslevesque:
I realized my implementation was incomplete, as it didn't properly support arrays as attribute arguments (attributes support primitive types, `Object`, `Type`, enums and single dimensional arrays of those types, see §17.1.3 in the C# specs). I updated the code to support this.

I also removed `AttributesToAddToGeneratedTypes` and related code (e.g. `AttributeDisassembler` seemed to exist only for this scenario, so I removed it as well).

I still need to add unit tests, but that's all for today.

BTW I'm having a hard time working on the Castle Core code base. The non-.NET Core solution doesn't build in Visual Studio 2015 because of the project.json files. The .NET Core solution builds, but running the tests in the IDE doesn't work (tried with the VS runner and the R# runner). Does it work for you @jonorossi ?

jonorossi:
> BTW I'm having a hard time working on the Castle Core code base. The non-.NET Core solution doesn't build in Visual Studio 2015 because of the project.json files. The .NET Core solution builds, but running the tests in the IDE doesn't work (tried with the VS runner and the R# runner).

I can't find the .NET CLI issue right now but the updated VS2015 MSBuild targets files break the csproj projects when there is a project.json file in the same directory, so I've been using VS2013, I was actually using VS2013 until someone else reported the problem a while back. My suggestion either use VS2013 or temporarily delete project.json, it is all short lived so I'm not interested in hacking our csproj files to workaround the defect. I use ReSharper's test runner in VS2013 for the .NET Framework projects, and the VS test runner in VS2015 for the .NET Core project.

adamralph:
Did we make a decision about the FIE API? I don't think we should expose a Castle.Core type in our public API. We don't currently, and effectively Castle.Core is a hidden implementation detail. We exclude the three types from internalisation when IL merging, but that's only due to the mechanics of Castle.Core. We don't expose those types in our API.

jonorossi:
> Did we make a decision about the FIE API?

Obviously that is for you guys to ultimately decide, but I liked @thomaslevesque suggestion and that was enough of a solution for me to proceed knowing you guys can continue to keep DynamicProxy an implementation detail:

> Or we could expose only a method that accepts an `Expression<Func<Attribute>>`.

blairconrad:
@adamralph, I would also like to
> expose only a method that accepts an `Expression<Func<Attribute>>`.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,294 @@
...
+	{
+		// Cached empty arrays to avoid unnecessary allocations
+		private static readonly PropertyInfo[] EmptyProperties = new PropertyInfo[0];
+		private static readonly FieldInfo[] EmptyFields = new FieldInfo[0];
+		private static readonly object[] EmptyValues = new object[0];
+
+		private static readonly IEqualityComparer<object> ValueComparer = new AttributeArgumentValueEqualityComparer();
+
+		private readonly CustomAttributeBuilder builder;
+		private readonly ConstructorInfo con;
```
blairconrad: A minor thing, but I don't love "con" as a name. Perhaps its well known, it did slow me down a little. Consider "constructor"?

thomaslevesque:
> My suggestion either use VS2013 or temporarily delete project.json

Thanks for the suggestion. I no longer have VS2013, so I've been deleting the project.json and project.lock.json.

> Did we make a decision about the FIE API?

Not a formal decision, but I think we all agree about not exposing a Castle Core type in our API. My preference would be to just expose the expression-based API, i.e.:

    A.Fake<IFoo>(o => o.WithAdditionalAttribute(() => new MyAttribute(42, "blah")));

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,294 @@
...
+				}
+
+				newExpression = memberInitExpression.NewExpression;
+
+				foreach (var binding in memberInitExpression.Bindings)
+				{
+					var assignment = binding as MemberAssignment;
+					if (assignment == null)
+					{
+						throw new ArgumentException("Only assignments bindings are supported");
```
blairconrad: "assignment bindings"

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,294 @@
...
+						foreach (var expr in newArrayExpr.Expressions)
+						{
+							object value = GetAttributeArgumentValue(expr, false);
+							array[index] = value;
+							index++;
+						}
+						return array;
+					}
+				}
+				throw new ArgumentException("Only constant and single-dimensional array expressions are supported");
```
blairconrad: We may reach this line in a case where a single-dimensional array expression is _not_ supported. (When `allowArray` is `false`.)
I'd consider reworking the flow to avoid a potentially misleading exception message.
While I'm commenting on it, if such a rework avoided a boolean parameter, I'd consider it an improvement. I never know what these mean. Way back on line 113, I was all '"true"? What's "true" mean?".
But it's not such a huge deal, given that it's a private method and we can find the answer pretty quickly.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		public CustomAttributeInfo(
+			ConstructorInfo con,
+			object[] constructorArgs)
+			: this(con, constructorArgs, EmptyProperties, EmptyValues, EmptyFields, EmptyValues)
+		{
+		}
+
+		public static CustomAttributeInfo FromExpression(Expression<Func<Attribute>> expression)
```
blairconrad: I'm sure this is perfectly correct, but it would be nice to see it exercised at least once in the tests.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
thomaslevesque: I don't love it either, but I chose it because it's the name used in the `CustomAttributeBuilder` constructor, so it seemed the natural choice since `CustomAttributeInfo` is meant to be a "clone" of that class. But I can change it to something else of course.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
thomaslevesque: oops. I'll fix it, thanks

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
blairconrad: I'm sorry. I must've missed that note. I thought you'd finished it up. What a nag you must think me. Of course now I see the `[WIP]`, which has never left.

thomaslevesque:
OK, I think I'm done now. Added some tests fixed a few issues, and rebased onto master.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
blairconrad: > it can only happen if we find an array inside an array

Ah! Yes. Thanks. I see it now after rereading. And having you tell me. :blush:

>  I can rename it and/or use a named argument on the call site to make things clearer if necessary.

The meaning of the parameter is quite clear on the inside of the method. And I like the name.
For me, a named argument would improve the call sites. I know they're not everyone's [cup of tea](http://idioms.thefreedictionary.com/not+cup+of+tea), though.

An alternative would be a microtype (probably an enum) so we could have explicitly named values at the call sites, but again, maybe that's too much.

In the end, it's a private method, and not hard for anyone who's confused by the calls to look up the meaning of the bool. Thanks for listening.

blairconrad:
FWIW, it looks very good to me. Thanks, @thomaslevesque!

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -107,15 +107,7 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
 
 		public Type BaseTypeForInterfaceProxy { get; set; }
 
-		[Obsolete(
-			"This property is obsolete and will be removed in future versions. Use AdditionalAttributes property instead. " +
-			"You can use AttributeUtil class to simplify creating CustomAttributeBuilder instances for common cases.")]
-		public IList<Attribute> AttributesToAddToGeneratedTypes
```
jonorossi: Can you please update the changelog with a breaking change for two of these properties and removing `IAttributeDisassembler`.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -167,6 +167,7 @@
     <Compile Include="Components.DictionaryAdapter.Tests\IPhoneWithFetch.cs" />
     <Compile Include="Components.DictionaryAdapter.Tests\NameValueCollectionAdapterTests.cs" />
     <Compile Include="DynamicProxy.Tests\AttributesToAvoidReplicatingTestCase.cs" />
+    <Compile Include="CustomAttributeInfoTestCase.cs" />
```
jonorossi: Could you move these unit tests into the DynamicProxy.Tests directory.

src/Castle.Core.Tests/CustomAttributeInfoTestCase.cs
```diff
@@ -0,0 +1,224 @@
...
+					new object[] { 42, "foo", new[] { 1, 2, 3 } }));
+		}
+
+		private static object[] CreateFromExpressionTestCase(Expression<Func<Attribute>> expr, CustomAttributeInfo expected)
+		{
+			return new object[] { expr, expected };
+		}
+	}
+}
+	;
```
jonorossi: There is a stray semicolon at the end here.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,301 @@
...
+			object[] fieldValues)
+		{
+			// Will take care of validating the arguments
+			this.builder = new CustomAttributeBuilder(constructor, constructorArgs, namedProperties, propertyValues, namedFields, fieldValues);
+
+			this.constructor = constructor;
+			this.constructorArgs = constructorArgs;
+
+			this.properties = namedProperties
+				.Zip(propertyValues, (p, v) => new { p.Name, Value = v })
```
jonorossi: `Enumerable.Zip` isn't available on .NET 3.5, it was added in .NET 4.0.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -0,0 +1,301 @@
...
+			PropertyInfo[] namedProperties,
+			object[] propertyValues,
+			FieldInfo[] namedFields,
+			object[] fieldValues)
+		{
+			// Will take care of validating the arguments
+			this.builder = new CustomAttributeBuilder(constructor, constructorArgs, namedProperties, propertyValues, namedFields, fieldValues);
+
+			this.constructor = constructor;
+			this.constructorArgs = constructorArgs;
```
jonorossi: Should we copy the constructorArgs array to make sure this type stays immutable?

src/Castle.Core.Tests/CustomAttributeInfoTestCase.cs
thomaslevesque: Oops, don't know how it ended up here...

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
thomaslevesque: Good point. Should probably do it for the other parameters too.

thomaslevesque: > Should probably do it for the other parameters too.

Scratch that. The other parameters are used to build dictionaries, not kept as is.

thomaslevesque:
@jonorossi I made the requested changes

thomaslevesque:
btw @jonorossi let me know if you want me to squash some or all the commits before the merge. I don't know how you usually handle this in Castle.Core

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
blairconrad: That was my thought at first as well. Then I had a second thought. It'll probably just reveal my inadequate grasp of what's going on here, but may not the other parameters (or even individual args of constructorArgs) be single-dimensional arrays, which could be mutated after the fact?

thomaslevesque:
> may not the other parameters (or even individual args of constructorArgs) be single-dimensional arrays, which could be mutated after the fact?

It doesn't matter for `namedProperties`, `propertyValues`, `namedFields`, and `fieldValues`, because we don't store them; we just create dictionaries from them.

However you have a point about individual args of `constructorArgs`, or even values of array properties or fields. But frankly, at this point, a user who is changing the content of an array passed to CustomAttributeInfo is just looking for trouble... so I'm tempted to leave it alone.

blairconrad:
> we just create dictionaries from them

Which is a way to store them, no? In dictionaries!

```c#
dict.Add(members[i].Name, values[i]);
```

shoves the element of the `propertiesValues` or `fieldValues` right in the dictionary, for later equality-checking. If the element is an array, it can be mutated.

I do agree that anyone changing the hypothetical arrays after the fact is looking for trouble. Then again, when they find it, they'll probably come back with questions and complaints! :wink:
Thanks for considering my comment.

thomaslevesque:
> shoves the element of the propertiesValues or fieldValues right in the dictionary, for later equality-checking. If the element is an array, it can be mutated.

Ah, I thought you were talking about `propertiesValues` themselves `fieldValues`, not the arrays they might contain...

> I do agree that anyone changing the hypothetical arrays after the fact is looking for trouble. Then again, when they find it, they'll probably come back with questions and complaints! 😉

True. @jonorossi, what do you think? Should the code account for that by deep-cloning the arrays? (actually, only nested arrays need to be cloned, since all other types allowed as attribute arguments are immutable)

jonorossi:
> Should the code account for that by deep-cloning the arrays? (actually, only nested arrays need to be cloned, since all other types allowed as attribute arguments are immutable)

`CustomAttributeBuilder` [copies the ctor args](
https://referencesource.microsoft.com/#mscorlib/system/reflection/emit/customattributebuilder.cs,135) (because it holds on to them for some reason) but doesn't clone any constructor argument that is an array so therefore `CustomAttributeBuilder` also isn't fully immutable (however `m_blob` is). It doesn't copy any arrays in property/field values because they are written into the byte array by the end of the constructor call and no reference is stored.

To avoid wasted allocations I think we should leave the code as is, and just add an XML comment to the class indicating that any arrays passed to the class are now owned by that class and should not be modified. I agree this is an edge case and I'm happy to direct defect reports to an XML comment.

jonorossi:
> let me know if you want me to squash some or all the commits before the merge. I don't know how you usually handle this in Castle.Core

I do usually like history in PRs squished if they aren't commits that align to nice clean patches (I'm old school, I like clean patch sets 😉 ), but have got soft over the years as just getting a complete PR seems to be a stretch lately. Many thanks for the work here and being so responsive.

thomaslevesque:
I don't mind squashing into clean commits, that's what we usually do in FakeItEasy.

I squashed the changes down to 4 commits.

jonorossi:
@thomaslevesque just wanted to make sure you didn't miss my reply to your question. Were you going to add that XML comment?

thomaslevesque:
> @thomaslevesque just wanted to make sure you didn't miss my reply to your question. Were you going to add that XML comment?

Actually, I did miss it... I'll add the comment.

thomaslevesque:
I added the XML comment in a fixup commit; I'll squash it after you approve it.

jonorossi:
> I added the XML comment in a fixup commit; I'll squash it after you approve it.

Exactly what I was expecting, just without the spelling mistake 😉 :

> Arrays passed to this class as constructor arguments or **property of field** values become owned by this class.

thomaslevesque:
> Exactly what I was expecting, just without the spelling mistake 😉 :

Oops! Fixed and squashed

jonorossi:
Many thanks @thomaslevesque, I'm ready to merge.

blairconrad:
Thanks, @thomaslevesque and @jonorossi!

thomaslevesque:
Thanks for the merge!

PR #223, opened 2016-11-15
Fix CustomAttributeInfo.FromExpression for VB.NET

### LINKED ISSUE #222: CustomAttributeInfo.FromExpression doesn't work in VB.NET
In the process of integrating 4.0.0-beta002 in FakeItEasy, I realized that the new expression-based API for custom attributes didn't work in VB.NET.

In C#, the body of the lambda expression `() => new MyAttribute()` is just a `NewExpression`, but the equivalent VB.NET lambda expression `Function() New MyAttribute()` introduces a conversion to `Attribute` (i.e. `Convert(NewExpression, Attribute)`). Not sure why, since this conversion doesn't do anything useful...

The fix should be pretty easy, I'll submit a PR ASAP.

thomaslevesque:
Fixes #222

### DIFF
--- a/src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
+++ b/src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
@@ -92,10 +92,12 @@ namespace Castle.DynamicProxy
 			var namedFields = new List<FieldInfo>();
 			var fieldValues = new List<object>();
 
-			var newExpression = expression.Body as NewExpression;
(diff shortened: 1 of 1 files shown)


thomaslevesque:
I would have added unit tests, but creating a VB.NET test project just for this seemed a bit overkill... I did test the fix manually, of course :wink:

jonorossi:
Thanks @thomaslevesque

PR #231, opened 2017-01-28
Ordering mixin constructor arguments by type name

### LINKED ISSUE #230: Constructor arguments vary by alphabetical order
When creating a proxy from two instances, the constructor that will be called with the two instances and the `IInterceptor[]` parameters will have its arguments vary alphabetically on the implemented interfaces, and won't match the expected order hence throwing a "couldn't find parameterless constructor" error. 

Here is the repro:
```
using System;
using Castle.DynamicProxy;

namespace ConsoleApp4
{
    public class Program
    {
        private static readonly ProxyGenerator _proxyGenerator = new ProxyGenerator();

        public static void Main(string[] args)
        {
            options.AddMixinInstance(new Foo());
            options.AddMixinInstance(new Bar());
            var proxy = _proxyGenerator.CreateClassProxy(typeof(Foo), options);

            Console.WriteLine(proxy);
        }
    }

    public interface IFoo1
    {
    }

    public class Foo : IFoo1
    {
    }
    
    public interface IFoo2
    {
    }
    
    public class Bar : IFoo2
    {
    }
}
```

This will throw an exception. Now if you rename `IFoo1` to `IFoo3` which is alphabetically after `IFoo2` the constructor won't be found (the order of the `constructorArguments` and the actual ctor method don't match anymore)

Environment: .NET Core 1.1 with Castle.Core 4.0.0

Maybe related to: #112

sebastienros:
Fixes #230


src/Castle.Core/DynamicProxy/Generators/ClassProxyGenerator.cs
```diff
@@ -74,7 +74,9 @@ protected virtual Type GenerateType(string name, Type[] interfaces, INamingScope
 				// TODO: redo it
 				if (contributor is MixinContributor)
 				{
-					constructorArguments.AddRange((contributor as MixinContributor).Fields);
+                    // constructor arguments are ordered by interface name as the MixinsData does
+                    var orderedFields = (contributor as MixinContributor).Fields.OrderBy(x => x.Fieldbuilder.FieldType.Name);
```
jonorossi: I'd much prefer `MixinContributor` to do the sorting rather than a proxy generator, what do you think about changing the `fields` field to a `SortedDictionary<Type, FieldReference>`?

src/Castle.Core.Tests/BugsReported/DynProxy230.cs
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests.BugsReported
+{
+    using System;
+
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class DynProxy230 : BasePEVerifyTestCase
```
jonorossi: Could you please move this test into `src/Castle.Core.Tests/Mixins` with a descriptive name, that `BugsReported` directory has tests named after old JIRA issues from way back.

sebastienros:
Updated

- I can see that my little refactoring on `ClassProxyGenerator.cs` also updated some white spaces because of the customized VS settings. I can revert the whole file changes if you prefer.
- Not sure the tests location is what you meant.

jonorossi:
Thanks for the fixes, unfortunately I've got a few more for you:

1. Yes, that is where I wanted to unit tests even though it currently only has mixin classes not actual tests, however you missed updating the namespace of the test class and suffixing the class with `TestCase`.
1. I assume you are developing with the .NET Core solution, however our .NET Framework builds use the csproj files so `FieldReferenceComparer.cs` and `ConstructorArgumentsOrderWithMixins.cs` need to be added to the two csproj files.
1. We still support .NET 3.5 so you can't use the `nameof` operator in `FieldReferenceComparer`.
1. You'll also need to mark your new mixin classes with `[Serializable]` (wrapped with `#if FEATURE_SERIALIZATION`) like the existing ones, because for .NET Framework unit tests (inside `BasePEVerifyTestCase`) we save the assembly to disk which serializes the generated types so that we can run `peverify` on the assembly.

jonorossi:
@sebastienros have you had a chance to take a look at my second lot of feedback?

jonorossi:
Closing in preference to #269 where I've fixed the review comments.

PR #236, opened 2017-03-12
Add IsAccessible(method) that provides a message when not accessible

blairconrad:
Fixes #235.

I may not have put the code in the best places.
Note that only the "references Castle.Core/strong-named" path is tested in the unit tests. If you'd like, I can add tests for the other paths, but believe they would require additional test projects.
I have performed manual testing for the other paths.

Thanks for considering this PR. I'm happy to make changes you might like.

PR #237, opened 2017-03-12
Deal with untracked files

blairconrad:
While working on #236 I noticed some untracked files sitting around after a build.
One was easy to deal with as an ignore, but the .NET Core test results looked to me to be going to the wrong directory, so I changed the test run to put them in a path analogous to their NET40, NET45, … siblings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ src/**/obj/
 src/**/bin/
 project.lock.json
 /.dnx/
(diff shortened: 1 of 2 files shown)


jonorossi:
Ah yes, `.dnx` being renamed to `.dotnet`. Thanks.

blairconrad:
Thanks, @jonorossi!

Issue #241, opened 2017-04-09
Move builds from TeamCity to AppVeyor and Travis CI


jonorossi:
These are the build configurations we've currently have in TeamCity with the configuration pulled apart from the set up. It would be great if they all went to using a `build` script rather than calling msbuild/xbuild directly.

I'm not sure exactly what we want to do but it might be time to move Castle Core to the new VS2017 tooling to support .NET Framework and .NET Core, this should hopefully make the NuGet packaging cleaner.

TeamCity: http://builds.castleproject.org/project.html?projectId=Core&tab=projectOverview
AppVeyor: https://ci.appveyor.com/project/castleproject/core
Travis CI: https://travis-ci.org/castleproject/Core

### .NET 3.5
```
RELEASE_BUILD=false
VERSION_SUFFIX=
$ powershell buildscripts\Set-VersionInfo.ps1
$ msbuild /p:Configuration=NET35-Release /t:Package buildscripts/Build.proj
```

### .NET 4.0
- Same as 3.5 with `NET40-Release` configuration

### .NET 4.5
- Same as 3.5 with `NET45-Release` configuration

### .NET Core
 ```
RELEASE_BUILD=false
VERSION_SUFFIX=
$ powershell buildscripts\Set-VersionInfo.ps1
$ build NETCORE
publish src\Castle.Core\bin\Release\netstandard1.3
publish src\Castle.Services.Logging.NLogIntegration\bin\Release\netstandard1.3
publish src\Castle.Services.Logging.SerilogIntegration\bin\Release\netstandard1.3
```

### Mono (4.6.1)
```
RELEASE_BUILD=false
VERSION_SUFFIX=
$ xbuild /p:Configuration=NET45-Release /t:RunAllTests buildscripts/Build.proj
import nunit results build/NET45/NET45-Release/bin/test-results/nunit-results.xml
publish build/NET45/NET45-Release/bin/**/*
```

### Pack
```
RELEASE_BUILD=%reverse.dep.*.env.release_build%
VERSION_SUFFIX=%reverse.dep.*.env.version_suffix%
$ powershell buildscripts\Set-VersionInfo.ps1
nuget pack
- spec:
  buildscripts/Castle.Core.nuspec
  buildscripts/Castle.Core-log4net.nuspec
  buildscripts/Castle.Core-NLog.nuspec
  buildscripts/Castle.Core-Serilog.nuspec
- properties:
  Configuration=Release
  CastleCoreVersion=%ReleaseVersion%
[...]

PR #242, opened 2017-04-11
Refactor build

fir3pho3nixx:
Just a check before I move on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ build/
 artifacts/
 src/**/obj/
 src/**/bin/
+buildscripts/**/obj/
+buildscripts/**/bin/
 project.lock.json
 /.dnx/
 /.dotnet/
--- a/Castle.Core-VS2017.sln
+++ b/Castle.Core-VS2017.sln
@@ -1,0 +1,71 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26403.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Build", "Castle Build", "{1B999D24-B7AB-4997-96E7-08FA05325694}"
+	ProjectSection(SolutionItems) = preProject
+		.gitattributes = .gitattributes
+		.gitignore = .gitignore
+		build-vs2017.cmd = build-vs2017.cmd
+		CHANGELOG.md = CHANGELOG.md
+		CONTRIBUTING.md = CONTRIBUTING.md
+		LICENSE = LICENSE
+		README.md = README.md
+	EndProjectSection
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BuildScripts", "buildscripts\BuildScripts.csproj", "{2E8BA116-1A54-425D-BDEF-6A292D965675}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Core-VS2017", "src\Castle.Core\Castle.Core-VS2017.csproj", "{C31C55D2-32FD-4158-8EC5-0FBB56DCE4D4}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Core.Tests-VS2017", "src\Castle.Core.Tests\Castle.Core.Tests-VS2017.csproj", "{F32D049E-C486-4F29-BAD5-FC5EA06E7B88}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Services.Logging.log4netIntegration-VS2017", "src\Castle.Services.Logging.log4netIntegration\Castle.Services.Logging.log4netIntegration-VS2017.csproj", "{AFD3B071-E971-499B-A95B-E98155EB66E8}"
+EndProject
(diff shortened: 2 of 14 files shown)


Castle.Core-VS2017.sln
```diff
@@ -0,0 +1,71 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
jonorossi: Wow, I just assumed Microsoft would have improved the solution file with the changes to project files.

buildscripts/BuildScripts.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net462</TargetFrameworks>
```
jonorossi: Ignoring .NET 3.5/4.0, we currently support .NET 4.5+, is this temporary too?

buildscripts/common.props
```diff
@@ -0,0 +1,21 @@
...
+
+  <PropertyGroup Condition="'$(Configuration)'=='Debug'">
+    <DefineConstants>TRACE;DEBUG</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)'=='net40'">
+    <DefineConstants>NET40;FEATURE_APPDOMAIN;FEATURE_SYSTEM_CONVERTER;FEATURE_LEGACY_REFLECTION_API;FEATURE_ISUPPORTINITIALIZE;FEATURE_DICTIONARYADAPTER_XML;FEATURE_BINDINGLIST;FEATURE_LISTSORT;FEATURE_IDATAERRORINFO;FEATURE_SERIALIZATION;FEATURE_GET_REFERENCED_ASSEMBLIES;FEATURE_SECURITY_PERMISSIONS;FEATURE_REMOTING;FEATURE_GAC;FEATURE_SYSTEM_CONFIGURATION;FEATURE_ASSEMBLYBUILDER_SAVE</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)'=='netstandard1.5'">
```
jonorossi: Any reason you are now targeting 1.5 rather than netstandard1.3?

buildscripts/common.props
```diff
@@ -0,0 +1,21 @@
+<Project>
+  <PropertyGroup>
+    <VersionPrefix>5.0.0</VersionPrefix>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/cryosharp/fortress</RepositoryUrl>
```
jonorossi: URL

src/Castle.Core/Compatibility/IntrospectionExtensions.cs
```diff
@@ -16,7 +16,8 @@
 
 namespace System.Reflection
 {
-	internal static class IntrospectionExtensions
+	// GVDM: Changing protection levels here! Castle.Services.Compatibility?
+	public static class IntrospectionExtensions
```
jonorossi: Why does this need to be public now? It actually can't be made public, it'll break all consumers of our library because of namespaces.

buildscripts/BuildScripts.csproj
fir3pho3nixx: Changed this to net45 all over.

buildscripts/common.props
fir3pho3nixx: Sorry!

buildscripts/common.props
fir3pho3nixx: All moniker sorted for now

fir3pho3nixx:
This should be safe to merge if you want. I would rather interactive rebase, squish and clean the commit's as I go. Cherry picking might have unintended consequences. 

Just let me add the netstandard moniker and get the build flags for that sorted.

jonorossi:
When were you planning to add the appveyor.yml file? I assumed it would be in PR1 based on your list in #241.

.gitignore
```diff
@@ -10,6 +10,8 @@ build/
 artifacts/
 src/**/obj/
 src/**/bin/
+buildscripts/**/obj/
+buildscripts/**/bin/
```
jonorossi: I tried working out what causes binaries to be written here but I'm lost.

fir3pho3nixx: If you open the solution up in VS2017, you will see that I have included all the build scripts/config inside the solution. Thought this would be handy for doing build work, rather than diving out to the file system but it needs gitignores for bin/obj output. 

solution-layout

fir3pho3nixx:
Have basic appveyor checked in, let me know if you need anymore changes.

.gitignore
jonorossi: Interesting, so it sort of works like another project rather than the old solutions items.

appveyor.yml
```diff
@@ -0,0 +1,15 @@
+version: 4.0.{build}
+
+image: Visual Studio 2017
+
+branches:
+
+  except:
+  
+    - gh-pages
```
jonorossi: We don't have a gh-pages branch.

fir3pho3nixx:
Great, now we can continue our discussion on the issue about how we tackle testing. I have a few questions on this, also we should see a build kick off in appveyor and my next PR should run a build status check. You can then update the project settings to only allow PR's that pass all status checks. Administrators can override this.

jonorossi:
> Great, now we can continue our discussion on the issue about how we tackle testing. I have a few questions on this

Fire away

> also we should see a build kick off in appveyor and my next PR should run a build status check

Yep, I could only get it to work by changing the default branch to refactor-build, it didn't want to automatically start a build on any branch.

> You can then update the project settings to only allow PR's that pass all status checks. Administrators can override this.

Yep, saw those options along with the protected branches, will have a think about how best to use them to get you pushing to the organisation repo rather than a fork.

.gitignore
fir3pho3nixx: Yes, it is using the new msbuild structure, so for anyone inclined to dive into the file system and add stuff here, it will be auto discovered by this project. 

Lovely for reducing friction when doing build work. Will eventually also double up as a candidate for housing custom msbuild tasks (let's hope that does not happen just yet).

PR #243, opened 2017-04-13
Fixed some failing tests

fir3pho3nixx:
Missing embedded resources within Core and some test resources were not being copied to the output directory.

### DIFF
--- a/src/Castle.Core.Tests/Castle.Core.Tests-VS2017.csproj
+++ b/src/Castle.Core.Tests/Castle.Core.Tests-VS2017.csproj
@@ -9,8 +9,12 @@
     <AssemblyOriginatorKeyFile>..\..\buildscripts\CastleKey.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <EmbeddedResource Include="Core.Tests\Resources\file1.txt" />
-    <EmbeddedResource Include="Core.Tests\Resources\MoreRes\file2.txt" />
+    <EmbeddedResource Include="Core.Tests\Resources\file1.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </EmbeddedResource>
+    <EmbeddedResource Include="Core.Tests\Resources\MoreRes\file2.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NUnit" Version="2.6.4" />
--- a/src/Castle.Core/Castle.Core-VS2017.csproj
+++ b/src/Castle.Core/Castle.Core-VS2017.csproj
@@ -10,6 +10,14 @@
     <AssemblyOriginatorKeyFile>..\..\buildscripts\CastleKey.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>True</SignAssembly>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <None Remove="DynamicProxy\DynProxy.snk" />
+  </ItemGroup>
+  
+  <ItemGroup>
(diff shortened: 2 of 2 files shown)


src/Castle.Core.Tests/Castle.Core.Tests-VS2017.csproj
```diff
@@ -9,8 +9,12 @@
     <AssemblyOriginatorKeyFile>..\..\buildscripts\CastleKey.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <EmbeddedResource Include="Core.Tests\Resources\file1.txt" />
-    <EmbeddedResource Include="Core.Tests\Resources\MoreRes\file2.txt" />
+    <EmbeddedResource Include="Core.Tests\Resources\file1.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </EmbeddedResource>
+    <EmbeddedResource Include="Core.Tests\Resources\MoreRes\file2.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
jonorossi: They are embedded resources, I'm not sure how copying them to the output directory fixes the problem.

https://github.com/castleproject/Core/blob/master/src/Castle.Core.Tests/project.json#L51-L55

src/Castle.Core/Castle.Core-VS2017.csproj
```diff
@@ -10,6 +10,14 @@
     <AssemblyOriginatorKeyFile>..\..\buildscripts\CastleKey.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>True</SignAssembly>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <None Remove="DynamicProxy\DynProxy.snk" />
```
jonorossi: Is the none remove actually needed, I'd have thought you'd only be able to have a file with one type or you'd have to make it `<Content Remove=...`.

If it has to stay probably best to put them in the same ItemGroup to make it obvious.

fir3pho3nixx:
That is really unfortunate that the checks are not kicking in. Let's get tests passing(will look deeper into this) and see if we can merge it back to the default branch. A good test!

src/Castle.Core.Tests/Castle.Core.Tests-VS2017.csproj
fir3pho3nixx: ``csharp

[Test]
		public void CreateWithRelativePath()
		{
			IResource resource = resFactory.Create( new CustomUri(basePath + "/file1.txt") );

			Assert.IsNotNull(resource);
			String line = resource.GetStreamReader().ReadLine();
			Assert.AreEqual("Something", line);
		}

```

Looked like it was trying to open the file off the filesystem and not using an embedded resource. Happy to be wrong here. Also looks like you are not copying anything out in your project.json. 

The old csproj however does copy the file out: 

```xml

 <EmbeddedResource Include="Core.Tests\Resources\file1.txt">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </EmbeddedResource>


```

src/Castle.Core/Castle.Core-VS2017.csproj
fir3pho3nixx: The remove is totally unnecessary. I am having a grand old time trying to keep this tidy. VS create new item groups of it's own random choosing. Some of the rules around how it maintains these files are a little `rough round the edges`.

src/Castle.Core.Tests/Castle.Core.Tests-VS2017.csproj
jonorossi: Makes sense now, `AssemblyResourceFactoryTestCase` and `FileResourceFactoryTestCase` both use those files from different places.

My guess for why it works with the project.json tooling is that the `Directory.GetCurrentDirectory()` returns where the project.json file is, MSBuild stuff always runs out of bin.

PR #256, opened 2017-04-30
Refactor build vnext

fir3pho3nixx:
This PR addresses the following: 

 - VS2017 project structure now allows for OS specific targeting(IsWindows, IsOSX, IsLinux). This was important because there are cases where you want to excluded certain things around strong naming on Linux without having to add more build flags.

 - DotNetCore tests are compiling and running on Ubuntu(16.04 LTS). This has also been enabled in the TravisCI build. You will notice less tests running because of exclusions around the strong naming of assemblies. If this comes back in subsequent versions of netstandard we just need to revert the exclusions in the MsBuild file which are targeting Linux. This project now supports Visual Studio Code development workflows.

 - DotNetCore tests are compiling and running on Windows along with separate exports of test results to appveyor. Similar to what was done for Windsor by Alina. Think I solved the tests overwriting each other problem by varying the output name for the assembly being generated. Will double check this test counts in TeamCity. 

Things I tried but doubled back on:

 - I completely reverse engineered the mono build for omnisharp and I did conclude I could get this working after rebuilding my corrupted linux box at home. The problem I had with doing this was that the CAKE scripts were downloading what appeared to be 'manually zipped' portions of the mono msbuild scripts and blending them into the dotnetcli/sdk folder to get round pathing issues. I abandoned this approach at that point because I did not think this was something we wanted. I recommend we wait a bit and see how things progress.


fir3pho3nixx:
Start reviewing from here: https://github.com/castleproject/Core/pull/256/commits/9a25135346cd3622f0ddf3cc65ac39ae5127b451

jonorossi:
> Start reviewing from here: 9a25135

You've rebased incorrectly. Compare the following two commits for example, the first is your most recent commit that has been merged into the refactor-build branch:

- https://github.com/castleproject/Core/commit/d6fb9ff6956a6413cbee19908d7e07f06ee5de80
-
 https://github.com/castleproject/Core/pull/256/commits/003f66ed328d8ad4e7390c9f397d9a3732b532bb

jonorossi:
> VS2017 project structure now allows for OS specific targeting(IsWindows, IsOSX, IsLinux). This was important because there are cases where you want to excluded certain things around strong naming on Linux without having to add more build flags.

I think I'm really missing something here, does .NET Core on Linux not support strong naming at all? Isn't all of corefx strong named? If we do this then we can't use the same assembly for all .NET Standard platforms which really defeats the point, and this would transitively flow into our consumers.

fir3pho3nixx:
When I enable `<SignAssembly>True</SignAssembly>` in the linux build I get the following error:

CSC : error CS7027: Error signing output with public key from file '../../buildscripts/CastleKey.snk' -- Assembly signing not supported. [/home/gavin/code/Core/src/Castle.Core/Castle.Core-VS2017.csproj]

Will see if there is any way of getting this to work.

fir3pho3nixx:
The TravisCI cfg runs the mono tests using the old VS2015 tooling but there is one test failing, please see https://travis-ci.org/fir3pho3nixx/Core#L2236. 

I also cannot get the build script to fail the build(even with set -e). Also something that needs a bit more investigation.

fir3pho3nixx:
Why are the TravisCI status checks not appearing? Strange ...

jonorossi:
> Why are the TravisCI status checks not appearing? Strange ...

The two status checks appear in GitHub's commit log page from e29b5b82e0d6d2d2a07e079f0ed2eb191f0b9511. My guess is that Travis doesn't think this pull request is based on that commit because it isn't, so your rebase is causing the problem. If it still doesn't work after that is fixed up I'll look further into it.

> The TravisCI cfg runs the mono tests using the old VS2015 tooling but there is one test failing, please see https://travis-ci.org/fir3pho3nixx/Core#L2236.

I went to this URL, the log only has 2231 lines, it looks like the unit tests are no longer being run. My guess is that new version of Mono might have broken one of our unit tests.

> I also cannot get the build script to fail the build(even with set -e)

What return code are you getting from running NUnitLite, the docs indicate it is meant to return non-zero however I wonder if that and running it via `dotnet` is all working.

fir3pho3nixx:
Bad rebase plus one too many force pushes here me thinks. Jumping over to: https://github.com/castleproject/Core/pull/256

PR #259, opened 2017-05-09
Deploy on tag

fir3pho3nixx:
Not sure how this gets tested except by trying it out :)

@jonorossi - Once you have merged this, you will need to generate a key and update the appveyor.yml file.   

~~Think one of my pet peeves with this commit is the fact that when you embed a LICENSE file into a NuGet package without an extension, you create a corrupt package? Weird.~~

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,3 @@
-version: 4.0.{build}
-
 image: Visual Studio 2017
 
 branches:
@@ -14,9 +12,16 @@ skip_commits:
     - CHANGELOG.md
     - LICENSE
 
-skip_tags: true
+skip_tags: false
 
 test: off
+
+init:
+- ps: >-
+    if ($env:APPVEYOR_REPO_TAG -eq "true")
+    {
+        Update-AppveyorBuild -Version "$env:APPVEYOR_REPO_TAG_NAME"
+    }
 
 after_build:
 - ps: >-
@@ -26,5 +31,19 @@ after_build:
 
     $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path "DesktopClrTestResults.xml"))
 
+on_success:
+- ps: >-
+    if ($env:APPVEYOR_REPO_TAG -eq "true")
+    {
+        nuget push ".\build\Castle.Core.${env:APPVEYOR_REPO_TAG_NAME}.nupkg" -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push ".\build\Castle.Services.Logging.log4netIntegration.${env:APPVEYOR_REPO_TAG_NAME}.nupkg" -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push ".\build\Castle.Services.Logging.NLogIntegration.${env:APPVEYOR_REPO_TAG_NAME}.nupkg" -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push ".\build\Castle.Services.Logging.SerilogIntegration.${env:APPVEYOR_REPO_TAG_NAME}.nupkg" -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+    }
+
 build_script:
 - cmd: build-vs2017.cmd
+
+artifacts:
+- path: build\*.nupkg
+  name: core
--- a/buildscripts/common.props
+++ b/buildscripts/common.props
@@ -4,8 +4,9 @@
 		<NoWarn>$(NoWarn);CS1591;CS3014;CS3003;CS3001;CS3021</NoWarn>
 		<RepositoryType>git</RepositoryType>
 		<RepositoryUrl>https://github.com/castleproject/Core</RepositoryUrl>
-		<BuildVersion Condition="'$(APPVEYOR_BUILD_VERSION)'==''">0.0.0</BuildVersion>
-		<BuildVersion Condition="'$(APPVEYOR_BUILD_VERSION)'!=''">$(APPVEYOR_BUILD_VERSION)</BuildVersion>
+		<BuildVersion Condition="'$(APPVEYOR_BUILD_VERSION)'==''AND'$(APPVEYOR_REPO_TAG)'!='true'">0.0.0</BuildVersion>
+		<BuildVersion Condition="'$(APPVEYOR_BUILD_VERSION)'!=''AND'$(APPVEYOR_REPO_TAG)'!='true'">$(APPVEYOR_BUILD_VERSION)</BuildVersion>
+		<BuildVersion Condition="'$(APPVEYOR_REPO_TAG)'=='true'">$(APPVEYOR_REPO_TAG_NAME)</BuildVersion>
 		<AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath> 
 	</PropertyGroup>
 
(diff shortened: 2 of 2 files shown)


LICENSE.txt
```diff
@@ -0,0 +1,13 @@
+Copyright 2004-2016 Castle Project - http://www.castleproject.org/
```
jonorossi: As we did in Windsor to work around NuGet not supporting files without an extension, we renamed LICENSE to LICENSE.txt rather than just copy it. Still not sure how we've not noticed this before.

appveyor.yml
```diff
@@ -28,3 +28,19 @@ after_build:
...
+  name: core
+
+deploy:
+- provider: NuGet
+  api_key:
+    secure: Please insert key here
+  skip_symbols: false
+  artifact: core 
+  on:
+      appveyor_repo_tag: true
```
jonorossi: I'm trying to understand how the version is set, the docs (https://www.appveyor.com/docs/branches/) make no reference to version, only `APPVEYOR_REPO_TAG` and `APPVEYOR_REPO_TAG_NAME`.

LICENSE.txt
fir3pho3nixx: Not sure how to read into your comment about noticing. I consciously chose to copy it as we still have a comprehensive clean up to do in a separate PR. So I did not think it was a problem.

appveyor.yml
fir3pho3nixx: Let me check this out. Will play around with fortress and make sure I verify this.

appveyor.yml
```diff
@@ -28,3 +28,19 @@ after_build:
...
 - cmd: build-vs2017.cmd
+
+artifacts:
+- path: build\*.nupkg
+  name: core
+
+deploy:
+- provider: NuGet
+  api_key:
+    secure: Please insert key here
```
jonorossi: It appears from the AppVeyor docs that we could define this with a secure environment variable rather than having to encrypt the API token and store it here, I'd much prefer that if possible.

fir3pho3nixx: Let me do more research on how we achieve this. Never tried it.

LICENSE.txt
jonorossi: Don't worry about my second sentence, just rambling really, wondering if this is a recent NuGet regression.

Happy to go which ever way you'd like, I just thought the history would be cleaner with a "rename".

appveyor.yml
fir3pho3nixx: @jonorossi - you are right. 

There is this ugly ass way of doing it. It would involve querying the version number from the [GitHub Api](https://www.appveyor.com/docs/build-worker-api/#update-build-details). I don't like this because we would have to write a client which involves exposing secure keys. 

Are you happy to conventionalise `APPVEYOR_REPO_TAG_NAME` to be the semver version?

fir3pho3nixx: We would have to test this with pre release (eg 'alpha1') suffixes ...

jonorossi: Definitely agree we don't want to call the GitHub API, would much prefer using `git describe`, however the AppVeyor environment variable seems to work good.

Can we fix AppVeyor setting the version to `0.5.276` when the tag is `0.5.6`, it also seems wrong having `version: 0.5.{build}` for all builds on any branch?

appveyor.yml
jonorossi: Do you know if you can define the NuGet deployment in the web UI with the API key and a deployment condition which adds to the YAML file?

appveyor.yml
fir3pho3nixx: You can control it through the settings tab of the build in appveyor and by removing `version: 0.5.{build}` from `appveyor.yml`.

Ideally we would want this to auto update itself on tag/releases. I am still toying with this idea.

appveyor.yml
fir3pho3nixx: > It appears from the AppVeyor docs that we could define this with a secure environment variable rather than having to encrypt the API token and store it here, I'd much prefer that if possible.

Can you check this out and reconsider: https://www.appveyor.com/docs/build-configuration/#secure-variables

LICENSE.txt
alinapopa: For the LICENSE file issue, the fix is to use this in common.props:
`<Content Include="..\..\LICENSE"> <PackagePath></PackagePath> </Content>
`
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/197

appveyor.yml
jonorossi: I'm not sure what I'm meant to be checking out, are you politely telling me we should just use secure variables and be done with it even though they are sort of ugly? If so, I'll give in otherwise it seems I'm fighting against the grain of how AppVeyor wants to work even though secrets don't belong in source control.

How do we test this before going live against nuget.org, since you can't delete packages just unlisted them? MyGet?

LICENSE.txt
jonorossi: @alinapopa great to know NuGet isn't broken, just the new tooling has a defect we can easily work around rather than renaming the file. Should we rename it back in the branch in Windsor?

appveyor.yml
fir3pho3nixx: The last thing I need to get going here is writing the build version number back into appveyor from the tag. Attempts so far have not been successful.

appveyor.yml
fir3pho3nixx: >I'm not sure what I'm meant to be checking out, are you politely telling me we should just use secure variables and be done with it even though they are sort of ugly? If so, I'll give in otherwise it seems I'm fighting against the grain of how AppVeyor wants to work even though secrets don't belong in source control.

No probs @jonorossi, just testing to water to get your thoughts. I did not think it was too bad but your way is definitely more secure. So let's try it. We can happily use environment variables. The build configuration for the project supports this under the environment link. 

appveyor-env-var

>How do we test this before going live against nuget.org, since you can't delete packages just unlisted them? MyGet?

The way I was approaching this, was to make changes to the `appveyor.yml` in fortress, and then see the tag publishing to NuGet followed by de-listing it. I am also using much lower version numbers so that it does not accidentally get in the way of an upgrade path. 

You can see the releases I have already tested with here(and will get rid of once this is complete): https://github.com/cryosharp/fortress/releases

Once I have fully validated the appveyor.yml, I was going give you instructions on how to setup the NuGet API key environment variable and then it should be a case of merging and hopefully presto! It just works.

appveyor.yml
fir3pho3nixx: Now have appveyor updating build version from tags, yeay!

Please see: fd8de4d8147915d1ffe33dcc3b4bd388fa6003a5

After creating this: https://github.com/cryosharp/fortress/releases/tag/0.5.11

Had the following build run: https://ci.appveyor.com/project/fir3pho3nixx/fortress/build/0.5.11

Going to move on to securing NuGet with env var.

fir3pho3nixx:
@jonorossi - I think this concludes this PR. 

After you merge this, you then want to navigate to the `Environment` link under the `Settings` of the build for Core in AppVeyor create a new secured environment var called `nuget_api_key` and populate it with the secret. 

Then when you create a release, it should deploy the new version(I tested this extensively with fortress, 99% confident it will work). 

appveyor-env-var-1

Castle.Core-VS2017.sln
```diff
@@ -13,7 +13,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Build", "Castle Buil
 		build.sh = build.sh
 		CHANGELOG.md = CHANGELOG.md
 		CONTRIBUTING.md = CONTRIBUTING.md
-		LICENSE = LICENSE
+		LICENSE.txt = LICENSE.txt
```
jonorossi: This needs to be reverted, the filename didn't end up changing.

appveyor.yml
```diff
@@ -26,5 +31,19 @@ after_build:
...
     $wc.UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path "DesktopClrTestResults.xml"))
 
+on_success:
+- ps: >-
+    if ($env:APPVEYOR_REPO_TAG -eq "true")
+    {
+        nuget push .\build\Castle.Core.*.nupkg -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push .\build\Castle.Services.Logging.log4netIntegration.*.nupkg -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push .\build\Castle.Services.Logging.NLogIntegration.*.nupkg -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
+        nuget push .\build\Castle.Services.Logging.SerilogIntegration.*.nupkg -ApiKey $env:NUGET_API_KEY -Source https://api.nuget.org/v3/index.json
```
jonorossi: Looks good, this is much clearer what is happening, however we can't publish the symbol packages to nuget.org.

appveyor.yml
```diff
@@ -26,5 +31,19 @@ after_build:
...
+    }
+
 build_script:
 - cmd: build-vs2017.cmd
+
+artifacts:
+- path: build\*.nupkg
+  name: core
```
jonorossi: What does "name: core" do?

Castle.Core-VS2017.sln
fir3pho3nixx: Good spot! Will fix.

appveyor.yml
fir3pho3nixx: Yes, this is a way of aliasing or naming the artifacts for use later on. We have removed all the providers so it does not make much sense on it's own here like this. 

An example of it being used is here: https://github.com/cryosharp/fortress/blob/f5c7e3b686946f6869ba8b3c0f45e0c2211d651a/appveyor.yml#L44

Want me to remove it?

appveyor.yml
fir3pho3nixx: Yeah, how do we want to deal with symbols? 

Or should we save that for another day ...

jonorossi:
Great work @fir3pho3nixx, I've configured the secure environment variable in AppVeyor for Core and Windsor. I created a new API token that can only push new versions in case an extra package finds its way into the mix.

In the past we've pushed symbol packages to symbolsource.org, should we set the -SymbolSource and API key?

appveyor.yml
jonorossi: Ah, I thought it might be something like that. Don't mind either way, just thought it might have had some effect in the UI but it doesn't appear to.

appveyor.yml
jonorossi: I don't mind if we leave symbols for another day, I assume we'd need to remove the wildcard here though.

appveyor.yml
fir3pho3nixx: Cool :)

fir3pho3nixx:
All squished and ready to be merged.

Issue #253, opened 2017-04-21
Hosting Process "File Lock"'s are causing Dynamic Proxy to fail on win10-x64


ghost:
Dynamic proxy uses static assembly names for generation of proxy types. This is very much needed when it comes to strong naming assemblies and declaring friend assemblies.  The trouble is you cannot control what hosting processes(or consumers of Castle DP) do. 

A file lock on "CastleDynProxy2.dll" via a test run in Resharper, will cause an I/O collision when running tests via the console. 

Is there perhaps a way of creating a "bottom up" way of revealing this?

jonorossi:
This is only a problem for unit tests because they persist their dynamic assembly to the test working directory so the `peverify` tool can be automatically run over the assembly.

Oh, I thought you said you had this problem also when running on the build server which doesn't have VS or ReSharper.

Does this only occur when you are running unit tests at the same time in both ReSharper and CLI? Or is a ReSharper process holding a file lock just sitting there idle?

fir3pho3nixx:
This might also be a problem in appveyor. 

See here: https://ci.appveyor.com/project/fir3pho3nixx/core/build/4.0.37

Errors, Failures and Warnings
1) Failed : Castle.DynamicProxy.Tests.ModuleScopeTestCase
One or more child tests had errors
2) Error : Castle.DynamicProxy.Tests.ModuleScopeTestCase.SaveWithPath
System.IO.IOException : The process cannot access the file 'C:\Users\appveyor\AppData\Local\Temp\1\tmp2388.tmp' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at Castle.DynamicProxy.ModuleScope.SaveAssembly(Boolean strongNamed)
   at Castle.DynamicProxy.Tests.ModuleScopeTestCase.SaveWithPath()

It happens when resharper is sitting there idle. Sometimes ...

jonorossi:
The error with `ModuleScopeTestCase.SaveWithPath` is strange, nothing outside the unit test uses the same temp file name, so either it is blocking itself or there is anti-virus running (which AppVeyor say they don't have).

```csharp
[Test]
public void SaveWithPath()
{
	var strongModulePath = Path.GetTempFileName();    //<<<<<<<<<<
	var weakModulePath = Path.GetTempFileName();    //<<<<<<<<<<

	File.Delete(strongModulePath);
[...]

PR #261, opened 2017-05-18
VS2017 cut over and removal of legacy or deprecated items

fir3pho3nixx:


jonorossi:
I just realised .NET 3.5 and 4.0 support hasn't been added to the new project files and AppVeyor build.

I was also surprised to see `CommonAssemblyInfo.cs` deleted, some of it hasn't been replaced by the new tooling.

fir3pho3nixx:
Good stuff, will see if I can get this going today

fir3pho3nixx:
Saw this in the logs: https://ci.appveyor.com/project/fir3pho3nixx/core/build/1.0.249#L172

It claims "The reference assemblies for framework ".NETFramework,Version=v3.5"" were not found.

I tried the `FrameworkPathOverride` from here: https://github.com/Microsoft/msbuild/issues/1333

Unfortunately it complains that System.Web could not be found. Not sure how to progress this. 

On the plus side `net40` seems to work OK.

Do we need net35?

jonorossi:
> Unfortunately it complains that System.Web could not be found. Not sure how to progress this.

Seems Brad and others are having the same problem.

Those reference assemblies are definitely strange for .NET 3.5. Since reference assemblies were a new thing for .NET 4.0 and then it seems ".NET 3.5 Client Profile" reference assemblies were made, can we set that `FrameworkPathOverride` to `C:\Windows\Microsoft.NET\Framework\v2.0.50727` which has all the assemblies?

> Do we need net35?

If possible, I'd like to keep .NET 3.5 support maybe until Microsoft stops supporting it (2020-01-13, since it shipped with Windows 7) to allow the widest range of usage, the main 3 mocking frameworks currently support:

- Moq is now .NET 4.5+
- FakeItEasy is .NET 4.0+
- NSubstitute is .NET 3.5+

fir3pho3nixx:
Got net35 working on appveyor. 

You can view them in the artifacts here: https://ci.appveyor.com/project/fir3pho3nixx/core/build/1.0.252/artifacts

I just need to deal with the partially trusted assemblies piece and we should be good to go.

fir3pho3nixx:
@jonorossi - Just a nudge to see if this OK to merge. Sorry this has taken so long, my internet is misbehaving. Would like to start on Windsor for this.

jonorossi:
I didn't realise you were done with `CommonAssemblyInfo.cs`, thanks.

I'm not sure how you got .NET 3.5 to compile without defining `FEATURE_LEGACY_REFLECTION_API`. Happy to merge it now, but we'll have to do an assembly diff against v4.0.0 before releasing.

fir3pho3nixx:
@jonorossi - Yes, I found the omission of FEATURE_LEGACY_REFLECTION_API build symbol a bit strange for net35. A better way of making sure this works would be to enable unit tests for all the framework monikers but this would start producing failures similar to https://github.com/castleproject/Core/issues/253. 

What tool would you use to diff the assemblies and what would you be looking for? Maybe this is something I can do in the background and report back on?

jonorossi:
Doesn't the new test runner have to run on .NET 4.6.1? I guess that means it isn't really testing much. Would be great to get unit testing for all targets back again.

I've used https://github.com/bitdiffer/bitdiffer to graphically compare assemblies before with success. Just change the filter settings to hide changes to privates and implementations.

fir3pho3nixx:
Yes, the reason I did that was so that the netcoreapp1.1 framework target would compile using the same csproj as the one we were using for DesktopClr(in this case net461, think the lowest I could go was net452). If you try to do netcoreapp1.1 with net35 for arguments sake it will kick out build errors saying the two are not compatible. 

You were right, there is a problem with net35. After bitdiff'ing I could see the RuntimeVersion value is listed as v4.0.30319 for the net35 assembly. I expected this to be v2.0.50727. 

Also tried your suggestion of setting the `FrameworkPathOverride` to `C:\Windows\Microsoft.NET\Framework\v2.0.50727`, doing this kicked out an entire host of errors to do with System.Linq and by changing it to use `C:\Windows\Microsoft.NET\Framework\v4.0.30319\` got things going again in appveyor but this RuntimeVersion tells me our targetting for net35 is a complete and utter sham on the VS2017 tooling ... 

Let me investigate this more, we cannot release this in it's current state.

PR #265, opened 2017-06-05
Add 5 missing symbols that used to be defined for .NET Framework

jonorossi:
I noticed doing a diff against Castle Core 4.0.0 there were some missing types in the .NET Framework build. I've changed the build scripts to match those symbols defined with project.json.

- Add:
  - `FEATURE_EMIT_CUSTOMMODIFIERS`
  - `FEATURE_EVENTLOG`
  - `FEATURE_SMTP`
  - `FEATURE_TARGETEXCEPTION`
  - `FEATURE_TEST_COM`
- Remove:
  - `FEATURE_SYSTEM_CONVERTER` - no idea where this came from, no code uses it


jonorossi:
@fir3pho3nixx could you take a quick look over this pull request.

You might not be overly happy about me restoring those 2 COM interop assemblies, but you won't get them from nuget.org.

fir3pho3nixx:
@jonorossi - COM? Wow! Now there is a blast from the past. This stuff should be flagged for rewriting using the delete key. 

LGTM 👍

jonorossi:
> COM? Wow! Now there is a blast from the past. This stuff should be flagged for rewriting using the delete key.

`FEATURE_TEST_COM` is only for unit tests, I don't think much actual DynamicProxy code has to do much for COM, people just want to be able to mock the interop interfaces and there were a few defects, but realistically those should probably just be wrapped and the wrappers mocked.

Did you know that the Windows Runtime (WinRT) that was introduced in Windows 8 is based on COM, and is the foundation for Metro/UWP apps. The technology is definitely not dead.

fir3pho3nixx:
I did not know that. I have had my head buried in websites for too long.

jonorossi:
> I did not know that. I have had my head buried in websites for too long.

I only know that because I was at //build/ 2011 when Windows 8 beta was all the rage, the only build conf I've been to, I've never actually written a Metro/UWP app 😏

Issue #121, opened 2016-01-02
ParamArrayAttribute lost in proxy


afscrome:
The only way to detect through reflection if a parameter is a parameter array is by checking if the parameter has the `ParamArrayAttribute` attribute defined on it.  When creating a proxy object, the `ParamArrayAttribute` is lost on the method & indexer parameter arrays.

This can be demonstrated in the following code - I'd expect both assertions to pass, but they fail.

``` csharp
public abstract class Base
{
    public abstract void Method(params int[] args);
    public abstract object this[params int[] args] { get; set; }
}

[Test]
public void ProxyParameterArgumentTest()
{
    var generator = new ProxyGenerator();
    var proxy = generator.CreateClassProxy<Base>();

    var proxyType = proxy.GetType();

    var methodParam = proxyType.GetMethod("Method").GetParameters()[0];
    var indexParam = proxyType.GetProperty("Item").GetIndexParameters()[0];

    Assert.IsTrue(methodParam.IsDefined(typeof(ParamArrayAttribute), false), "Method arg is not params array");
    Assert.IsTrue(indexParam.IsDefined(typeof(ParamArrayAttribute), false), "Index arg is not params array");
}
```

jonorossi:
@afscrome thanks for the detailed report. We've got some support for `ParamArrayAttribute`, however it looks like we've got a defect. Feel free to submit a pull request.

I've assigned this defect to v4.0 milestone.

drieseng:
**ProxyGenerator** indeed does not "replicate" custom attributes for which AttributeUsageAttribute.Inherited is **true**. This check is performed in Castle.DynamicProxy.Internal.AttributeUtil.ShouldSkipAttributeReplication(Type).

Can you elaborate on why this is an issue for you ?

Note that if you use the following asserts, then the first will pass and the second still fails:

`Assert.IsNotNull(Attribute.GetCustomAttribute(methodParam, typeof(ParamArrayAttribute), true));`
`Assert.IsNotNull(Attribute.GetCustomAttribute(indexParam, typeof(ParamArrayAttribute), true));`

I'm inclined to think that's a bug in .NET.

drieseng:
FYI, I submitted a bug against .NET Framework 4.6 ([#2352748](https://connect.microsoft.com/VisualStudio/feedback/details/2352748)) and .NET Core ([#6078](https://github.com/dotnet/corefx/issues/6078)) for the fact that inherited custom attributes are not returned for indexer arguments.

[...]

PR #270, opened 2017-06-10
Ensure `ParamArrayAttribute` is always replicated

stakx:
This is a fix for #121.

This commit introduces special handling of `ParamArrayAttribute` to ensure it always get replicated in generated proxy types.

Up until now, custom attributes only get replicated when they are marked as non-inheritable. `ParamArrayAttribute` should be an exception to this rule. Even the C# compiler has special handling that ensures it gets replicated in subtypes where it has been omitted.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
@@ -234,6 +234,15 @@ namespace Castle.DynamicProxy.Internal
 				return true;
 			}
 
+			// Later, there might be more special cases requiring attribute replication,
+			// which might justify creating a `SpecialCaseAttributeThatShouldBeReplicated`
+			// method and an `AttributesToAlwaysReplicate` class. For the moment, `Param-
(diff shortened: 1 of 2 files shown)


src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -234,6 +234,11 @@ private static bool ShouldSkipAttributeReplication(Type attribute)
 				return true;
 			}
 
+			if (SpecialCaseAttributeThatShouldBeReplicated(attribute))
+			{
+				return false;
+			}
```
stakx: ... or should this be placed a little further up in so it would take preference over `SpecialCaseAttributeThatShouldNotBeReplicated`?

fir3pho3nixx:
@stakx - Your changes seem perfectly reasonable to me, also testing the compiler gives a good signal for when this changes going forward. 👍

jonorossi:
@stakx a simple if-statement is definitely good enough for now. Agreed with @fir3pho3nixx that it is nice to see you are testing the compiler behaviour.

PR #278, opened 2017-06-24
Fix incorrect handling of custom modifiers (modopt and modreq)

### LINKED ISSUE #277: Proxy creation fails where method parameters carry certain modopt combinations
I am forwarding issue https://github.com/moq/moq4/issues/244 here as this appears to be a problem with DynamicProxy, not with Moq.

@mvdtom noticed that trying to mock the following C++/CLI interface with Moq will trigger a `TypeLoadException`:

```csharp
public interface class IFoo
{
    virtual void DoWork(const char type);
};
```

Moq would create a proxy of this interface as follows (simplified):

```csharp
proxyGenerator.CreateClassProxy(typeof(object), new[] { typeof(IFoo) });
```

The exception thrown is:

```
System.TypeLoadException: Signature of the body and declaration in a method implementation do not match.  Type: 'Castle.Proxies.ObjectProxy'.  Assembly: 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Reflection.Emit.TypeBuilder.TermCreateClass(RuntimeModule module, Int32 tk, ObjectHandleOnStack type)
   at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
   at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateType(TypeBuilder type)
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.BuildType()
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateType(String name, Type[] interfaces, INamingScope namingScope)
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.<>c__DisplayClass1_0.<GenerateCode>b__0(String n, INamingScope s)
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateCode(Type[] interfaces, ProxyGenerationOptions options)
   at Castle.DynamicProxy.DefaultProxyBuilder.CreateClassProxyType(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options)
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, Object[] constructorArguments, IInterceptor[] interceptors)
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, IInterceptor[] interceptors)
   at …
```

This probably has to do with C++/CLI's use of optional modifiers (`modopt`) in the method signature. Those can be seen in the IL:

```
.class interface public auto ansi abstract beforefieldinit IFoo
{
    .method public hidebysig newslot abstract virtual instance void DoWork(
      int8
      modopt([mscorlib]System.Runtime.CompilerServices.IsConst)
      modopt([mscorlib]System.Runtime.CompilerServices.IsSignUnspecifiedByte) 'type') cil managed
    { }
}
```

@mvdtom also noticed that the exception doesn't happen if the C++ `const` modifier is omitted, or if he chooses a type other than `char`. Based on this, I ran some additional tests of my own.

I defined a handful of types in IL that differ only in the specific combination of `modopt`s the parameter carries (see [`TypesWithModopt.il`](https://github.com/castleproject/Core/files/1098938/TypesWithModopt.il.txt), which was compiled using `ilasm` and checked with `peverify`). Then I tried to create proxies for all those types (see [`Program.cs`](https://github.com/castleproject/Core/files/1098939/Program.cs.txt)). Here's the result:

modopt

In summary, DynamicProxy...

* ... doesn't care about the order of modopts.
* ... can deal with any number of `IsConst` modopts.
* ... can deal with one single `IsSignUnspecifiedByte` modopt that is not accompanied by any other modopts.
* ... can deal with one single custom (user-defined) modopt that is not accompanied by any other modopt.

Given that the last two bullet points are very similar, I would guess that DynamicProxy really only supports `IsConst`, and otherwise tolerates a single modopt. I haven't tried yet what happens if several modopts are spread across more than one parameter.

What is DynamicProxy's position and support for modopts? I realise it is probably geared mainly towards C# and VB.NET, but I'm still curious if DynamicProxy wants to support them at all, or not.

stakx:
This fixes #277.

In response to https://github.com/castleproject/Core/issues/277#issuecomment-310828795, this PR begins by removing the external DLL `Rhino.Mocks.CPP.Interfaces.dll`, which was used for testing DynamicProxy's handling of modopts. It gets replaced with an equivalent dynamically generated assembly.

According to the same principle, a new test fixture targeting DynamicProxy's handling of custom modifiers (`CustomModifiersTestCase`) emits more test types into a dynamically generated assembly and then tries to create proxies for those.

Finally, a bugfix in `MethodEmitter.cs` makes all tests pass. (@fir3pho3nixx found the error cause&mdash;which is that .NET reports custom modifiers in reversed order&mdash;and provided the initial bugfix; thanks for that!)


fir3pho3nixx:
The only thing I think we are missing is an entry in CHANGELOG.md

stakx:
I'll look into / add it tomorrow.

stakx:
Not sure if that last comment was directed at me... if so, I can add an entry in CHANGELOG.md if you can let me know the proper format / version number for a yet unpublished version.

fir3pho3nixx:
Usually done this: 

```

## Unreleased
  		  
 * Fixed modopts (@stakx, #241)

```

Just at the top of the CHANGELOG.md

stakx:
Done.

fir3pho3nixx:
Awesome. Will wait for @jonorossi to double check.

PR #279, opened 2017-06-24
Add failing tests for modopt method signatures

stakx:
This PR adds unit tests targeting the bug identified in #277. You probably don't want to merge this just yet, since some of those unit tests currently fail.

If you are satisfied with the changes PR #278 proposes, we can then probably continue with this one here and try to add a fix for the bug identified.


stakx:
I'll combine this PR with #278 in response to https://github.com/castleproject/Core/issues/277#issuecomment-310862054, so closing this one.

Issue #280, opened 2017-06-29
Replace adodb.dll used in unit tests with dynamically generated type


jonorossi:
We currently have some random `adodb.dll` committed to our source repository used by `ComInterfacesTests`, can we get rid of it the same way we got rid of `Rhino.Mocks.CPP.Interfaces.dll` in #277?

/cc @stakx

stakx:
I saw that DLL, but wasn't sure where it belonged to, so I opted not to touch it. I can look into it later today, if you'd like.

jonorossi:
@stakx that would be great.

stakx:
### What is `adodb.dll`? ###

`adodb.dll` turns out to be an older version of the ActiveX Data Objects (ADO) Primary Interop Assembly. That is, it contains COM interop type definitions describing COM components that need to be installed on your system for your unit tests to run successfully. For example, on my box, the unit tests in `ComInterfaceTypes.cs` (which reference `adodb.dll`) will cause COM components in `C:\Program Files\Common Files\System\ado\msado15.dll` to be loaded at runtime. Fortunately, from what I can tell, those ADO components are probably always installed on a VS developer box.

### How can we get rid of it, and what do we gain by it? ###

We can get rid of `adodb.dll` quite easily, namely by reproducing the COM interop type declarations it contains as C# interfaces (decorated with `[ComImport]`, `[Guid]`, and the like) in the unit test project. However, getting rid of `adodb.dll` does not remove the runtime dependency on the COM components those declarations describe!

### Can we get rid of the runtime dependency, too? ###

If we want Castle.Core unit tests to be fully self-contained, i.e. have no dependencies on any pre-installed  external COM DLLs, we would need to implement our own COM or COM-like component. I can do this and send a PR, but before I do so, I'd like you to know what this implies:

 * We should not implement our own COM component as managed code, but using unmanaged C or C++. A single small, [lightweight COM](https://stackoverflow.com/questions/38174095/what-exactly-is-lightweight-com) component would do. We don't even need to register the resulting DLL with the operating system's COM subsystem (which only exists on Windows, anyway). All we'd rely on is .NET's P/Invoke and COM interop facilities.

 * But we still have C or C++ code which we need to be able to compile. We'll need to make sure the CI servers have MSVC and are configured to compile a VS C++ project. I expect this to be the main difficulty.

[...]

PR #281, opened 2017-06-29
Only release non-null COM interface pointers

stakx:
While playing around a bit for #280, I stumbled upon what looks like a minor bug:

When given a COM object as the target, `ProxyGenerator.CreateInterfaceProxyWithTargetInterface` checks whether that target implements the specified `interfaceToProxy`. It does this by calling `Marshal.QueryInterface` on the target, seeing whether it got back a non-null interface pointer, and
releasing the retrieved interface pointer (since it doesn't actually need it for anything).

However, the retrieved interface pointer must only be released if it is not a null reference. This commit adds the necessary check for that.

### DIFF
--- a/src/Castle.Core.Tests/ComInterfacesTests.cs
+++ b/src/Castle.Core.Tests/ComInterfacesTests.cs
@@ -15,13 +15,23 @@
 #if FEATURE_TEST_COM && !__MonoCS__ // Avoid loading adodb.dll
 namespace Castle.DynamicProxy.Tests
 {
+	using System;
 	using System.Reflection;
+	using System.Runtime.InteropServices;
+
 	using ADODB;
+
 	using NUnit.Framework;
 
 	[TestFixture]
 	public class ComInterfacesTests:BasePEVerifyTestCase
 	{
+		[Test]
+		public void Marshal_Release_throws_when_called_with_IntPtr_Zero()
+		{
+			Assert.Throws<ArgumentNullException>(() => Marshal.Release(IntPtr.Zero));
+		}
+
 		[Test]
 		public void Can_proxy_ADO_RecorsSet()
 		{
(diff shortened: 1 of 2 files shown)


src/Castle.Core.Tests/ComInterfacesTests.cs
```diff
@@ -15,14 +15,24 @@
...
 	using NUnit.Framework;
 
 	[TestFixture]
 	public class ComInterfacesTests:BasePEVerifyTestCase
 	{
 		[Test]
+		public void Marshal_Release_throws_when_called_with_IntPtr_Zero()
+		{
+			Assert.Throws<ArgumentNullException>(() => Marshal.Release(IntPtr.Zero));
+		}
```
stakx: This test probably shouldn't be in that fixture... but it felt like overkill to create a separate one for just this basic bit of .documentation.

src/Castle.Core/DynamicProxy/ProxyGenerator.cs
```diff
@@ -595,7 +595,10 @@ public object CreateInterfaceProxyWithTarget(Type interfaceToProxy, object targe
 						var result = Marshal.QueryInterface(iUnknown, ref interfaceId, out interfacePointer); // Increment the reference count
 						var isInterfacePointerNull = interfacePointer == IntPtr.Zero;
 						Marshal.Release(iUnknown); // Decrement the reference count
-						Marshal.Release(interfacePointer); // Decrement the reference count
+						if (!isInterfacePointerNull)
+						{
+							Marshal.Release(interfacePointer); // Decrement the reference count
+						}
 
 						if (result == 0 && isInterfacePointerNull)
```
stakx: Actually, I suspect there's another bug right here: If a COM component does not implement a particular interface, `QueryInterface` is supposed to return a `HRESULT` of `E_NOINTERFACE` (0x80004002), not `S_OK` (0). Therefore I strongly suspect this check should really be:

```csharp
if (result == 0x80004002 && isInterfacePointerNull)
```

but I'd like to prove this with a test before I submit the fix.

jonorossi: Looks like all the error cases weren't handled in this code. This code path probably hasn't been used much.

src/Castle.Core.Tests/ComInterfacesTests.cs
jonorossi: This test doesn't really add anything other than documentation, however it is useful because this type of call in C++ would just return 0 references left as exceptions really are for exceptional situations (e.g. OOM, AV). I had actually thought Marshal.Release wouldn't throw with a null pointer, but it does.

Issue #282, opened 2017-06-30
Mono build is broken but Travis reports success


jonorossi:
The build was broken by #279, but looks like build.sh returns 0 on failure.

```
e/travis/build/castleproject/Core/src/Castle.Core/bin/Release/net45/Castle.Core.dll
e/travis/build/castleproject/Core/src/Castle.Services.Logging.SerilogIntegration/bin/Release/net45/Castle.Services.Logging.SerilogIntegration.dll
e/travis/build/castleproject/Core/src/Castle.Services.Logging.log4netIntegration/bin/Release/net45/Castle.Services.Logging.log4netIntegration.dll
e/travis/build/castleproject/Core/src/Castle.Services.Logging.NLogIntegration/bin/Release/net45/Castle.Services.Logging.NLogIntegration.dll
CustomModifiersTestCase.cs(128,9): error CS0117: 'BasePEVerifyTestCase' does not contain a definition for 'RunPEVerifyOnGeneratedAssembly' [/home/travis/build/castleproject/Core/src/Castle.Core.Tests/Castle.Core.Tests.csproj]
Build FAILED.
CustomModifiersTestCase.cs(128,9): error CS0117: 'BasePEVerifyTestCase' does not contain a definition for 'RunPEVerifyOnGeneratedAssembly' [/home/travis/build/castleproject/Core/src/Castle.Core.Tests/Castle.Core.Tests.csproj]
    0 Warning(s)
    1 Error(s)
Time Elapsed 00:00:15.73
Running NET461 Tests
--------------------
./build.sh: line 52: ./src/Castle.Core.Tests/bin/Release/net461/Castle.Core.Tests.exe: No such file or directory
---------------------------
Running NETCOREAPP1.1 Tests
---------------------------
No executable found matching command "dotnet-./src/Castle.Core.Tests/bin/Release/netcoreapp1.1/Castle.Core.Tests.dll"
grep: NetCoreClrTestResults.xml: No such file or directory
grep: DesktopClrTestResults.xml: No such file or directory
The command "./build.sh" exited with 0.
Done. Your build exited with 0.
```
(https://travis-ci.org/castleproject/Core/builds/248289034)

/cc @fir3pho3nixx @stakx

jonorossi:
@stakx looks like all the new unit tests fail on Mono too with something like this:

```
[...]

PR #283, opened 2017-06-30
Add missing Mono guard for PEVerify

stakx:
Once the build bug described in #282 is fixed, this should take care of the compilation error I caused on the Mono build.

Perhaps it would be better to add the `#if !__MonoCS__` directly inside the method body of  `BasePEVerifyTestCase.RunPEVerifyOnGeneratedAssembly` to prevent that preprocessor symbol from spilling into dependent code?


jonorossi:
Don't you also need to do the same to `RhinoMocksTestCase`?

`RunPEVerifyOnGeneratedAssembly` is only used by these two test cases with these changes so this change is fine. It looks like the Roslyn team is working on an open source peverify which we'll be able to move to in the future.

stakx:
I guess not, there's a giant `#if FEATURE_TEST_COM && !__MonoCS__` around all code in that file.

PR #89, opened 2015-04-24
Unit test reproducing ProxyGenerationException using 3+ observer (issue #88)

smudge202:
Multiple (of the same) generic interface containing non-generic members can now be implemented.

Issue #88, opened 2015-04-24
ProxyGenerationException proxying the same generic interface multiple times


herecydev:
I've made a [gist](https://gist.github.com/herecydev/081a7703866f19005b95) to demonstrate the ProxyGenerationException I get when trying to mock an interface with multiple observers.

The issue seems to only occur when running at least 3 observers (2 is fine strangely).

```
Test Outcome:   Failed
Test Duration:  0:00:00.211

Result StackTrace:  
at Castle.DynamicProxy.Generators.TypeElementCollection`1.Add(TElement item)
   at Castle.DynamicProxy.Contributors.CompositeTypeContributor.CollectElementsToProxy(IProxyGenerationHook hook, MetaType model)
   at Castle.DynamicProxy.Generators.InterfaceProxyWithoutTargetGenerator.GenerateType(String typeName, Type proxyTargetType, Type[] interfaces, INamingScope namingScope)
   at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.<>c__DisplayClass1.<GenerateCode>b__0(String n, INamingScope s)
   at Castle.DynamicProxy.Generators.BaseProxyGenerator.ObtainProxyType(CacheKey cacheKey, Func`3 factory)
   at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.GenerateCode(Type proxyTargetType, Type[] interfaces, ProxyGenerationOptions options)
   at Castle.DynamicProxy.DefaultProxyBuilder.CreateInterfaceProxyTypeWithoutTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options)
   at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithoutTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, IInterceptor[] interceptors)
   at Moq.Proxy.CastleProxyFactory.CreateProxy(Type mockType, ICallInterceptor interceptor, Type[] interfaces, Object[] arguments)
   at Moq.Mock`1.<InitializeInstance>b__2()
   at Moq.PexProtector.Invoke(Action action)
   at Moq.Mock`1.InitializeInstance()
   at Moq.Mock`1.OnGetObject()
   at Moq.Mock`1.get_Object()

Result Message: Castle.DynamicProxy.ProxyGenerationException : Duplicate element: Castle.DynamicProxy.Generators.MetaMethod

```

[...]

PR #285, opened 2017-07-02
Prevent member name collision when proxy implements same generic interface more than twice

stakx:
After #89, this is a second attempt at fixing #88.

When a proxy is made to implement the same generic interface more than twice, and that generic interface contains a method that does not use any generic type parameters, this can result in a member name collision. This would happen e.g. with a proxy implementing `IObserver<T>` three times.

DynamicProxy already checks for member name collisions, and when it detects one, it switches to explicit interface implementation and prefixes member names with the interface's type name. For the above example, this will add the following methods to the proxy:

    OnCompleted
    IObserver`1.OnCompleted
    IObserver`1.OnCompleted

Because the actual type arguments are omitted and only their count is included, the current collision checks are not sufficient. This commit makes explicitly implemented methods' names unique by including the full type names of the generic type arguments; for instance:

    OnCompleted
    System.IObserver`1[[System.Int32, mscorlib, Version=…, Culture=…, PublicKeyToken=…]].OnCompleted
    System.IObserver`1[[System.String, mscorlib, Version=…, Culture=…, PublicKeyToken=…]].OnCompleted

(I've omitted some detail above for brevity, using `…`.)

This is about the simplest possible solution to #88, but I'm not sure if you are OK with these kinds of member names. They are unique, but also long and ugly (which could be partly mitigated by using `System.CodeDom` to produce more C#-like type names), and they contain whitespace and other special characters; I have no idea whether these are always safe for proxy consumers.

Feel free to request modifications or to close this PR if it's not acceptable.

### DIFF
--- a/src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs
+++ b/src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs
@@ -217,6 +217,52 @@ namespace Castle.DynamicProxy.Tests
 			Assert.IsNotNull(method2);
 		}
 
+		[Test]
+		public void Should_choose_noncolliding_member_names_when_implementing_same_generic_interface_several_times()
+		{
+			var boolInterfaceType = typeof(IGenericWithNonGenericMethod<bool>);
+			var intInterfaceType = typeof(IGenericWithNonGenericMethod<int>);
+			var nestedGenericBoolInterfaceType = typeof(IGenericWithNonGenericMethod<IGenericWithNonGenericMethod<bool>>);
+
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(
+				boolInterfaceType,
+				new[] { intInterfaceType, nestedGenericBoolInterfaceType });
+
+			var type = proxy.GetType().GetTypeInfo();
+
+			var boolMethod = type.GetRuntimeInterfaceMap(boolInterfaceType).TargetMethods[0];
+			Assert.AreEqual("DoSomething", boolMethod.Name);
+
+			var intMethod = type.GetRuntimeInterfaceMap(intInterfaceType).TargetMethods[0];
+			Assert.AreEqual("IGenericWithNonGenericMethod`1[Int32].DoSomething", intMethod.Name);
+
+			var nestedGenericBoolMethod = type.GetRuntimeInterfaceMap(nestedGenericBoolInterfaceType).TargetMethods[0];
+			Assert.AreEqual("IGenericWithNonGenericMethod`1[IGenericWithNonGenericMethod`1[Boolean]].DoSomething", nestedGenericBoolMethod.Name);
+		}
+
+		[Test]
+		public void Should_choose_noncolliding_member_names_when_implementing_same_generic_interface_with_two_type_args_several_times()
+		{
+			var boolIntInterfaceType = typeof(IGenericWithNonGenericMethod<bool, int>);
+			var intBoolInterfaceType = typeof(IGenericWithNonGenericMethod<int, bool>);
+			var intNestedGenericBoolInterfaceType = typeof(IGenericWithNonGenericMethod<int, IGenericWithNonGenericMethod<bool>>);
+
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(
+				boolIntInterfaceType,
+				new[] { intBoolInterfaceType, intNestedGenericBoolInterfaceType });
+
+			var type = proxy.GetType().GetTypeInfo();
+
+			var boolIntMethod = type.GetRuntimeInterfaceMap(boolIntInterfaceType).TargetMethods[0];
+			Assert.AreEqual("DoSomething", boolIntMethod.Name);
+
+			var intBoolMethod = type.GetRuntimeInterfaceMap(intBoolInterfaceType).TargetMethods[0];
+			Assert.AreEqual("IGenericWithNonGenericMethod`2[Int32,Boolean].DoSomething", intBoolMethod.Name);
+
+			var intNestedGenericBoolMethod = type.GetRuntimeInterfaceMap(intNestedGenericBoolInterfaceType).TargetMethods[0];
+			Assert.AreEqual("IGenericWithNonGenericMethod`2[Int32,IGenericWithNonGenericMethod`1[Boolean]].DoSomething", intNestedGenericBoolMethod.Name);
+		}
+
 		private ParameterInfo[] GetMyTestMethodParams(Type type)
 		{
 			MethodInfo methodInfo = type.GetMethod("MyTestMethod", BindingFlags.Instance | BindingFlags.Public);
--- a/src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs
+++ b/src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs
@@ -218,7 +218,7 @@ namespace Castle.DynamicProxy.Tests
 		}
 
 		[Test]
-		public void Should_choose_noncolliding_member_names_when_implementing_same_generic_interface_several_times()
+		public void Should_choose_noncolliding_method_names_when_implementing_same_generic_interface_several_times()
 		{
 			var boolInterfaceType = typeof(IGenericWithNonGenericMethod<bool>);
 			var intInterfaceType = typeof(IGenericWithNonGenericMethod<int>);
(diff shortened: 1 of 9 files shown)


src/Castle.Core.Tests/BasicInterfaceProxyTestCase.cs
```diff
@@ -213,10 +213,37 @@ public void Should_implement_explicitly_duplicate_interface_members()
 			MethodInfo method = type.GetMethod("Foo", BindingFlags.Instance | BindingFlags.Public);
 			Assert.IsNotNull(method);
 			Assert.AreSame(method, type.GetTypeInfo().GetRuntimeInterfaceMap(typeof(IIdenticalOne)).TargetMethods[0]);
-			MethodInfo method2 = type.GetMethod("IIdenticalTwo.Foo", BindingFlags.Instance | BindingFlags.Public);
+			MethodInfo method2 = type.GetMethod($"{typeof(IIdenticalTwo).FullName}.Foo", BindingFlags.Instance | BindingFlags.Public);
```
stakx: The reason why I chose not to hardcode the full type name here is because it includes things like DLL version numbers and public keys, which I wasn't sure whether they are the same on Mono and the CLR, are they're irrelevant to the test anyway.

jonorossi:
As you suspected, I'm not a fan of those member names. It appears they also won't be CLS compliant with the spaces: https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components#naming, I'm also not sure dots would be compliant but they don't seem to be mentioned. However, I'm not sure CLS compiliance really matters since no one is going to manually call them from language compiled code. I'd hate to see a stacktrace with them though, would be awfully confusing.

What do the CodeDom identifiers look like?

I'd be happy with using just the parameter short names (i.e. not namespace or assembly-qualified) which is what is already done for the first bit (e.g. `IIdenticalTwo` rather than `Castle.DynamicProxy.Tests.Interfaces.IIdenticalTwo`):
```
OnCompleted
IObserver`1.OnCompleted
IObserver`1.OnCompleted

OnCompleted
IObserver`1[Int32].OnCompleted  // <<<
IObserver`1[String].OnCompleted // <<<
```

The [docs for `Type.GetType`](https://msdn.microsoft.com/en-us/library/w3f99sx1.aspx) state that the extra set of brackets is only needed around assembly-qualified types, so the format would look like this for multiple parameters:

```
IObserver`2[Int32,String].OnCompleted
```

Unless I missed it I can't see any property in the BCL on a type to get the unqualified name as well as its generic parameters, but should be easy to implement.

stakx:
> What do the CodeDom identifiers look like?

I only mentioned this as a potential route to take, I haven't actually tried it out just yet. The general idea is that CodeDom can be used to create type names that look more like C# rather than IL. For instance, `<>` brackets for generic type arguments instead of the backtick and the `[]` brackets). My understanding of this technique is that it is customizable to some extent (e.g. whether you want to include namespaces or not).  That all being said, it might actually be better to keep an additional subsystem like CodeDom out of the picture, if it's not absolutely required.

>     IObserver`2[Int32,String].OnCompleted

I like that suggestion. The resulting member names won't be guaranteed to be unique (e.g. if you use two types `NamespaceA.Foo` and `NamespaceB.Foo` as type arguments), but in practice, that probably won't matter at all in most cases.

Shall I go ahead and change the PR accordingly?

jonorossi:
> That all being said, it might actually be better to keep an additional subsystem like CodeDom out of the picture, if it's not absolutely required.

Completely in agreement, not even sure if .NET Core has CodeDom.

> The resulting member names won't be guaranteed to be unique

Agreed, it'll be exactly like the base type right now, you'd be silly to inherit/implement two `Foo` classes/interfaces 😉 

> Shall I go ahead and change the PR accordingly?

Yep, don't forget the changelog. Thanks.

stakx:
I think something might still be missing here: This PR only takes care of method name collisions, but does nothing about events and properties. I'll try to see if a similar change needs to be made to the other types overriding `MetaTypeElement.SwitchToExplicitImplementation` (`MetaEvent.cs`, `MetaProperty.cs`).

src/Castle.Core/DynamicProxy/Generators/MetaMethod.cs
```diff
@@ -101,7 +102,30 @@ internal override void SwitchToExplicitImplementation()
...
+		}
+
+		private static void AppendTypeNameTo(StringBuilder stringBuilder, Type type)
+		{
+			stringBuilder.Append(type.Name);
+			if (type.GetTypeInfo().IsGenericType)
+			{
+				stringBuilder.Append('[');
+				var genericTypeArguments = type.GetGenericArguments();
+				for (int i = 0, n = genericTypeArguments.Length; i < n; ++i)
```
fir3pho3nixx: This method might introduce a performance hit on generation.

fir3pho3nixx: Was just basing that on a rudimentary find of usages for `MetaMethod`. 

image

stakx: That's correct. The recursion depth will be proportional to the nesting depth of the (generic) type for which it generates a name. Non-generic types, however, will not trigger any recursion.

(Replacing recursion with an imperative `Stack`-and-loop approach would probably be even worse, since it would likely entail an additional heap allocation that can be avoided with recursion.)

fir3pho3nixx: It fixes the original problem, but should we flag performance as a feature and make it part of the proxy options with a sensible default?

jonorossi: I don't think the performance will be affected. This only gets run when `SwitchToExplicitImplementation` is called once per `MetaMethod` and it is then cached in its `name` field. The `StringBuilder` will perform pretty similar to the existing `string.Format` and users are rarely going to have more than a single level of generic nesting, and even then the cost here is negligible compared to the actual type generation or other work DP does.

jonorossi:
Looks good so far, just need to address properties and events.

stakx:
@jonorossi - Done.

I've extracted the logic for creating explicit implementation member names into a separate helper class `MetaTypeElementUtil`, since it's identical for all types of members (methods, events, and properties).

This opens up an opportunity to minimize unnecessary heap allocations. @fir3pho3nixx, like you showed in your review before, generating names for explicitly implemented members is possibly on a hot code path.  It would be silly, for instance, having to instantiate 3 `StringBuilder` objects for just one property (1 for the property name, 1 for the getter method name, 1 for the setter method name). That's why I introduced a `[ThreadStatic] StringBuilder` that gets reused over and over again.

fir3pho3nixx:
@stakx - In this context I think it plays well with async. There is a [caution](https://stackoverflow.com/questions/26516730/is-the-use-of-threadstatic-at-odds-with-asynchronous-code) about the use of ThreadStatic. We also switched this out for use of AsyncLocal for netcore.

stakx:
@fir3pho3nixx - Thanks for checking up on my use of `[ThreadStatic]`. I was initially unsure whether it was really *that* easy to work with, but I've come to believe it should be perfectly fine in this particular case.

If my understanding of the TPL and `async` / `await` is correct, that Stack Overflow question revolves around the topic of task scheduling: is the continuation (i.e. the code after the `await`) guaranteed to execute on the same thread as the code before the `await`?

I am almost certain that won't be an issue here, for the simple reason that our code in question is entirely synchronous. It should be impossible that code execution would switch from one thread to another all of a sudden and thus end up "stealing" the original thread's `StringBuilder`.

Unless I'm mistaken with any or all of the above, `AsyncLocal<T>` would be overkill here, because we have no `Task` or `async` / `await` induced async boundary across which we need to persist state.

fir3pho3nixx:
@stakx - I agree. It is all sync so it should be fine.

fir3pho3nixx:
@stakx - is string builder threadsafe?

jonorossi:
> In this context I think it plays well with async

Agreed. In general `ThreadStatic` will not work correctly in async/await, however there are no continuations in this code.

> This opens up an opportunity to minimize unnecessary heap allocations.

What concerns me is that `ThreadStatic` and TLS are not free, the CLR has to call a helper function to access TLS to fetch that pointer for you, and this means the JIT can't aggressively optimise around accesses to this field. StringBuilders are fairly cheap and they'd only end up in Gen0 so would be efficiently garbage collected, and would mean StringBuilders aren't left around on the heap with pointers also left in TLS after generation is done (especially something long lived like a web app). I haven't done any performance tests, but I did come across one comment that said accessing a `ThreadStatic` field can be slower than an array allocation on .NET 3.5 (StringBuilder's constructor is pretty much just doing a char array allocation).

With that said the .NET Framework does actually have an [internal `StringBuilderCache` class](http://referencesource.microsoft.com/#mscorlib/system/text/stringbuildercache.cs,a6dbe82674916ac0) which is used by methods including `String.Format`, `String.Join` and `String.Concat(IEnumerable<>)`. The class is a little more complex to optimise reuse of StringBuilders with enough capacity, because of fragmentation and internally StringBuilders are a linked list of StringBuilders to prevent any of it being allocated on the large object heap.

In this case it appears there is enough evidence that this will likely improve performance by some margin, but the performance improvement is currently unknown and likely negligible compared to the reflection calls to get the type names along with generic types; and this is all small compared to the reflection emit time.

It is best to avoid premature optimization (especially premature micro-optimisations) without measuring/benchmarking especially when the readability is hindered. For example, you could add a fast path to `CreateNameForExplicitImplementation ` that avoids the StringBuilder if it isn't a generic type and call `string.Concat(type.Name, ".", name)` so it hits `String`'s `FastAllocateString` method.

Let me know your thoughts, and if you disagree.

stakx:
@jonorossi: Your response makes perfect sense to me.

I've heard similar things about the relatively large overhead of `[ThreadStatic]` on .NET 3.5. I've also read that this seems to have been remedied in .NET 4, where the overhead of using `[ThreadStatic]` is said to be negligible. But clearly, we'd simply have to run benchmarks to see whether `[ThreadStatic]` performs acceptably or not.

I'm happy to change the code to just newing up `StringBuilder`s as needed and add a fast path for non-generic types if you prefer that.

One thing to note, though: The framework's inbuilt `StringBuilderCache` relies on `[ThreadStatic]`, too!

Let me know how I should proceed.

fir3pho3nixx:
Thanks for making this thread interesting. Did not know threadstatic performance issue existed(or was fixed), just generally avoided it like herpes altogether :) Learn't something 👍

jonorossi:
@stakx thanks for making the change, and for getting this fix in. Merging.

PR #286, opened 2017-07-04
Fix Mono build by excluding failing modopt tests

stakx:
as discussed in https://github.com/castleproject/Core/issues/277#issuecomment-312653793. I am happy to provide a proper fix for Mono at a later time, but I guess for now it's more important to have a successful build again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 Bugfixes:
-- Fix incorrect replication (reversed order) of custom modifiers (modopts and modreqs) (@stakx, #277)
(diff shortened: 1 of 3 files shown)


jonorossi:
@stakx could you change the library code to use Array.Reverse rather than Linq.

jonorossi:
@stakx could you also update the entry in the changelog to mention this does not work on Mono, doesn't need to say why or anything just those 5 words.

stakx:
@jonorossi - yep, will do. I wanted to see first whether the build works allright. I'll add a commit.

stakx:
> could you change the library code to use Array.Reverse rather than Linq.

Sorry, overlooked that one. Will do that, too.

jonorossi:
Thanks

PR #290, opened 2017-07-19
Make `ProxyUtil.IsAccessible(MethodBase)` take into account declaring type's accessibility

### LINKED ISSUE #289: ProxyUtil.IsAccessible(MethodBase) reports false positive for method declared in internal class
I just noticed that in Castle.Core 4.1.1, [`Castle.DynamicProxy.ProxyUtil.IsAccessible(MethodBase method)`](https://github.com/castleproject/Core/blob/1e2b33249e2e23448f68a6ef5674b8eb532d1dae/src/Castle.Core/DynamicProxy/ProxyUtil.cs#L86-L93) appears to report false positives for methods declared within internal classes:

```csharp
using Castle.DynamicProxy;

// [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] has not been declared.

public abstract class P
{
    internal abstract void I();
}

internal abstract class I
{
    public abstract void P();
}

public class ProxyUtilTestCase
{
    [Test]
    public void IsAccessible_GivenInternalMethodInPublicClass_ReturnsFalse()
    {
        var pi = typeof(P).GetMethod("I", BindingFlags.NonPublic | BindingFlags.Instance);
        Assert.False(ProxyUtil.IsAccessible(pi));  // passes
    }

    [Test]
    public void IsAccessible_GivenPublicMethodInInternalClass_ReturnsFalse()
    {
        var ip = typeof(I).GetMethod("P", BindingFlags.Public | BindingFlags.Instance);
        Assert.False(ProxyUtil.IsAccessible(ip));  // fails
    }
}
```

How can method `I.P` be accessible to DynamicProxy when the declaring type `I` isn't? After all, `new ProxyGenerator().CreateClassProxy<I>()` would fail.

Am I misunderstanding the purpose of that method, or can you confirm that this is a bug?

stakx:
This should fix #289.

I tried to follow [@jonorossi's suggestions](https://github.com/castleproject/Core/issues/289#issuecomment-316034650) as closely as possible, but I might've done a few things too many here, or misunderstood something. If so, please let me know. See the commit notes for details. A few points regarding the code moves:

* I moved everything from `InternalsUtil` to `ProxyUtil`. The former class now only contains "forwarding" methods which are hidden from IntelliSense, but not marked as `[Obsolete]`; none of them has an exact correspondence with a public method in `ProxyUtil`. (I chose **not** to make the moved methods `public` right away as that would've led to additional duplicate methods in the public API.)

* [Those three methods from `DefaultProxyBuilder`](https://github.com/castleproject/Core/blob/v4.1.1/src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs#L148-L165) become a single method `ProxyUtil.IsAccessibleType`. (One is a helper method, and the other two only ever get called together. Combining them also allows optimizing away repeated calls to `GetTypeInfo`.)

* Moving extension methods into the `public /* non-static */ class ProxyUtil` means they can no longer be extension methods. (I opted **not** to re-declare `ProxyUtil` as `static` right away.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Make ProxyUtil.IsAccessible(MethodBase) take into account declaring type's accessibility so it doesn't report false negatives for e.g. public methods in inaccessible classes. (@stakx, #289)
+
 ## 4.1.1 (2017-07-12)
 
 Bugfixes:
--- a/src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
@@ -28,6 +28,7 @@ namespace Castle.DynamicProxy.Internal
 		///   <c>true</c> if the specified method is internal; otherwise, <c>false</c>.
 		/// </returns>
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete]
 		public static bool IsInternal(this MethodBase method)
 		{
 			return ProxyUtil.IsInternal(method);
@@ -38,6 +39,7 @@ namespace Castle.DynamicProxy.Internal
 		/// </summary>
 		/// <param name = "asm">The assembly to inspect.</param>
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete]
 		public static bool IsInternalToDynamicProxy(this Assembly asm)
 		{
 			return ProxyUtil.AreInternalsVisibleToDynamicProxy(asm);
@@ -49,6 +51,8 @@ namespace Castle.DynamicProxy.Internal
 		/// <param name = "method"></param>
 		/// <returns></returns>
 		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Use " + nameof(ProxyUtil) + "." + nameof(ProxyUtil.IsAccessible) + " instead, " +
+		          "which performs a more accurate accessibility check.")]
 		public static bool IsAccessible(this MethodBase method)
 		{
 			return ProxyUtil.IsAccessibleMethod(method);
--- a/src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
@@ -15,19 +15,11 @@
 namespace Castle.DynamicProxy.Internal
 {
 	using System;
-	using System.Collections.Generic;
-	using System.Linq;
+	using System.ComponentModel;
 	using System.Reflection;
-	using System.Runtime.CompilerServices;
-
-	using Castle.Core.Internal;
-	using Castle.DynamicProxy.Generators.Emitters;
 
 	public static class InternalsUtil
 	{
-		private static readonly IDictionary<Assembly, bool> internalsToDynProxy = new Dictionary<Assembly, bool>();
-		private static readonly Lock internalsToDynProxyLock = Lock.Create();
-
 		/// <summary>
 		///   Determines whether the specified method is internal.
 		/// </summary>
@@ -35,43 +27,20 @@ namespace Castle.DynamicProxy.Internal
 		/// <returns>
 		///   <c>true</c> if the specified method is internal; otherwise, <c>false</c>.
 		/// </returns>
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static bool IsInternal(this MethodBase method)
 		{
-			return method.IsAssembly || (method.IsFamilyAndAssembly
-			                             && !method.IsFamilyOrAssembly);
+			return ProxyUtil.IsInternal(method);
 		}
 
 		/// <summary>
 		///   Determines whether this assembly has internals visible to dynamic proxy.
 		/// </summary>
 		/// <param name = "asm">The assembly to inspect.</param>
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static bool IsInternalToDynamicProxy(this Assembly asm)
 		{
-			using (var locker = internalsToDynProxyLock.ForReadingUpgradeable())
-			{
-				if (internalsToDynProxy.ContainsKey(asm))
-				{
-					return internalsToDynProxy[asm];
-				}
-
-				locker.Upgrade();
-
-				if (internalsToDynProxy.ContainsKey(asm))
-				{
-					return internalsToDynProxy[asm];
-				}
-
-				var internalsVisibleTo = asm.GetCustomAttributes<InternalsVisibleToAttribute>();
-				var found = internalsVisibleTo.Any(VisibleToDynamicProxy);
-
-				internalsToDynProxy.Add(asm, found);
-				return found;
-			}
-		}
-
-		private static bool VisibleToDynamicProxy(InternalsVisibleToAttribute attribute)
-		{
-			return attribute.AssemblyName.Contains(ModuleScope.DEFAULT_ASSEMBLY_NAME);
+			return ProxyUtil.AreInternalsVisibleToDynamicProxy(asm);
 		}
 
 		/// <summary>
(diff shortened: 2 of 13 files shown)


src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -84,22 +125,22 @@ public static bool IsProxyType(Type type)
...
-		/// Checks to see if a method is accessible to DynamicProxyGenAssembly2.</summary>
-		/// <param name="method">The method to check</param>
-		/// <returns><c>true</c> if the method is accessible, <c>false</c> otherwise</returns>
+		/// Checks whether the specified method is accessible to DynamicProxy.</summary>
+		/// <param name="method">The method to check.</param>
+		/// <returns><c>true</c> if the method is accessible to DynamicProxy, <c>false</c> otherwise.</returns>
 		public static bool IsAccessible(MethodBase method)
 		{
-			return method.IsAccessible();
+			return IsAccessibleMethod(method) && IsAccessibleType(method.DeclaringType);
```
stakx: This is the actual bug fix.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+		{
+			return method.IsAssembly || (method.IsFamilyAndAssembly && !method.IsFamilyOrAssembly);
+		}
+
+		/// <summary>
+		/// Provides instructions that a user could follow to make a type or method in <paramref name="targetAssembly"/>
+		/// visible to DynamicProxyGenAssembly2.</summary>
+		/// <param name="targetAssembly">The assembly containing the type or method.</param>
+		/// <returns>Instructions that a user could follow to make a type or method visible to DynamicProxyGenAssembly2.</returns>
+		internal static string CreateInstructionsToMakeVisible(Assembly targetAssembly)
```
jonorossi: I think this method should go into `ExceptionMessageBuilder` instead, otherwise `ProxyUtil` is going to get big quickly.

The other option would be to keep `InternalsUtil` which can hold the implementation and `ProxyUtil` just forwards through, we could keep the obsolete public methods at the top of the file for now forwarding to internal methods below. Thoughts?

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+
+			var instructions = String.Format(instructionsFormat,
+				assemblyToBeVisibleTo,
+				GetAssemblyName(targetAssembly),
+				strongNamedOrNotIndicator);
+			return instructions;
+
+			string GetAssemblyName(Assembly ta)
+			{
+				return ta.GetName().Name;
```
jonorossi: I'd just inline this, the method was only used once in the old code.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+
+			bool ReferencesCastleCore(Assembly ia)
+			{
+#if FEATURE_GET_REFERENCED_ASSEMBLIES
+				return ia.GetReferencedAssemblies()
+					.Any(r => r.FullName == Assembly.GetExecutingAssembly().FullName);
+#else
+			// .NET Core does not provide an API to do this, so we just fall back to the solution that will definitely work.
+			// After all it is just an exception message.
+			return false;
```
jonorossi: Indentation.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+		{
+			string strongNamedOrNotIndicator = " not"; // assume not strong-named
+			string assemblyToBeVisibleTo = "\"DynamicProxyGenAssembly2\""; // appropriate for non-strong-named
+
+			if (targetAssembly.IsAssemblySigned())
+			{
+				strongNamedOrNotIndicator = "";
+				assemblyToBeVisibleTo = ReferencesCastleCore(targetAssembly)
+					? "InternalsVisible.ToDynamicProxyGenAssembly2"
+					: '"' + InternalsVisible.ToDynamicProxyGenAssembly2 + '"';
```
jonorossi: I know this was in the old code exactly as is, however do we actually need this, what does the message look like with the extra double quotes?

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+
+		/// <summary>
+		///   Checks whether the specified method is accessible to DynamicProxy.
+		///   Unlike with <see cref="IsAccessible(MethodBase)"/>, the declaring type's accessibility is ignored.
+		/// </summary>
+		/// <param name = "method">The method to check.</param>
+		/// <returns><c>true</c> if the method is accessible to DynamicProxy, <c>false</c> otherwise.</returns>
+		internal static bool IsAccessibleMethod(MethodBase method)
+		{
+			// Accessibility supported by the full framework and CoreCLR
```
jonorossi: This comment can be deleted, "CoreCLR" in this context referred to Silverlight's "CoreCLR".

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+			if (method.IsPublic || method.IsFamily || method.IsFamilyOrAssembly)
+			{
+				return true;
+			}
+
+			if (method.IsFamilyAndAssembly)
```
jonorossi: This could be merged with the previous if statement, actually `method.IsFamilyAndAssembly == true` means the method is accessible to derived classes inside this assembly, meaning we should actually be performing the same `AreInternalsVisibleToDynamicProxy` check below with these.

I guess this hasn't been a problem because I don't think you can define a method like this in C#, `protected internal` gives you `IsFamilyOrAssembly`.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +150,134 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+			{
+				return true;
+			}
+
+			if (AreInternalsVisibleToDynamicProxy(method.DeclaringType.GetTypeInfo().Assembly) && method.IsAssembly)
```
jonorossi: The `method.IsAssembly` check should be first as it is much less expensive.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -15,15 +15,56 @@
...
+				var internalsVisibleTo = asm.GetCustomAttributes<InternalsVisibleToAttribute>();
+				var found = internalsVisibleTo.Any(VisibleToDynamicProxy);
+
+				internalsVisibleToDynamicProxy.Add(asm, found);
+				return found;
+			}
+
+			bool VisibleToDynamicProxy(InternalsVisibleToAttribute attribute)
+			{
+				return attribute.AssemblyName.Contains(ModuleScope.DEFAULT_ASSEMBLY_NAME);
```
jonorossi: I know this was its own method before, but this should just be an inline lambda.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -15,15 +15,56 @@
...
 	public class ProxyUtil
 	{
+		private static readonly IDictionary<Assembly, bool> internalsVisibleToDynamicProxy = new Dictionary<Assembly, bool>();
+		private static readonly Lock internalsVisibleToDynamicProxyLock = Lock.Create();
+
+		/// <summary>
+		///   Determines whether this assembly has internals visible to dynamic proxy.
+		/// </summary>
+		/// <param name = "asm">The assembly to inspect.</param>
+		internal static bool AreInternalsVisibleToDynamicProxy(Assembly asm)
```
jonorossi: Could you move this internal function further down, with the methods for accessibility.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -15,15 +15,56 @@
...
 
-	using Castle.DynamicProxy.Internal;
 #if FEATURE_REMOTING
 	using System.Runtime.Remoting;
 #endif
 
+	using Castle.Core.Internal;
+	using Castle.DynamicProxy.Generators.Emitters;
+
 	public class ProxyUtil
```
jonorossi: Let's go ahead and make this static like it should have been, however don't change the functions back to extension methods.

src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
```diff
@@ -15,135 +15,43 @@
...
-		private static readonly Lock internalsToDynProxyLock = Lock.Create();
-
 		/// <summary>
 		///   Determines whether the specified method is internal.
 		/// </summary>
 		/// <param name = "method">The method.</param>
 		/// <returns>
 		///   <c>true</c> if the specified method is internal; otherwise, <c>false</c>.
 		/// </returns>
+		[EditorBrowsable(EditorBrowsableState.Never)]
```
jonorossi: We should just mark these obsolete, your call if you want to leave EditorBrowsable on there too.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
blairconrad: I'd submit that we do need it. Okay, maybe not "need". The bit after `?` provides a convenience mechanism for assemblies that reference Castle.Core. They can just add 

```c#
[assembly: InternalsVisibleTo("InternalsVisible.ToDynamicProxyGenAssembly2")]
```

instead of providing the full reference to the assembly themselves. The bit after the `:` gives a message like

```
Can not create proxy for type FluentAssertions.Common.NullReflector because it is not accessible. Make it public, or internal and mark your assembly with [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")] attribute, because assembly FluentAssertions.Core is strong-named.
```

a

jonorossi: Thanks @blairconrad, I think I was starring at that code too long, completely missed that it was referencing that field with the strong name key and not just a constant string literal.

All good, it can stay as is.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
stakx: OK!

src/Castle.Core/DynamicProxy/ProxyUtil.cs
stakx: In my opinion, given its somewhat vague name, `InternalsUtil` could easily become a Kitchen Sink Of Lost Methods, and it would perhaps be better to stop putting new stuff in there and get rid of it as soon as is practical.

I think too that `ExceptionMessageBuilder`' would be a good and fitting new home for `CreateInstructionsToMakeVisible`.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Make ProxyUtil.IsAccessible(MethodBase) take into account declaring type's accessibility so it doesn't report false negatives for e.g. public methods in inaccessible classes. (@stakx, #289)
+
```
stakx: Some things are actually missing here:

1. The `ProxyUtil` class having become `static`. (Possibly a breaking change, even though it's very unlikely.)
2. Members of `InternalsUtil` now being obsolete.

Under which heading do you usually list changes that aren't breaking changes?

src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
stakx: Done. I left the `[EditorBrowsable]` attribute, hope you really don't mind. It could in theory be put on the whole class, too, but perhaps you're still going to use that class in the future, so I didn't do that.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -109,17 +116,109 @@ public static bool IsAccessible(MethodBase method, out string message)
...
+		{
+			if (method.IsPublic || method.IsFamily || method.IsFamilyOrAssembly)
+			{
+				return true;
+			}
+
+			if (method.IsAssembly || method.IsFamilyAndAssembly)
+			{
+				return AreInternalsVisibleToDynamicProxy(method.DeclaringType.GetTypeInfo().Assembly);
+			}
```
stakx: Instead of merging the check for `method.IsFamilyAndAssembly` with the first `if` block, I merged it with the latter (as per your comment).

src/Castle.Core/DynamicProxy/Internal/InternalsUtil.cs
jonorossi: > I left [EditorBrowsable] in place though, hope you really don't mind.

Nope, it's all good.

CHANGELOG.md
jonorossi: 1. I wouldn't, if anyone was inheriting this class, then bad luck 😉 
2. It could be breaking if someone re-compiles and has warnings as errors, however I think we just mention it when we actually remove them so we don't bloat the changelog.

> If so, under what heading do you usually list non-breaking changes?

I think I've used "Enhancements" in one of the changelogs but that isn't applicable here obviously. Something to decide when it comes up.

Issue #288, opened 2017-07-13
Castle.Core having minor and patch level version in AssemblyVersion


skovsende:
I had a good discussion with @stakx from the Moq project in this thread: https://github.com/moq/moq4/issues/418

It seems to be Castle.Core does not follow best practice when it comes to AssemblyVersion, which leads to all kinds of problems when people depend on Castle.Core.

We ended up finding a really good explanation of the problem here: https://codingforsmarties.wordpress.com/2016/01/21/how-to-version-assemblies-destined-for-nuget/

Is it something you have any opinions about? If needed a may be able to find some time in the coming weeks to create a PR, but as I believe a major version bump is required, I would like to run the idea by you before doing anything.

fir3pho3nixx:
@skovsende - I did not read the discussion but did the the blog post. I think the blog post obfiscates the problem and the solution in a pile of waffle and is not very well written. Can you summarise this for clarity?

skovsende:
I'll try. The current situation as I see it:

Imaging I have an application, MyApp, that uses two nuget packages: PackageX and Castle.Core

Library PackageX is a nuget package containing a strong named assembly that uses version 4.1.0 of Castle Core. In the nuspec the dependency is set to 4+, to be compatible with new versions as long as Castle.Core follows semantic versioning.

Then version 4.1.1 of Castle.Core is released, and I update MyApp to use the new version of Castle.Core. This gives problems for PackageX, because it's strong named and uses the AssemblyVersion of the Castle.Core dll to make sure it's running against the correct version. A way to fix this is to add assembly redirects in your app.config, which tells that whenever anyone want's 4.1.0.0 - that 4.1.1.0 is usable instead. It works, but it's pretty cumbersome, especially when the amounts of packages increase.

If Castle.Core instead changed it's strategy to set AssemblyVersion to 4.0.0.0 for all nuget package releases with version 4.x, the situation becomes much simple for the consumers of Castle.Core.

In that case MyApp updates the reference, the strong-named PackageX uses the new dll from Castle.Core without problems because the AssemblyVersion is the same as the one it was built against originally.

I have seen a lot of libraries handle AssemblyVersion this way, Serilog and EntityFramework is examples I've run in to and all Microsoft.VisualStudio dlls/packages handle versioning this way as well.

And note that this is only for AssemblyVersion - AssemblyFileVersion would still follow the version of the nuget package, so if you do a Properties->Details on the dll, you would still get the full version number.

Hope that makes my point slightly clearer.

[...]

PR #298, opened 2017-09-12
Fixing assembly patch versions

fir3pho3nixx:
This is so we can avoid assembly binding errors for minor upgrades as per #288

### DIFF
--- a/buildscripts/common.props
+++ b/buildscripts/common.props
@@ -6,14 +6,16 @@
 		<RepositoryUrl>https://github.com/castleproject/Core</RepositoryUrl>
 		<BuildVersion>0.0.0</BuildVersion>
 		<BuildVersion Condition="'$(APPVEYOR_BUILD_VERSION)'!=''">$(APPVEYOR_BUILD_VERSION)</BuildVersion>
+		<BuildVersionMajor>$(BuildVersion.Split('.')[0])</BuildVersionMajor>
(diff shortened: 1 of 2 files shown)


buildscripts/common.props
```diff
@@ -6,14 +6,17 @@
...
 	</PropertyGroup>
 
 	<PropertyGroup>
 		<Product>Castle Core</Product>
-		<Version>$(BuildVersion)</Version>
-		<AssemblyVersion>$(BuildVersion)</AssemblyVersion>
 		<Authors>Castle Project Contributors</Authors>
+		<FileVersion>$(BuildVersion)</FileVersion>
+		<VersionPrefix>$(BuildVersion)</VersionPrefix>
+		<AssemblyVersion>$(BuildVersionMajor).$(BuildVersionMinor).0</AssemblyVersion>
```
jonorossi: I thought we decided to do x.0.0.0? i.e. just major. If we do major.minor then we'll still get problems with very minor release which we've got right now.

> We used to do `x.x.0.0`, but `x.0.0.0` sounds even better. `AssemblyFileVersion` should still have the actual version. (https://github.com/castleproject/Core/issues/288#issuecomment-315387834)

fir3pho3nixx: My bad, all fixed.

jonorossi:
Thanks

Issue #292, opened 2017-08-05
Fix dependency specification of logging packages to the exact same explicit version of Castle.Core package


DianaKoenraadt:
Hi all,

I had some trouble getting Castle NLog integration to work.

I'm using:

Castle.Core v4.1.1
Castle.Core-NLog v3.3.3
Castle.LoggingFacility v4.0.0
Castle.Windsor v4.0.0
Castle.Windsor-NLog v3.4.0
NLog v4.4.1.1
NLog.Config v4.4.1.1
NLog.Schema v4.4.1.1

Now, when I setup a simple installer:

    public class Installer : IWindsorInstaller
    {
        public void Install(IWindsorContainer container, IConfigurationStore store)
        {
            // Install logging facility
            container.AddFacility<LoggingFacility>(f => f.LogUsing(LoggerImplementation.NLog)
                                                         .WithConfig("NLog.config"));
        }
    }

and write a straightforward test

    [TestClass]
    public class InstallerTests
    {
        protected WindsorContainer Container;
        
        [TestInitialize]
        public void Initialize()
        {
            Container = new WindsorContainer();
        }

        [TestMethod]
        public void Install_ShouldInstallLogger()
        {
            // Arrange
            Container.Install(FromAssembly.Containing<Installer>());

            // Act
            ILogger result = Container.Resolve<ILogger>();

            // Assert
            Assert.IsNotNull(result);
        }
    }

I got the error:

**Could not convert string 'Castle.Services.Logging.NLogIntegration.NLogFactory,Castle.Services.Logging.NLogIntegration,Version=4.0.0.0, Culture=neutral,PublicKeyToken=407dd0808d44fbdc' to a type. ---> System.IO.FileLoadException: Could not load file or assembly 'Castle.Services.Logging.NLogIntegration, Version=4.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference.** 

In order to resolve this, I had to add in app.config:

      <dependentAssembly>
        <assemblyIdentity name="Castle.Services.Logging.NLogIntegration" publicKeyToken="407dd0808d44fbdc" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.0.0" newVersion="3.3.0.0" />
      </dependentAssembly>

What is causing this behavior? How can I resolve it in a better way? Or is this the right way?

jonorossi:
[...]

PR #302, opened 2017-09-19
Explicit Package Versions for Castle

fir3pho3nixx:
I tried a number of approaches, this was the cleanest way I could think of doing after backing out previous attempts.

image


jonorossi:
@fir3pho3nixx did you want to update the changelog to include this fix?

fir3pho3nixx:
@jonorossi - Ha! Name and shame myself, just kidding will add it in :)

jonorossi:
Fixes #292.

PR #303, opened 2017-09-19
Fix MethodSignatureComparer to handle nested generics

BitWizJason:
I added test cases that demonstrate the issue in the first commit and a potential fix in the second commit.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
@@ -91,9 +91,20 @@ namespace Castle.DynamicProxy.Generators
 					return false;
 				}
 			}
-			else if (x.GetGenericArguments().Length > 0)
+			else if (x.GetTypeInfo().IsGenericType)
 			{
-				return GenericArugmentsEqual(x, y);
+				var xArgs = x.GetGenericArguments();
+				var yArgs = y.GetGenericArguments();
+
+				if (xArgs.Length != yArgs.Length)
+				{
+					return false;
+				}
+
+				for (var i = 0; i < xArgs.Length; ++i)
+				{
+					 if(!EqualSignatureTypes(xArgs[i], yArgs[i])) return false;
+				}
 			}
 			else
 			{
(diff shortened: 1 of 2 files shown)


src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
```diff
@@ -86,11 +86,15 @@ public bool EqualSignatureTypes(Type x, Type y)
 
 			if (x.GetTypeInfo().IsGenericParameter)
 			{
-				if (x.GenericParameterPosition != y.GenericParameterPosition)
+				if (x.GetTypeInfo().GenericParameterPosition != y.GetTypeInfo().GenericParameterPosition)
 				{
 					return false;
 				}
 			}
+			else if (x.GetGenericArguments().Length > 0)
```
jonorossi: Could this use `x.IsGenericType` instead?

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
```diff
@@ -128,5 +132,36 @@ private bool EqualNames(MethodInfo x, MethodInfo y)
...
+			var yArgs = y.GetGenericArguments();
+
+			if (xArgs.Length != yArgs.Length)
+			{
+				return false;
+			}
+
+			for (var i = 0; i < xArgs.Length; ++i)
+			{
+				if (xArgs[i].GetTypeInfo().IsGenericParameter != yArgs[i].GetTypeInfo().IsGenericParameter)
```
jonorossi: Could this loop just call `EqualSignatureTypes(xArgs[i], yArgs[i])`? This probably means this method isn't needed and the recursion could happen in `EqualSignatureTypes` instead.

If this method stays it is spelt wrong.

BitWizJason:
Thank you @stakx and @jonorossi for reviewing. Pesky typos.... Good suggestions! Updated pull request.

jonorossi:
@BitWizJason Great. Before merging I wanted to make sure you are happy with the email address on those commits, that email address isn't associated with your GitHub account.

BitWizJason:
@jonorossi Thank you for pointing that out. I updated my profile. It should be associated now.

Issue #260, opened 2017-05-11
Reason for __interceptors being public?


martijnburgers:
Hi there,

We are using dynamic proxies with Automapper a lot and came a cross a issue where Automapper is mapping the `__interceptors` field between two proxies. The reason for this is that Automapper maps every public field by default.

Is there a reason the `IInterceptor[] __interceptors` field on a proxy is public?

I can easily fix this with Automapper itself but just wanted to know if there is a valid reason for this field to be public.

jonorossi:
Hi @martijnburgers, I didn't know the answer when you logged this issue and forgot all about it.

The history for why all static and instance fields on proxies are public would probably be a decision from 10+ years ago, but my guess is that it was required to run under partial trust which was common with shared hosting providers of the time.

I was interested to see what unit tests would fail if I changed it, and only the binary serialization ones failed until I made the 2nd change below:

```diff
diff --git a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
index 6ef8c5e..53289dd 100644
--- a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
@@ -159,7 +159,7 @@ namespace Castle.DynamicProxy.Generators
 
 		protected void CreateInterceptorsField(ClassEmitter emitter)
 		{
-			var interceptorsField = emitter.CreateField("__interceptors", typeof(IInterceptor[]));
+			var interceptorsField = emitter.CreateField("__interceptors", typeof(IInterceptor[]), FieldAttributes.Private);
 
 #if FEATURE_SERIALIZATION
 			emitter.DefineCustomAttributeFor<XmlIgnoreAttribute>(interceptorsField);
diff --git a/src/Castle.Core/DynamicProxy/Serialization/ProxyObjectReference.cs b/src/Castle.Core/DynamicProxy/Serialization/ProxyObjectReference.cs
index 49ad305..f0d6e4d 100644
--- a/src/Castle.Core/DynamicProxy/Serialization/ProxyObjectReference.cs
+++ b/src/Castle.Core/DynamicProxy/Serialization/ProxyObjectReference.cs
@@ -310,7 +310,7 @@ namespace Castle.DynamicProxy.Serialization
 
 		private void SetInterceptors(IInterceptor[] interceptors)
 		{
[...]

PR #305, opened 2017-09-24
Make the internal fields used by Castle on dynamic proxies private instead of public

spencercw:
See #260.

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/SerializableClassTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/SerializableClassTestCase.cs
@@ -19,6 +19,7 @@ namespace Castle.DynamicProxy.Tests
 	using System;
 	using System.Collections;
 	using System.IO;
+	using System.Reflection;
 	using System.Runtime.Serialization;
 	using System.Runtime.Serialization.Formatters.Binary;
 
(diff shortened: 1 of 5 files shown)


src/Castle.Core.Tests/DynamicProxy.Tests/SerializableClassTestCase.cs
```diff
@@ -250,8 +251,8 @@ public void ProxyGenerationOptionsRespectedOnDeserialization()
 			Assert.AreEqual(proxy.GetType(), proxy.GetType().GetMethod("get_Current").DeclaringType);
 			Assert.AreNotEqual(proxy.GetType(), proxy.GetType().GetMethod("CalculateSumDistanceNow").DeclaringType);
 			Assert.AreEqual(proxy.GetType().BaseType, proxy.GetType().GetMethod("CalculateSumDistanceNow").DeclaringType);
-			var options2 =
-				(ProxyGenerationOptions)proxy.GetType().GetField("proxyGenerationOptions").GetValue(null);
+			var options2 = (ProxyGenerationOptions)proxy.GetType().
+				GetField("proxyGenerationOptions", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null);
```
fir3pho3nixx: Why is this static a thing? Mostly instance based or have I missed something.

fir3pho3nixx:
From what you have done, it looks good. Could you also update the CHANGELOG.md accordingly.

src/Castle.Core.Tests/DynamicProxy.Tests/SerializableClassTestCase.cs
spencercw: `proxyGenerationOptions` is a static field (see BaseProxyGenerator.cs). If you don't pass any binding flags then you get `Instance|Static|Public` (see [Type.GetField()](https://referencesource.microsoft.com/#mscorlib/system/type.cs,31ecddf2e1a221ae)), so I need to set the flags manually to get it to return the private field.

fir3pho3nixx:
Just get an entry into the CHANGELOG.md and then we are good to go.

spencercw:
@fir3pho3nixx Enhancement or bugfix do you think? I'm not sure it qualifies as a bugfix because it's not strictly broken.

fir3pho3nixx:
I would say enhancement.

spencercw:
Ok that's done

fir3pho3nixx:
Welcome to the club fella! :) Wish we had more people reporting problems and fixing them like you. Most people tend to use this issue tracker as a substitute for Stack Overflow. 

@jonorossi - will have to merge this because he soul surviving owner ATM. 

Thanks again!

spencercw:
Heh, thanks. :) I probably ought to contribute to these things more than I do. This just happened to be a nice easy one to get into.

jonorossi:
Looks good, yes you should contribute more 😉 .

PR #310, opened 2017-10-04
Add missing equality checks to `MethodSignatureComparer.EqualSignatureTypes`

### LINKED ISSUE #309: Regression in 4.2.0 related to `MethodSignatureComparer`
@cbruun just reported a regression in Moq (https://github.com/moq/moq4/issues/469) that looks like it's a regression in Castle Core 4.2.0, caused by #303 (commit 0fe612951276d73f8fe789ebc11478b5a42b7046).

### Repro code:

```csharp
public interface IBreak
{
	void DoStuff<T>(Implementation1<T> a);
	void DoStuff<T>(Implementation2<T> a);
}

public abstract class BaseType<T> { }
public class Implementation1<T> : BaseType<T> { }
public class Implementation2<T> : BaseType<T> { }

// The following call to `CreateClassProxy` worked in 4.1.1, but throws in 4.2.0:
generator.CreateClassProxy(typeof(object), new[] { typeof(IBreak) } /* , someInterceptor */);
```

```
System.TypeLoadException: Method 'DoStuff' in type 'Castle.Proxies.ObjectProxy' from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
   at System.Reflection.Emit.TypeBuilder.TermCreateClass(RuntimeModule module, Int32 tk, ObjectHandleOnStack type)
   at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
   at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateType(TypeBuilder type)
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.BuildType()
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateType(String name, Type[] interfaces, INamingScope namingScope)
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.<>c__DisplayClass1_0.<GenerateCode>b__0(String n, INamingScope s)
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateCode(Type[] interfaces, ProxyGenerationOptions options)
   at Castle.DynamicProxy.DefaultProxyBuilder.CreateClassProxyType(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options)
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, Object[] constructorArguments, IInterceptor[] interceptors)
   at Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, IInterceptor[] interceptors)
   at Program.Main() in …
```

### Probable cause of the regression:

`MethodSignatureComparer.EqualSignatureTypes(x: typeof(Implementation1<T>), y: typeof(Implementation2<T>))` no longer returns `false`. It now returns `true` because it only checks the type arguments for equality, not the actual generic type definition itself.

### Possible solution:

See the PR referenced below.

/cc @BitWizJason

stakx:
Given two generic types, `EqualSignatureTypes` only checks whether their generic type arguments match, but it doesn't check whether the actual generic type definition (i.e. the "open" generic type) matches. This PR adds those missing checks.

This should fix #309.

Please beware: I was a bit pressed for time when I put this PR together, so my understanding of `MethodSignatureComparer`'s purpose and exact uses is incomplete. I've relied mostly on the unit test suite to tell me whether I'm introducing any breaking change or not.

### DIFF
--- a/src/Castle.Core.Tests/MethodComparerTestCase.cs
+++ b/src/Castle.Core.Tests/MethodComparerTestCase.cs
@@ -355,5 +355,28 @@ namespace Castle.DynamicProxy.Tests
 			Assert.IsTrue(mc.Equals(typeof(Base).GetMethod("GenericMethod7"),
 				typeof(Inherited).GetMethod("GenericMethod7")));
 		}
+
+		[Test]
+		public void Compare_two_method_overloads_with_generic_arg_types()
+		{
+			var mc = MethodSignatureComparer.Instance;
+			var methods = typeof(IHaveOverloadedGenericMethod).GetMethods();
+			var firstOverload = methods[0];
+			var secondOverload = methods[1];
+
+			// The overloads must obviously have different signatures, or we could never even have
+			// compiled this test code successfully. The arguments must have a different type:
+			Assert.IsFalse(mc.Equals(firstOverload, secondOverload));
+		}
+
+		private interface IHaveOverloadedGenericMethod
+		{
+			void GenericMethod<T>(GenericClass1<T> arg);
+			void GenericMethod<T>(GenericClass2<T> arg);
+		}
+
+		private class GenericClass1<T> { }
+
+		private class GenericClass2<T> { }
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Add missing equality checks in `MethodSignatureComparer.EqualSignatureTypes` to fix `TypeLoadException`s ("Method does not have an implementation") (@stakx, #310)
+
 ## 4.2.0 (2017-09-28)
 
 Enhancements:
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ExplicitInterfaceTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ExplicitInterfaceTestCase.cs
@@ -176,6 +176,13 @@ namespace Castle.DynamicProxy.Tests
 
 			Assert.AreEqual(5, result);
 		}
+
+		[Test]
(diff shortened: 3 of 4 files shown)


stakx:
Please don't merge just yet. I need to add a regression test that shows that this actually fixes the linked issue.

src/Castle.Core.Tests/MethodComparerTestCase.cs
```diff
@@ -355,5 +355,28 @@ public void Compare_inherited_double_nested_return_method()
...
+
+		private interface IHaveOverloadedGenericMethod
+		{
+			void GenericMethod<T>(GenericClass1<T> arg);
+			void GenericMethod<T>(GenericClass2<T> arg);
+		}
+
+		private class GenericClass1<T> { }
+
+		private class GenericClass2<T> { }
```
stakx: I felt it was important to keep the two tests separate:

* The unit test here deals with some internal cogwheel;
* The regression unit test acts at the level of the public API.

Yet, declaring the same test types twice seems somewhat repetitive. Do you mind? (If so, where would be the best place to put them?)

jonorossi: > Yet, declaring the same set of test types (the interface, plus GenericClass1<T> and GenericClass2<T>) twice seems somewhat repetitive. Do you mind? (If so, where would be the best place to put them?)

Two tests sounds fine to me, means if the internal implementation changes we'll at least still have the regression one.

PR #314, opened 2017-10-09
Harmonising core and windsor

fir3pho3nixx:
- Ignoring the xml files, they are coming out in the root of each project.
 - Also adding docs to the solution, this is handy if you use visual studio.


fir3pho3nixx:
Hold on, something has gone wrong

PR #315, opened 2017-10-09
Fix XML documentation generation and add Markdown docs to VS solution

fir3pho3nixx:
Bringing docs in on both sides and ignoring unstaged xml doc files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,10 @@ project.lock.json
 /.dotnet/
 tools/**/obj/
 tools/**/build/
+
+# Docs artifacts
+docs/obj/
+docs/bin/
+
+# Xml doc/test files
+*.xml
--- a/.gitignore
+++ b/.gitignore
@@ -7,20 +7,10 @@ _ReSharper*/
 
 # Build output
 build/
-artifacts/
-src/**/obj/
-src/**/bin/
-buildscripts/**/obj/
-buildscripts/**/bin/
-project.lock.json
-/.dnx/
 /.dotnet/
-tools/**/obj/
+**/obj/**
+**/bin/**
 tools/**/build/
-
-# Docs artifacts
-docs/obj/
-docs/bin/
 
 # Xml doc/test files
 *.xml
--- a/Castle.Core.sln
+++ b/Castle.Core.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26403.7
+VisualStudioVersion = 15.0.26730.16
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Build", "Castle Build", "{1B999D24-B7AB-4997-96E7-08FA05325694}"
 	ProjectSection(SolutionItems) = preProject
@@ -30,6 +30,10 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Services.Logging.SerilogIntegration", "src\Castle.Services.Logging.SerilogIntegration\Castle.Services.Logging.SerilogIntegration.csproj", "{91B2A82F-63F6-46B1-8EDC-5D029BCF6A2B}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Services", "Castle Services", "{A598EE9B-41CE-4BE8-BF93-2C91F919F97E}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Docs", "Castle Docs", "{EC3FDC8C-1D6C-49DA-AD19-E1521A75728E}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Documentation", "docs\Documentation.csproj", "{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -61,6 +65,10 @@ Global
 		{91B2A82F-63F6-46B1-8EDC-5D029BCF6A2B}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{91B2A82F-63F6-46B1-8EDC-5D029BCF6A2B}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{91B2A82F-63F6-46B1-8EDC-5D029BCF6A2B}.Release|Any CPU.Build.0 = Release|Any CPU
+		{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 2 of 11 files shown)


Castle.Core.sln
```diff
@@ -31,6 +31,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Services.Logging.Ser
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Services", "Castle Services", "{A598EE9B-41CE-4BE8-BF93-2C91F919F97E}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Docs", "Castle Docs", "{EC3FDC8C-1D6C-49DA-AD19-E1521A75728E}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Documentation", "docs\Documentation.csproj", "{5F7AA0D3-FE16-40BD-9492-7B8BD509FD57}"
```
jonorossi: The new tooling allows project files to support wildcards, do solution files not support wildcards for files?

.gitignore
```diff
@@ -17,3 +17,14 @@ project.lock.json
...
+# Docs artifacts
+docs/obj/
+docs/bin/
+
+# Xml doc/test files
+*.xml
+src/Castle.Core/Castle.Core.xml
+src/Castle.Services.Logging.NLogIntegration/Castle.Services.Logging.NLogIntegration.xml
+src/Castle.Services.Logging.SerilogIntegration/Castle.Services.Logging.SerilogIntegration.xml
+src/Castle.Services.Logging.log4netIntegration/Castle.Services.Logging.log4netIntegration.xml
```
jonorossi: Oh, don't these usually get written to the output `bin` directory next to the DLL and PDB files? That is what VS used to default to before the new CLI tooling.

Castle.Core.sln
fir3pho3nixx: It would be great if they sorted solution files out, but I have not come across anything that allows that behaviour in VS latest. 

Nothing [here](https://visualstudio.uservoice.com/forums/121579-visual-studio-ide?query=solution) either

.gitignore
fir3pho3nixx: I was almost sure that [OutputPath](https://github.com/castleproject/Core/blob/master/src/Castle.Core/Castle.Core.csproj#L22) was the way to do it. But it isn't working here for some reason.

jonorossi: I'm not sure what is different to Windsor because Windsor 4.1.0 has XML files for each target framework in their bin/debug or bin/release and in the NuGet packages, and it doesn't specify the `DocumentationFile` element.

We need to fix this because I just realised that if the file is getting written here than wouldn't that mean that the last target to get built becomes the output for all targets, i.e. if netstandard1.6 is built last than that's the docs for all targets.

fir3pho3nixx: Will have a fix for this soon.

src/Castle.Core/Castle.Core.csproj
```diff
@@ -19,6 +19,7 @@
 		<PublicSign Condition="'$(OS)'=='Unix'">true</PublicSign>
 		<FrameworkPathOverride Condition="'$(OS)'=='Unix'">$(NuGetPackageFolders)microsoft.targetingpack.netframework.v4.6.1\1.0.1\lib\net461\</FrameworkPathOverride>
 		<PackageTags>castle dynamicproxy dynamic proxy dynamicproxy2 dictionaryadapter emailsender</PackageTags>
+		<OutputPath>bin\$(Configuration)\$(TargetFramework)\</OutputPath>
 		<DocumentationFile>$(OutputPath)Castle.Core.xml</DocumentationFile>
```
jonorossi: > Will have a fix for this soon.

I thought the fix was going to be removing the `DocumentationFile` since Windsor has XML docs in its packages without this element.

fir3pho3nixx: Yes, it is actually

```
<GenerateDocumentationFile>true</GenerateDocumentationFile>
```

I must have missed that earlier.

.gitignore
```diff
@@ -17,3 +17,10 @@ project.lock.json
 /.dotnet/
 tools/**/obj/
 tools/**/build/
+
+# Docs artifacts
+docs/obj/
+docs/bin/
```
jonorossi: Maybe we should just ignore obj and bin directories everywhere, we've got 4 sets of these patterns now.

.gitignore
```diff
@@ -17,3 +17,10 @@ project.lock.json
...
+
+# Xml doc/test files
+*.xml
```
jonorossi: I don't think ignoring all XML files everywhere is a good idea, this isn't needed anymore anyway.

jonorossi:
@fir3pho3nixx ping

fir3pho3nixx:
Ha! This PR is like death by a thousand cuts. What do we need? just updating .gitignore?

.gitignore
fir3pho3nixx: Mind you, this is only the root folder. It also prevents `DesktopClrTestResults.xml` and `NetCoreClrTestResults.xml` from being unstaged and accidentally committed.

fir3pho3nixx:
@jonorossi pong

.gitignore
```diff
@@ -7,13 +7,10 @@ _ReSharper*/
...
-project.lock.json
-/.dnx/
 /.dotnet/
-tools/**/obj/
+**/obj/**
+**/bin/**
 tools/**/build/
+
+# Xml doc/test files
+*.xml
```
jonorossi: > Mind you, this is only the root folder. It also prevents DesktopClrTestResults.xml and NetCoreClrTestResults.xml from being unstaged and accidentally committed.

@fir3pho3nixx I think you misunderstand how gitignores work, sorry for this pull request being death by a thousand cuts.

> If the pattern does not contain a slash /, Git treats it as a shell glob pattern and checks for a match against the pathname relative to the location of the .gitignore file (relative to the toplevel of the work tree if not from a .gitignore file).
>
> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).
>
> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".
>
> (https://git-scm.com/docs/gitignore)

`*.xml` is a glob pattern across all directories (like the `*.suo` pattern), it would need to be `/*.xml` to just match the root.

There is also no need for the `**`, you can just do `bin/` to match the directory rather the contents.

fir3pho3nixx: Okiday, I checked the globs by running git status and it appeared to be OK. Will fix.

jonorossi: Cheers, I know I'm a nitpicking bastard 😉 

P.S. this PR now also conflicts because of another PR I merged earlier.

fir3pho3nixx: Don't worry I will sort that.

fir3pho3nixx: @jonorossi - I think I misled you earlier when I said the XML artifacts are only in the root. 

I cleaned the project, using `git clean -x -f -d`, then changed the `.gitignore` to have a leading slash in front of the `*.xml` -> `/*.xml`. 

After running a full build and tests, I then double checked using `git status` with the following result. 

```
Your branch is up-to-date with 'origin/harmonising-core-and-windsor-2'.

All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)

Changes to be committed:

        modified:   CHANGELOG.md
        modified:   src/Castle.Services.Logging.log4netIntegration/Castle.Services.Logging.log4netIntegration.csproj

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   .gitignore
        modified:   Castle.Core.sln.DotSettings

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        src/Castle.Services.Logging.log4netIntegration/Castle.Services.Logging.Log4netIntegration.xml
``` 

Notice how git thinks it needs to add the xml build artifact from the Log4netIntegration?

I believe my original approach is correct because this is a build artifact. It was my fault for not being clear on this in the first place but I do test my work before committing(admittedly I don't always test it on linux hence the previous casing issue). I am also puzzled as to why the casing issue cropped up and was not detected by Travis. Bizarre. 

The globbing pattern I thought would be OK because of this:

```
Two consecutive asterisks ("**") in patterns matched against full pathname may have special meaning:

A leading "**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo". "**/foo/bar" matches file or directory "bar" anywhere that is directly under directory "foo". <- !! Why I put the glob on the front of bin/obj. !!

A trailing "/**" matches everything inside. For example, "abc/**" matches all files inside directory "abc", relative to the location of the .gitignore file, with infinite depth. <- !! Why I put the glob on the end !!

A slash followed by two consecutive asterisks then a slash matches zero or more directories. For example, "a/**/b" matches "a/b", "a/x/b", "a/x/y/b" and so on.

Other consecutive asterisks are considered invalid.
```

You might consider it unnecessary although it is still valid. I just failed to explain myself perhaps properly. 

I am going to fix the merge conflict and add `Castle.Core.sln.DotSettings` to the gitignore. 

If none of this is good for you, my alternative is to close out the PR. I am done wasting time on this.

fir3pho3nixx:
I will harmonise the projects after we get this through: https://github.com/castleproject/Windsor/pull/351

jonorossi:
I've committed 5a9bcab6e7387f28005d48992df0678f77b79242 that fixes XML documentation file generation to use `GenerateDocumentationFile` to avoid the [race condition](https://github.com/dotnet/sdk/issues/1599) between building multiple TFMs, and also fixes shipping the same XML file for each TFM because each one overwrites the previous.

Sorry for not realising you were talking about 3 different XML files from the unit test execution, we obviously want the XML doc files written to the bin directories paired with their DLL and PDB. I've explicitly ignored those 3 other XML files so if we add an XML file for something else it'll show up to be committed.

PR #316, opened 2017-10-12
Only write back modified invocation arguments to by-ref parameters of intercepted method

### LINKED ISSUE #311: Interceptor cannot modify captured variable if it is also passed by-ref to intercepted method
The following test (running against Castle Core 4.2.0) demonstrates that any code executed during a proxy method interception cannot modify a captured variable if that captured variable was also passed by-ref to the intercepted method:

```csharp
[TestCase(false)] // passes
[TestCase(true)]  // does not pass
public void InterceptorCanModifyCapturedVariable(bool passCapturedVariableByRef)
{
    // Arrange:
    int capturedVariable = 0;
    var interceptor = new Interceptor(invocation =>
    {
        capturedVariable = 1;
    });
    var voo = generator.CreateInterfaceProxyWithoutTarget<IVoo>(interceptor);

    // Act:
    if (passCapturedVariableByRef)
    {
        voo.Do(ref capturedVariable);
    }
    else
    {
        int otherVariable = 0;
        voo.Do(ref otherVariable);
    }

    // Assert:
    Assert.AreEqual(1, capturedVariable);
}

public interface IVoo
{
    void Do(ref int arg);
}

public class Interceptor : IInterceptor
{
    private Action<IInvocation> callback;

    public Interceptor(Action<IInvocation> callback)
    {
        this.callback = callback;
    }

    {
        this.callback?.Invoke(invocation);
    }
}
```

Contrast this with plain C#, where the same scenario (modifying a captured variable that also gets passed by-ref) works just fine:

```csharp
[Test] // passes
public static void LambdaCanModifyCapturedVariableWhenPassedByRef()
{
    // Arrange:
    int capturedVariable = 0;
    DoAction vooDo = (ref int arg) =>
    {
        capturedVariable = 1;
    }

    // Act:
    vooDo(ref capturedVariable);

    // Assert:
    Assert.AreEqual(1, capturedVariable);
}

delegate void DoAction(ref int x);
```

I suspect this inconsistency is caused by DynamicProxy copying all arguments into an `object[] arguments` array for the `IInvocation` passed to the interceptor, then after the interception, it will copy back values from that array (they might have been modified with `IInvocation.SetArgumentValue`) into all by-ref parameters, effectively overwriting any changes that might have been applied behind DynamicProxy's back to the variable.

The problem here is that the interceptor has no way of knowing that the variable it has captured corresponds to a `ref` parameter, and that it can only change the captured variable by also setting the argument value.

Is there a misunderstanding on my part, or is this a feature, a fixable bug, or a known limitation?

(If it's a bug, and if my above suspicion is more or less correct, then perhaps `IInvocation.SetArgumentValue` should additionally set some "dirty flag" so that only parameters get written back that were actually modified.)

stakx:
This fixes #311, as far as is possible. I'll admit that I've come to question whether this is a good enhancement to implement, for these two reasons:

1. It is only a partial fix for the problem described in #311, because changes to a captured variable (that is passed to an intercepted method via a `ref` parameter) won't be reflected in the arguments seen via `IInvocation`, simply because arguments are only read once (at the beginning of an interception) and then "buffered" in an `arguments` array; they don't get "refreshed" from the original arguments. This can't be easily changed, so we will see differences in variable capture / modification behaviour of regular C# vs. DynamicProxy intercepted methods even after this is merged.

2. The price we pay (in `AbstractInvocation`) for the feature added here is quite large: We need to keep a second copy of the passed-in `arguments` array around (`originalArguments`) just so we can later compare whether any arguments were changed in the meantime. Granted, this copy is only made if there are any by-ref parameters, but still.

   My first approach (which would've been way more efficient and perfectly sufficent for most real-world scenarios) was to have an additional `int argumentDirtyFlags` field in `AbstractInvocation` storing a "dirty" flag for the first 32 parameters. (Any additional parameters would be considered "dirty" by default. This is a tradeoff between correctness and efficiency.) `SetArgumentValue` would then set those "dirty" flags. Unfortunately, this method can be bypassed by setting arguments directly via the `object[] Arguments` property.

   It would still be possible to go with a "dirty" flags approach by setting *all* dirty flags if someone queries the `Arguments` property (because once someone has a reference to the `arguments` array, any and all modifications could happen to parameters without us tracking them, so we'd need to assume the worst).

I fully understand if you don't want to merge this after you've read the above.

However, if you'd like to see this go through, I'm happy to refine this some more.

### DIFF
--- a/src/Castle.Core/DynamicProxy/AbstractInvocation.cs
+++ b/src/Castle.Core/DynamicProxy/AbstractInvocation.cs
@@ -15,13 +15,17 @@
 namespace Castle.DynamicProxy
 {
 	using System;
+	using System.Collections.Generic;
 	using System.Diagnostics;
 	using System.Reflection;
+
+	using Castle.DynamicProxy.Generators.Emitters;
 
 	public abstract class AbstractInvocation : IInvocation
 	{
 		private readonly IInterceptor[] interceptors;
 		private readonly object[] arguments;
+		private readonly object[] originalArguments;
 		private int currentInterceptorIndex = -1;
 		private Type[] genericMethodArguments;
 		private readonly MethodInfo proxiedMethod;
@@ -38,6 +42,15 @@ namespace Castle.DynamicProxy
 			this.interceptors = interceptors;
 			this.proxiedMethod = proxiedMethod;
 			this.arguments = arguments;
+
+			var parameters = proxiedMethod.GetParameters();
+			if (ArgumentsUtil.IsAnyByRef(proxiedMethod.GetParameters()))
+			{
+				Debug.Assert(arguments != null, "arguments != null");
+				var argumentCount = arguments.Length;
+				this.originalArguments = new object[argumentCount];
+				Array.Copy(arguments, this.originalArguments, argumentCount);
+			}
 		}
 
 		public void SetGenericMethodArguments(Type[] arguments)
@@ -95,6 +108,11 @@ namespace Castle.DynamicProxy
 		public object GetArgumentValue(int index)
 		{
 			return arguments[index];
+		}
+
+		public bool CouldArgumentValueHaveChanged(int index)
+		{
+			return originalArguments == null || !Equals(originalArguments[index], arguments[index]);
 		}
 
 		public void Proceed()
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/IfStatement.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/IfStatement.cs
@@ -1,0 +1,39 @@
+﻿// Copyright 2004-2017 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Generators.Emitters.SimpleAST
+{
+	using System.Reflection.Emit;
+
+	public class IfStatement : Statement
+	{
+		private readonly Expression condition;
+		private readonly Statement then;
+
+		public IfStatement(Expression condition, Statement then)
+		{
+			this.condition = condition;
+			this.then = then;
+		}
+
+		public override void Emit(IMemberEmitter member, ILGenerator gen)
+		{
+			var fi = gen.DefineLabel();
+			condition.Emit(member, gen);
+			gen.Emit(OpCodes.Brfalse_S, fi);
+			then.Emit(member, gen);
+			gen.MarkLabel(fi);
+		}
+	}
+}
--- a/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
@@ -40,7 +40,10 @@ namespace Castle.DynamicProxy.Generators
 					continue;
 				}
 
-				emitter.CodeBuilder.AddStatement(AssignArgument(dereferencedArguments, i, arguments));
+				emitter.CodeBuilder.AddStatement(
+					new IfStatement(
+						condition: new MethodInvocationExpression(invocation, InvocationMethods.CouldArgumentValueHaveChanged, new LiteralIntExpression(i)),
+						then: AssignArgument(dereferencedArguments, i, arguments)));
 			}
 		}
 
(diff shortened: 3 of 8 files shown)


src/Castle.Core/DynamicProxy/AbstractInvocation.cs
```diff
@@ -38,6 +42,15 @@ public abstract class AbstractInvocation : IInvocation
 			this.interceptors = interceptors;
 			this.proxiedMethod = proxiedMethod;
 			this.arguments = arguments;
+
+			var parameters = proxiedMethod.GetParameters();
+			if (ArgumentsUtil.IsAnyByRef(proxiedMethod.GetParameters()))
```
stakx: It's occurred to me that it's probably not very clever to have this check executed upon every invocation. Perhaps the `AbstractInvocation` constructor should have an additional `bool proxiedMethodHasAnyByRefParameters` parameter that gets passed down from the custom-built invocation types (via `CompositionInvocation`); meaning the check would only run once per type generation, instead of once per invocation.

src/Castle.Core/DynamicProxy/AbstractInvocation.cs
```diff
@@ -15,13 +15,17 @@
...
 	using System.Diagnostics;
 	using System.Reflection;
 
+	using Castle.DynamicProxy.Generators.Emitters;
+
 	public abstract class AbstractInvocation : IInvocation
 	{
 		private readonly IInterceptor[] interceptors;
 		private readonly object[] arguments;
+		private readonly object[] originalArguments;
```
stakx: This is the expensive bit. We end up keeping two copies of the whole argument list (when there are by-ref parameters). It would've been much more efficient to have a set of dirty flags, one per parameter. (See PR description above.)

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/IfStatement.cs
```diff
@@ -0,0 +1,39 @@
...
+		{
+			this.condition = condition;
+			this.then = then;
+		}
+
+		public override void Emit(IMemberEmitter member, ILGenerator gen)
+		{
+			var fi = gen.DefineLabel();
+			condition.Emit(member, gen);
+			gen.Emit(OpCodes.Brfalse_S, fi);
```
stakx: Using a `brfalse.s` opcode will work fine in practice, but it's a risk to take if someone will reuse the `IfStatement` class and pass a really large `then` statement.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -40,7 +40,10 @@ public static class GeneratorUtil
 					continue;
 				}
 
-				emitter.CodeBuilder.AddStatement(AssignArgument(dereferencedArguments, i, arguments));
+				emitter.CodeBuilder.AddStatement(
+					new IfStatement(
+						condition: new MethodInvocationExpression(invocation, InvocationMethods.CouldArgumentValueHaveChanged, new LiteralIntExpression(i)),
+						then: AssignArgument(dereferencedArguments, i, arguments)));
```
stakx: This is the important bit. Writing back arguments to the `ref` or `out` parameters is skipped in cases where their values still match the ones originally passed in (and stored in `originalArguments`).

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/IfStatement.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Castle.DynamicProxy.Generators.Emitters.SimpleAST
+{
+	using System.Reflection.Emit;
+
+	public class IfStatement : Statement
+	{
+		private readonly Expression condition;
+		private readonly Statement then;
+
+		public IfStatement(Expression condition, Statement then)
```
stakx: Could later be expanded to support an `else` clause, too. For now, it's not needed.

jonorossi:
> I'll admit that I've come to question whether this is a good enhancement to implement, for these two reasons

I agree. Ignoring any performance problems in point 2, I think point 1 is a deal breaker, we are better off not supporting this scenario at all and documenting it rather than having strange edge cases. The implementation is also relying on equality which isn't always reliable and also isn't what C# does.

I don't think there is anyway we can allow writing to ref parameters like C# because ref-ness can only be defined on formal parameters so we'd need to invoke a custom `Intercept` method that had ref parameters to allow the interceptor to actually modify the locals of the caller. No idea if there is any CLR magic we can do, but the locals must still be in scope otherwise the IL won't be verifiable.

stakx:
> I don't think there is anyway we can allow writing to ref parameters like C# 

I've reached the same conclusion. :-/

> No idea if there is any CLR magic we can do

It *might* be possible in IL, using pointers and pinning via `GCHandle`, but like you said, I also suspect the result will be unverifiable, so that route's not a great one to follow.

I think we both agree that this unmerged PR and the issue can be closed. It's been fascinating delving into & learning about DynamicProxy's IL generation facilities, so I'm happy to leave it at that.

jonorossi:
@stakx yes, they can be closed. Thanks for jumping head first, you are definitely getting right into it.

It would be good to get something into the docs about ref parameters though, did you want to add something?

PR #332, opened 2018-01-17
Add documentation about by-ref parameter behavior

stakx:
Hi @jonorossi, as promised in #311 a while ago, here's some additional documentation about by-ref parameters. See it [here](https://github.com/stakx/Castle.Core/blob/docs-ref-args/docs/dynamicproxy-by-ref-parameters.md) in rendered form.

I'm perhaps not terribly good at writing short and concise documentation, so any edit suggestions are welcome! :)


jonorossi:
Looks great, thanks @stakx.

PR #335, opened 2018-01-24
Fixing #334 by making AttributesToAvoidReplicating thread safe

### LINKED ISSUE #334: Static AttributesToAvoidReplicating is not Thread Safe
We rarely get the following exception (Castle.Core 4.2.1):

> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
>    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
>    at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
>    at System.Collections.Generic.List`1.Enumerator.MoveNext()
>    at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
>    at Castle.DynamicProxy.Generators.AttributesToAvoidReplicating.ShouldAvoid(Type attribute)
>    at Castle.DynamicProxy.Internal.AttributeUtil.ShouldSkipAttributeReplication(Type attribute)
>    at Castle.DynamicProxy.Internal.AttributeUtil.<GetNonInheritableAttributes>d__5.MoveNext()
>    at Castle.DynamicProxy.Contributors.CompositeTypeContributor.ImplementMethod(MetaMethod method, ClassEmitter class, ProxyGenerationOptions options, OverrideMethodDelegate overrideMethod)
>    at Castle.DynamicProxy.Contributors.CompositeTypeContributor.Generate(ClassEmitter class, ProxyGenerationOptions options)
>    at Castle.DynamicProxy.Generators.InterfaceProxyWithoutTargetGenerator.GenerateType(String typeName, Type proxyTargetType, Type[] interfaces, INamingScope namingScope)
>    at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.<>c__DisplayClass6_0.<GenerateCode>b__0(String n, INamingScope s)
>    at Castle.DynamicProxy.Generators.BaseProxyGenerator.ObtainProxyType(CacheKey cacheKey, Func`3 factory)
>    at Castle.DynamicProxy.Generators.InterfaceProxyWithTargetGenerator.GenerateCode(Type proxyTargetType, Type[] interfaces, ProxyGenerationOptions options)
>    at Castle.DynamicProxy.DefaultProxyBuilder.CreateInterfaceProxyTypeWithoutTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options)
>    at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithoutTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, IInterceptor[] interceptors)

Now, the obvious error case would be a concurrent call to `AttributesToAvoidReplicating.Add<T>()` or `AttributesToAvoidReplicating.Add(Type attribute)`.

As it happens we're also using MOQ which is in fact [modifying this collection](https://github.com/moq/moq4/blob/612c6fae5d8695d6d13cc028e1cc0ec86e76dca8/Source/ProxyFactories/CastleProxyFactory.cs#L74-L87).
Given that our use of MOQ happens concurrent to our own code creating proxies it's, IMHO, quite clear that this can occasionaly cause the above exception.

Even though MOQ and our own code don't share the same `ProxyGenerator` instance, as `AttributesToAvoidReplicating` is static it's being shared.
The only way (workaround) to synchronize access in this case is to cause MOQ to initialize before use Castle.Core in our own code. Of course, that's quite the design smell.

Two alternative suggestions for fixing the issue:

- make `AttributesToAvoidReplicating` part of a specific `ProxyGenerator` Instance. Means attributes need to be supplied when creating it. Also means; 
  - breaking change
  - `AttributesToAvoidReplicating` configuration not leaking to other users of DynamicProxy
  - don't know how this will affect the cleanliness of the implementation itself, as the information would have to be passed around since it can't be static anymore...
- make `AttributesToAvoidReplicating` thread safe
  - suggestion: use `Interlocked.CompareExchange` in `Add` to replace the backing collection.
  - performance overhead should be negliglible as `Add` is only called a few times...

BrunoJuchli:
This PR fixes #334 by making `AttributesToAvoidReplicating` thread safe by replacing the backing collection instead of updating it (...while consumers are enumerating it...).

It employs `Interlocked.CompareExchange` to replace the collection if it's still the same as the original. If replacing fails (because another thread has already exchanged the collection) it retries until it succeeds.

Alternative would have been to replace the backing collection but to `lock` the `Add` method so concurrent `Add`-ing would be sequentialised.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
@@ -41,18 +41,18 @@ namespace Castle.DynamicProxy.Generators
 
 		public static void Add(Type attribute)
 		{
+			// notes:
+			// this class is made thread-safe by replacing the backing list rather than adding to it
+			// this loop runs until the attributes collection contains the new attribute
+			// it is necessary to ensure that if multiple threads concurrently replace the list that they
+			// don't override each others changes
 			IList<Type> originalAttributes;
-			IList<Type> newAttributes;
-			do
+			while (!(originalAttributes = attributes).Contains(attribute))
 			{
-				originalAttributes = attributes;
-				if (originalAttributes.Contains(attribute))
-				{
-					return;
-				}
+				IList<Type> newAttributes = new List<Type>(originalAttributes) { attribute };
 
-				newAttributes = new List<Type>(originalAttributes) { attribute };
-			} while (!ReferenceEquals(Interlocked.CompareExchange(ref attributes, newAttributes, originalAttributes), originalAttributes));
+				Interlocked.CompareExchange(ref attributes, newAttributes, originalAttributes);
+			};
 		}
 
 		public static void Add<T>()
--- a/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
@@ -18,10 +18,11 @@ namespace Castle.DynamicProxy.Generators
 	using System.Collections.Generic;
 	using System.Linq;
 	using System.Reflection;
-	using System.Threading;
 
 	public static class AttributesToAvoidReplicating
 	{
+		private static readonly object lockObject = new object();
+
 		private static IList<Type> attributes;
 
 		static AttributesToAvoidReplicating()
@@ -41,18 +42,11 @@ namespace Castle.DynamicProxy.Generators
 
 		public static void Add(Type attribute)
 		{
-			// notes:
-			// this class is made thread-safe by replacing the backing list rather than adding to it
-			// this loop runs until the attributes collection contains the new attribute
-			// it is necessary to ensure that if multiple threads concurrently replace the list that they
-			// don't override each others changes
-			IList<Type> originalAttributes;
-			while (!(originalAttributes = attributes).Contains(attribute))
+			// note: this class is made thread-safe by replacing the backing list rather than adding to it
+			lock (lockObject)
 			{
-				IList<Type> newAttributes = new List<Type>(originalAttributes) { attribute };
-
-				Interlocked.CompareExchange(ref attributes, newAttributes, originalAttributes);
-			};
+				attributes = new List<Type>(attributes) { attribute };
+			}
 		}
 
 		public static void Add<T>()
--- a/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
@@ -15,32 +15,44 @@
 namespace Castle.DynamicProxy.Generators
 {
 	using System;
-	using System.Reflection;
 	using System.Collections.Generic;
 	using System.Linq;
+	using System.Reflection;
+	using System.Threading;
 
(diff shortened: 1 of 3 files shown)


stakx:
Looks good to me. One question about `Add`, line 50:

```csharp
while (!(originalAttributes = attributes).Contains(attribute))
```

Does the runtime guarantee in this case that `attributes` will be re-read on each iteration, even without a memory barrier or when `attributes` isn't declared `volatile`?

src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
```diff
@@ -18,10 +18,11 @@ namespace Castle.DynamicProxy.Generators
 	using System.Collections.Generic;
 	using System.Linq;
 	using System.Reflection;
-	using System.Threading;
 
 	public static class AttributesToAvoidReplicating
 	{
+		private static readonly object lockObject = new object();
```
stakx: Is this needed? I think you could just lock on `attributes` directly.

src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
```diff
@@ -41,18 +42,11 @@ static AttributesToAvoidReplicating()
...
 		{
-			// notes:
-			// this class is made thread-safe by replacing the backing list rather than adding to it
-			// this loop runs until the attributes collection contains the new attribute
-			// it is necessary to ensure that if multiple threads concurrently replace the list that they
-			// don't override each others changes
-			IList<Type> originalAttributes;
-			while (!(originalAttributes = attributes).Contains(attribute))
+			// note: this class is made thread-safe by replacing the backing list rather than adding to it
+			lock (lockObject)
```
stakx: I think you also need to put `lock`s around places where `attributes` is accessed for reading; that is, in `Contains` and in `ShouldAvoid`. (Not sure about the former, but almost certain in case of the latter.)

BrunoJuchli: IMO not necessary. Exchanging object references is atomic. So `Contains` and `ShouldAvoid` will either get the old reference or the new reference.
Bug #334 is fixed by just replacing the list instead of extending it. However, the lock is necessary to safeguard against concurrent calls to `Add`. Even though they are not very likely, of course.

src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
BrunoJuchli: As the list is not accessible externally of the class this is of course a viable alternative.

The only reason not to do so is to prevent future issues where collections could be made available externally and an external consumer could lock it.

I'm fine either way. I won't go around locking other "people's" objects, so I most likely won't be dealing with that issue, should it ever arise ;-)

jonorossi:
Thanks @BrunoJuchli, merged.

src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs
stakx: Oops, I mis-read the diff (or rather, I didn't read all of it). You're absolutely right, of course.

PR #336, opened 2018-01-24
Update Readme.md - note where Symbols can be found

BrunoJuchli:
Note that symbols package for releases can be found on AppVeyor


jonorossi:
Thanks

Issue #145, opened 2016-03-02
Asynchronous Interceptors


ghost:
Are async interceptors in the pipeline for this library?

This would be such a good feature, as it would allow for async actions before the invocation is proceeded.

I would not mind contributing to this, as long as I could get some architecture advice.

hammett:
Not sure what you mean? 
We have adjusted some of our infrastructure to use async by inspecting the "awaitable" of a return type, and if detected, add a continuation. Not sure this needs to be directly supported by the interceptor, but I'm interested in your proposal.

joelweiss:
Even if I try to hook up a continuation like suggested on [StackOverflow](http://stackoverflow.com/questions/14288075/intercept-the-call-to-an-async-method-using-dynamicproxy) it will still not work properly if I want to call in the continuation `invocation.Proceed()` because the interceptor index gets decremented [here](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/AbstractInvocation.cs#L141), so it will call the first inceptor again and again.

hammett:
I dont think calling proceed in the continuation is a good idea anyways, you wont get a Task instance until you run the actual method.

Looking fwd to your proposal.

kkozmic:
We've had this discussion some time ago. It breaks the abstraction

On Thu, 3 Mar 2016, 10:34 hamilton verissimo <notifications@github.com>
wrote:

> I dont think calling proceed in the continuation is a good idea anyways,
> you wont get a Task instance until you run the actual method.
>
> Looking fwd to your proposal.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/castleproject/Core/issues/145#issuecomment-191493625>.
>
-- 

sent from my phone
Krzysztof

joelweiss:
@hammett You are correct, but you can make it work if you return `TaksCompletionSource<>.Task` and when you get the final task you can await it and then call `taskCompletionSource.SetResult(task.Result)`.

My scenario is I have an `IFooService` and when I call `fooService.GetData()` the first interceptor is a CacheInterceptor that tries to fetch the data from the cache asynchronasly and if it's not in the cache, I want to call `invocation.Proceed()` to the HttpInterseptor and make an asynchronous http call to get the data.

I was thinking something like this might enable my scenario.
```csarp
	interceptors[currentInterceptorIndex].Intercept(this);
#if DOTNET40
	if (ReturnValue != null)
	{
		var returnType = Method.ReturnType;
[...]

PR #337, opened 2018-02-16
Example: Added tests to show a race condition when intercepting an async method

JSkimming:
This is not a functional PR, just an attempt to demonstrate a race condition in [AbstractInvocation](https://github.com/castleproject/Core/blob/v4.2.1/src/Castle.Core/DynamicProxy/AbstractInvocation.cs#L23-L25 "class AbstractInvocation").

This demonstrates the issue I outlined in #145, specifically [this comment](https://github.com/castleproject/Core/issues/145#issuecomment-365404690).

### DIFF
--- a/src/Castle.Core.Tests/Interceptors/AsyncInterceptor.cs
+++ b/src/Castle.Core.Tests/Interceptors/AsyncInterceptor.cs
@@ -1,0 +1,41 @@
+﻿// Copyright 2004-2010 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/Castle.Core.Tests/Interceptors/AsyncInterceptor.cs
```diff
@@ -0,0 +1,41 @@
...
+	{
+		public void Intercept(IInvocation invocation)
+		{
+			invocation.ReturnValue = InterceptAsyncMethod(invocation);
+		}
+
+		private static async Task InterceptAsyncMethod(IInvocation invocation)
+		{
+			// It all falls down when executing async before calling Proceed().
+			await Task.Delay(10).ConfigureAwait(false);
```
fir3pho3nixx: What happens if you change the line above to:

```csharp
Task.Delay(10).GetAwaiter().GetResult();
```

fir3pho3nixx: Managed to get test passing here: https://github.com/fir3pho3nixx/Core/pull/28. Maybe @StephenCleary can validate my approach as I borrowed some code from Nito.AsyncEx to make this work.

jonorossi:
@JSkimming thanks again for the unit tests. I'm going to close this pull request to hide it from the list, it is linked to from #145 so won't get lost. We haven't forgotten about it, however as you already know it isn't a quick fix.

Issue #339, opened 2018-03-12
Castle.Core fails to mock interfaces passing value types by reference


TAGC:
(Cross-posting some details from nsubstitute/NSubstitute#378)

C# 7.2 introduced [the possibility of using reference semantics for value types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#reference-semantics-with-value-types) - this is done by applying the `in` modifier for value-type parameters in method signatures.

Castle.Core does not seem to support generating mocks for interfaces that use `in` modifiers for value types. This has been reproduced using Castle.Core indirectly through the latest stable versions of Moq and NSubstitute:

```cs
namespace SomeProject.Tests
{
    public readonly struct Struct
    {
    }

    public interface IStructByRefConsumer
    {
        void Consume(in Struct message);
    }

    public interface IStructByValueConsumer
    {
        void Consume(Struct message);
    }

    public class TempSpec
    {
        // Fails.
        [Fact]
        internal void Struct_ByRef_Moq_Test()
        {
            _ = Mock.Of<IStructByRefConsumer>();
        }

        // Fails.
        [Fact]
        internal void Struct_ByRef_NSubstitute_Test()
        {
            _ = Substitute.For<IStructByRefConsumer>();
        }

        // Passes.
        [Fact]
        internal void Struct_ByValue_Moq_Test()
        {
            _ = Mock.Of<IStructByValueConsumer>();
        }

        // Passes.
        [Fact]
        internal void Struct_ByValue_NSubstitute_Test()
        {
            _ = Substitute.For<IStructByValueConsumer>();
        }
    }
}
```

Posted below is the part of the stracktrace that's common between the failing Moq and NSubstitute tests:

```
Result StackTrace:	
at System.Reflection.Emit.TypeBuilder.TermCreateClass(RuntimeModule module, Int32 tk, ObjectHandleOnStack type)
   at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
   at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateType(TypeBuilder type)
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.BuildType()
   at Castle.DynamicProxy.Generators.InterfaceProxyWithoutTargetGenerator.GenerateType(String typeName, Type proxyTargetType, Type[] interfaces, INamingScope namingScope)
[...]

PR #344, opened 2018-04-20
Add support for `in` parameter modifier on .NET Core by adding .NET Standard 1.5 as additional target

stakx:
This is an attempt at solving #339.

It would be good if someone who's intimately familiar with the build process could take a closer look at what I've done in order to add the additional `netstandard1.5` target:

* Have I done all that's required when adding a new TFM, or did I miss something, somewhere?
* Should the 3 logging integration projects also get the new `netstandard1.5` target to stay in-sync with Castle.Core?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Enhancements:
 - Added .NET Standard/.NET Core support for NLog (@snakefoot, #200)
 - Added .NET Standard/.NET Core support for log4net (@snakefoot, #201)
+- Add .NET Standard 1.5 as additional target to NuGet package. This adds support for `in` parameter modifiers on e.g. .NET Core (@stakx, #339)
 
 Bugfixes:
 - Fix Castle.Services.Logging.Log4netIntegration assembly file name casing which breaks on Linux (@beginor, #324)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 Enhancements:
 - Added .NET Standard/.NET Core support for NLog (@snakefoot, #200)
 - Added .NET Standard/.NET Core support for log4net (@snakefoot, #201)
-- Add .NET Standard 1.5 as additional target to NuGet package. This adds support for `in` parameter modifiers on e.g. .NET Core (@stakx, #339)
+- DynamicProxy supported C# `in` parameter modifiers only on the .NET Framework up until now. Adding .NET Standard 1.5 as an additional target to the NuGet package makes them work on .NET Core, too (@stakx, #339)
 
 Bugfixes:
 - Fix Castle.Services.Logging.Log4netIntegration assembly file name casing which breaks on Linux (@beginor, #324)
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
@@ -22,7 +22,7 @@ namespace Castle.DynamicProxy.Tests
 	// that was introduced with C# language version 7.2.
 	[TestFixture]
 	public class ValueTypeReferenceSemanticsTestCase : BasePEVerifyTestCase
-    {
+	{
 		// This test isn't interesting by itself. It only establishes a reference "baseline" for the next test.
 		[Test]
 		public void Can_proxy_method_having_valuetyped_parameter_without_in_modifier()
@@ -86,5 +86,5 @@ namespace Castle.DynamicProxy.Tests
 		{
 			void Method(in ReadOnlyStruct readOnlyStruct);
 		}
-    }
+	}
 }
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
@@ -1,0 +1,90 @@
+﻿// Copyright 2004-2018 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.f
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests
+{
+	using Castle.DynamicProxy.Tests.Interceptors;
+
+	using NUnit.Framework;
+
+	// The purpose of this test fixture is to ensure that DynamicProxy can handle the `in` parameter modifier
+	// that was introduced with C# language version 7.2.
+	[TestFixture]
+	public class ValueTypeReferenceSemanticsTestCase : BasePEVerifyTestCase
+    {
+		// This test isn't interesting by itself. It only establishes a reference "baseline" for the next test.
+		[Test]
+		public void Can_proxy_method_having_valuetyped_parameter_without_in_modifier()
+		{
+			var proxy = this.generator.CreateInterfaceProxyWithoutTarget<IWithoutInModifier>(new DoNothingInterceptor());
+			var readOnlyStruct = new ReadOnlyStruct();
+			proxy.Method(readOnlyStruct);
+		}
+
+#if FEATURE_EMIT_CUSTOMMODIFIERS
+
+		// ^^^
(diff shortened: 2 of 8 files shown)


stakx:
Regarding the Travis CI build failures:

> ```
> CSC : error CS1617: Invalid option '7.2' for /langversion; must be ISO-1, ISO-2, Default
> or an integer in range 1 to 6.
> [/home/travis/build/castleproject/Core/src/Castle.Core.Tests/Castle.Core.Tests.csproj]
> ```

Since this seems to be Microsoft's compiler (`csc`), I wonder why it doesn't know language versions >6. Is this because of the .NET 4.6 targeting pack?

Also, it seems that Mono's C# compiler (`mcs`) does not yet know C# 7.2 features. Both the `<LangVersion>` in `Castle.Core.Tests.csproj` as well as the unit tests added in this PR should be hidden from Mono. I'm not exactly sure how to do this properly.

> ```
> /usr/share/dotnet/sdk/1.0.4/NuGet.targets(97,5): error : Unable to resolve 
> 'Microsoft.DiaSymReader.Native (>= 1.4.1)' for '.NETCoreApp,Version=v1.1'.
> [/home/travis/build/castleproject/Core/buildscripts/BuildScripts.csproj]

This seems to be a general issue with the build process, and therefore unrelated to this PR.

jonorossi:
> Have I done all that's required when adding a new TFM, or did I miss something, somewhere?

Looks great, pretty sure that's it.

> Should the 3 logging integration projects also get the new netstandard1.5 target to stay in-sync with Castle.Core?

.NET Standard versions of dependencies:
* log4net - 1.3
* NLog - 1.3, 1.5, 2.0
* Serilog - 1.0, 1.3

Do you know what happens if a user picks NLog and is on a platform where NuGet could choose .NET Standard 1.5/2.0 but we only have a 1.3, .NET Standard is obviously additive, but if NuGet picked a higher platform it would add members/types etc that could break us depending on it. This will probably determine if we should add it to these too.

> Also, it seems that Mono's C# compiler (mcs) does not yet know C# 7.2 features. Both the <LangVersion> in Castle.Core.Tests.csproj as well as the unit tests added in this PR should be hidden from Mono. I'm not exactly sure how to do this properly.
>
> This seems to be a general issue with the build process, and therefore unrelated to this PR.

I disagree that it is unrelated to this pull request, this pull request merged would break the build.

Yes, Mono now ships Roslyn as its default compiler (`csc`) even though it still ships `mcs`. Since we are still using Mono 4.6.1 (30 Sep 2016) I think we should upgrade it first, it is likely that a new version of the compiler has been shipped with a newer Mono version.

src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+	using Castle.DynamicProxy.Tests.Interceptors;
+
+	using NUnit.Framework;
+
+	// The purpose of this test fixture is to ensure that DynamicProxy can handle the `in` parameter modifier
+	// that was introduced with C# language version 7.2.
+	[TestFixture]
+	public class ValueTypeReferenceSemanticsTestCase : BasePEVerifyTestCase
+    {
```
jonorossi: Indentation mix up here.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 Enhancements:
 - Added .NET Standard/.NET Core support for NLog (@snakefoot, #200)
 - Added .NET Standard/.NET Core support for log4net (@snakefoot, #201)
+- Add .NET Standard 1.5 as additional target to NuGet package. This adds support for `in` parameter modifiers on e.g. .NET Core (@stakx, #339)
```
jonorossi: Could to clarify the changelog entry so it is clear this wasn't a problem on .NET Framework. Thanks.

stakx: I rewrote the changelog entry as follows:

> DynamicProxy supported C# `in` parameter modifiers only on the .NET Framework up until now. Adding .NET Standard 1.5 as an additional target to the NuGet package makes them work on .NET Core, too (@stakx, #339)

Is that OK?

src/Castle.Core.Tests/DynamicProxy.Tests/ValueTypeReferenceSemanticsTestCase.cs
stakx: Fixed. Sorry for that.

stakx:
> ```
> CSC : error CS1617: Invalid option '7.2' for /langversion; must be ISO-1, ISO-2, Default
> or an integer in range 1 to 6.
> [/home/travis/build/castleproject/Core/src/Castle.Core.Tests/Castle.Core.Tests.csproj]
> ```

I found through trial and error that this is caused by the use of an outdated version of the .NET Core SDK (1.0.4). We need to update the `dotnet` entry in `.travis.yml`.

* .NET Core SDK `2.0.3` only supports C# up to language version 7.1. Too bad we need 7.2.
* .NET Core SDK `2.1.105` does support C# 7.2, but no longer installs .NET Core Runtime 1.1 (which the Castle Core test projects target).

I'd prefer to deal with the .NET Core SDK upgrade in a separate PR (in order to keep this one unsoiled), then rebase this later on.

jonorossi:
> Executing .NET Framework 4.6.1 tests suddenly doesn't work anymore (Permission denied):
> 
> I don't know where that problem came from, nor how to resolve it... would simply preceding these commands with mono possibly help?

Possibly if binfmt isn't set up anymore after changing the Travis settings. Sounds like a good change anyway.

CHANGELOG.md
jonorossi: Yep

stakx:
Joy! Prepending `mono ` to the commands that invoke an `.exe` indeed worked. 🎆

stakx:
There's still this open question from you:

> .NET Standard versions of dependencies:
>
> * log4net - 1.3
> * NLog - 1.3, 1.5, 2.0
> * Serilog - 1.0, 1.3
>
> Do you know what happens if a user picks NLog and is on a platform where NuGet could choose .NET Standard 1.5/2.0 but we only have a 1.3, .NET Standard is obviously additive, but if NuGet picked a higher platform it would add members/types etc that could break us depending on it. This will probably determine if we should add it to these too.

I don't know as of now. I'll probably need to run a few experiments later on.

stakx:
>> Do you know what happens if a user picks NLog and is on a platform where NuGet could choose .NET Standard 1.5/2.0 but we only have a 1.3, .NET Standard is obviously additive, but if NuGet picked a higher platform it would add members/types etc that could break us depending on it. This will probably determine if we should add it to these too.
>
> I don't know as of now. I'll probably need to run a few experiments later on.

**TL;DR:** While we might be on the safer side by adding `netstandard1.5` to the logging packages, I suspect that it wouldn't make any difference in practice.

I ran two quick experiments:

----

#### (1)

* Created a `BaseLibrary` NuGet that targets both .NET Standard 1.3 and .NET Standard 1.5. (This is the equivalent to `Castle.Core`.)
* Created a `DependentLibrary` NuGet that depends on `BaseLibrary`, but targets only .NET Standard 1.3. (This is the equivalent of e.g. `Castle.Core-NLog`.)
* A console app `Consumer` that depends on `DependentLibrary` and runs on .NET Framework 4.6.2 (which implements .NET Standard 1.5).

In this case, this is what gets loaded when running `Consumer`:
* `BaseLibrary`, .NET Standard 1.5 version. (!)
* `DependentLibrary`, .NET Standard 1.3 version (as there is no other).

----

#### (2)

Same setup as (1), but `DependentLibrary` (aka `Castle.Core-NLog`) targets both .NET Standard 1.3 and .NET Standard 1.5.

In this case, this is what gets loaded when running `Consumer`:
* `BaseLibrary`, .NET Standard 1.5 version.
* `DependentLibrary`, .NET Standard 1.5 version.

----

In both cases, the version of `BaseLibrary` (aka `Castle.Core`) that targets .NET Standard 1.5 got chosen, supposedly because that's what `Consumer`'s platform (.NET Framework 4.6.2) supports.

This would suggest that there's no point in adding a `netstandard1.5` TFM to the logging packages, because the presence or absence of that TFM won't influence which `Castle.Core` target will get picked.

Disclaimer: Obviously, the .NET tooling being what it currently is, YMMV... I simply don't have a thorough-enough understanding of the target/version resolution algorithm of `msbuild` / `dotnet` / `nuget`. I'd be happy if someone could validate my above reasoning. /cc @fir3pho3nixx

jonorossi:
> In both cases, the version of BaseLibrary (aka Castle.Core) that targets .NET Standard 1.5 got chosen, supposedly because that's what Consumer's platform (.NET Framework 4.6.2) supports.
> 
> This would suggest that there's no point in adding a netstandard1.5 TFM to the logging packages, because the presence or absence of that TFM won't influence which Castle.Core target will get picked.

That is what I thought would happen, it is expected that newer .NET Standard versions won't make breaking changes and the IL generated when compiling against lower .NET Standard versions will always be compatible even if new members appear in a higher .NET Standard compatible assembly.

jonorossi:
Squished and merged the changes. Thanks @stakx.

Issue #327, opened 2017-12-07
ModuleScope.SaveAssembly produces incorrect / incomplete assemblies when proxying types from a weak-named assembly


stakx:
**TL;DR:** I am either misunderstanding how `ModuleScope.SaveAssembly` works, or DynamicProxy &mdash; when it produces proxies for types in weak-named assemblies &mdash; puts generated proxy types in a weak-named module and proxy method invocation types in a strong-named module, resulting in failure or incomplete output by `ModuleScope.SaveAssembly`.

The generated assemblies mentioned below can be found in [`CastleDynProxy2_strong_and_weak.zip`](https://github.com/castleproject/Core/files/1537996/CastleDynProxy2_strong_and_weak.zip).

### Saving with `ModuleScope.SaveAssembly(strongNamed: false)`

Suppose I'm trying to save the dynamic assembly generated by DynamicProxy as follows:

```csharp
using Castle.DynamicProxy;

public interface IFoo
{
    void Method();
}

public abstract class Bar
{
    public abstract void Method();
}

class Program
{
    public static void Main()
    {
        var proxyGenerator = new ProxyGenerator(new PersistentProxyBuilder());
        var foo = proxyGenerator.CreateInterfaceProxyWithoutTarget<IFoo>();
        var bar = proxyGenerator.CreateClassProxy<Bar>();
        proxyGenerator.ProxyBuilder.ModuleScope.SaveAssembly(strongNamed: false);
    }
}
```

Note that I am *not* strong-naming the assembly containing above code&mdash;if I did, then everything would work as expected.

This generates an assembly `CastleDynProxy2`, which appears to have two problems:

1. It is missing types `Castle.Proxies.Invocations.IFoo_Method` and `Castle.Proxies.Invocations.Bar_Method`. PEVerify is a bit cryptic about it, but seems to tell me the same thing:

   ```
   [IL]: Error: [C:\...\CastleDynProxy2.dll : Castle.Proxies.IFooProxy::Method]
   [HRESULT 0x80070002] - The system cannot find the file specified.
   
   [IL]: Error: [C:\...\CastleDynProxy2.dll : Castle.Proxies.BarProxy::Method
   [HRESULT 0x80070002] - The system cannot find the file specified.
   
   2 Error(s) Verifying CastleDynProxy2.dll
   ```

2. There is an assembly reference to a strong-named `DynamicProxyGenAssembly2`, which is confusing, because the generated assembly already has the exact same name (except for being weak-named).

[...]

PR #346, opened 2018-04-22
Emit invocation types in same module as their associated proxy type

stakx:
My attempt at fixing #327, which is about DynamicProxy generating two invalid dynamic assemblies when proxying types from non-strong-named assemblies (because the proxy types end up in the weak-named assembly while the associated invocation types end up in the strong-named assembly).

This PR does two things:

1. Add an additional unit test project whose assembly does not get strong-named. Otherwise, we have no way of testing this issue (since the main unit test project gets signed).

2. Make sure that the dynamically generated invocation types end up in the same assembly as the generated proxy type.

<strike>Not to be merged just yet. I'd like to check some more that applying `forceUnsigned` when generating invocation types won't have any undesired side effects.</strike>


src/Castle.Core/DynamicProxy/Generators/Emitters/ClassEmitter.cs
```diff
@@ -66,6 +66,11 @@ public ModuleScope ModuleScope
 			get { return moduleScope; }
 		}
 
+		internal bool InStrongNamedModule
+		{
+			get { return StrongNameUtil.IsAssemblySigned(TypeBuilder.Assembly); }
+		}
```
stakx: This property gets checked for every invocation type being generated, so this *might* have a slight performance impact. I guess the result returned by `StrongNameUtil.IsAssemblySigned` could be cached here, in `ClassEmitter`. Seeing that `StrongNameUtil` already performs caching, all we'd really save at this point is the time it takes to enter and leave the `lock` inside `StrongNameUtil`. And caching here might actually introduce a `lock` of its own, so just going ahead and performing this call instead of caching the result is probably good enough.

build.sh
```diff
@@ -34,34 +34,38 @@ dotnet restore ./src/Castle.Services.Logging.log4netIntegration/Castle.Services.
...
 echo ---------------------------
 echo Running NETCOREAPP1.1 Tests
 echo ---------------------------
 
 dotnet ./src/Castle.Core.Tests/bin/Release/netcoreapp1.1/Castle.Core.Tests.dll --result=NetCoreClrTestResults.xml;format=nunit3
+dotnet ./src/Castle.Core.Tests.NonSigned/bin/Release/netcoreapp1.1/Castle.Core.Tests.NonSigned.dll --result=NetCoreClrNonSignedTestResults.xml;format=nunit3
 
 # Unit test failure
-NETCORE_FAILCOUNT=$(grep -F "One or more child tests had errors" NetCoreClrTestResults.xml | wc -l)
+NETCORE_FAILCOUNT=$(grep -F "One or more child tests had errors" NetCoreClrTestResults.xml NetCoreClrNonSignedTestResults.xml | wc -l)
```
stakx: Can you confirm that this `grep` call syntax (mentioning several space-separated file names) is correct and will work correctly in the case of a failed unit test?

build.sh
```diff
@@ -34,34 +34,38 @@ dotnet restore ./src/Castle.Services.Logging.log4netIntegration/Castle.Services.
...
 if [ $NETCORE_FAILCOUNT -ne 0 ]
 then
     echo "NetCore Tests have failed, failing the build"
     exit 1
 fi
 
-MONO_FAILCOUNT=$(grep -F "One or more child tests had errors" DesktopClrTestResults.xml | wc -l)
+MONO_FAILCOUNT=$(grep -F "One or more child tests had errors" DesktopClrTestResults.xml DesktopClrNonSignedTestResults.xml | wc -l)
```
stakx: (as above)

src/Castle.Core.Tests.NonSigned/Castle.Core.Tests.NonSigned.csproj
```diff
@@ -0,0 +1,66 @@
...
+	</PropertyGroup>
+
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<Authors>Castle Contributors</Authors>
+		<AssemblyName>Castle.Core.Tests.NonSigned</AssemblyName>
+		<RootNamespace>Castle</RootNamespace>
+		<Version>0.0.0.0</Version>
+		<AssemblyVersion>0.0.0.0</AssemblyVersion>
+		<SignAssembly>False</SignAssembly>
```
jonorossi: It is going to really suck to have to maintain a nearly exact copy of the unit test project. Do you think we could get build the same project twice, with the second build passing a parameter to not strong name?

Castle.Core.sln
```diff
@@ -31,6 +31,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Services.Logging.Ser
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Castle Services", "Castle Services", "{A598EE9B-41CE-4BE8-BF93-2C91F919F97E}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Castle.Core.Tests.NonSigned", "src\Castle.Core.Tests.NonSigned\Castle.Core.Tests.NonSigned.csproj", "{14D86762-CF9B-4560-80C9-10C16DBE246C}"
```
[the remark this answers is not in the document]
stakx: Sure. I've actually been a little undecided about the correct term to use. I'll change the name to `WeakNamed` (mainly because it's shorter than `NonStrongNamed`).

src/Castle.Core.Tests.NonSigned/Castle.Core.Tests.NonSigned.csproj
stakx: I'll look into it.

Castle.Core.sln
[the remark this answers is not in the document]
jonorossi: Great, my preference would be "strong" and "weak" named too.

stakx:
> Do you think we could get build the same project twice, with the second build passing a parameter to not strong name?

I'm not sure this is going to work out so well.

* Castle.Core's internals are made visible to Castle.Core.Tests. Since the former is strong-named, the latter needs to be strong-named as well. If we want a build property that determines strong-naming vs. not strong-naming, it will have to be applied to both projects, not just to the test project.

* Alternatively, we can surround all tests that access Castle.Core's internals with a `#if STRONG_NAMED` guard. Sprinkling *more* conditional compilation blocks across the test project isn't exactly neat, though.

What's your preference here? Go ahead with any of the above options, or keep a separate weak-named test project?

* If we keep a separate weak-named test project, we could probably factor out some of the identical bits of both test projects into a MSBuild `.props` file.

stakx:
(I forgot a fourth alternative: leaving this change untested, and therefore not having a weak-named test project at all.)

jonorossi:
> What's your preference here? Go ahead with any of the above options, or keep a separate weak-named test project?

Oops, I thought you were running all unit tests in both test assemblies, but the new weak named one just runs those new ones you added. Let's keep it with 2 projects, the weak named one will likely only need to be updated when new tests are added to it which won't be often.

Issue #227, opened 2017-01-23
Change usages of __MonoCS__ to the NUnit PlatformAttribute


jonorossi:
We currently use the `__MonoCS__` compiler define in unit tests to alter the behaviour of unit tests when run on the Mono runtime, however this define was only meant to indicate the Mono compiler was being used.

With Mono moving to the Roslyn compiler this define will no longer be defined:
https://lists.dot.net/pipermail/mono-devel-list/2017-January/044104.html

We should probably move the remaining uses of this define to that `[Platform]` attribute.

fir3pho3nixx:
After https://github.com/castleproject/Core/issues/241 I would be happy to check this out.

fir3pho3nixx:
@jonorossi - is this still relevant?

jonorossi:
We have worked around this as we've changed to the new .NET tooling to use Mono on Linux by defining this ourselves in common.props, however we should still migrate to using `PlatformAttribute` like some of our unit tests already do, however I didn't want to just do it as a bulk find/replace because you really do want to put a reason in the attribute's constructor so you know why it can't work on Mono (some might already be fixed on a newer version of Mono).

```
<DesktopClrConstants Condition="'$(OS)'=='Unix'">$(CommonDesktopClrConstants);__MonoCS__</DesktopClrConstants>
```
(https://github.com/castleproject/Core/blob/v4.1.0/buildscripts/common.props#L42)

stakx:
> We should probably move the remaining uses of [the `__MonoCS__` conditional compilation symbol] to that `[Platform]` attribute.

@jonorossi - I've just noticed that this is probably not going to work out: Tests marked with the `[Platform]` attribute won't compile on .NET Core because that attribute is not available in NUnit's .NET Standard version (which that's the one that gets used when compiling for .NET Core).

The reason why this wasn't a problem so far was because all tests carrying a `[Platform]` attribute were already skipped by conditional compilation symbols that aren't defined for .NET Core (`FEATURE_ASSEMBLYBUILDER`, `FEATURE_CONFIGURATION`, `FEATURE_SECURITY_PERMISSIONS`).

jonorossi:
@stakx great work. Your pull request will definitely help when we add a .NET Standard 2.0 build in the future and start adding some features back in.

PR #347, opened 2018-04-23
Configure Travis CI to use Mono version 5.10.1

stakx:
This is in response to https://github.com/castleproject/Core/pull/344#issuecomment-383375492:

> Mono now ships Roslyn as its default compiler (csc) even though it still ships mcs. Since we are still using Mono 4.6.1 (30 Sep 2016) I think we should upgrade it first, it is likely that a new version of the compiler has been shipped with a newer Mono version.

5.10.1 is the latest stable release of Mono, so let's update to this one.


NuGet.config
```diff
@@ -0,0 +1,11 @@
...
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+
+    <!-- This package source is required for package Microsoft.TargetingPack.NETFramework.v4.6.1,
+         which is not available on NuGet.org. This package contains the reference assemblies
+         required to compile for .NET Framework on Unix. -->
+    <add key="dotnet.myget.org" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
+  </packageSources>
+</configuration>
```
stakx: I'm not certain that this is the best we can do. Configuring an additional NuGet package feed at the repo level results in some packages being restored from the MyGet feed, even if they were available on nuget.org.

It would probably be better to configure this additional package source in a way such that it gets used only when restoring `BuildScripts.csproj`. I'm not sure how to do this. Moving this `NuGet.config` into `buildscripts/` does not appear to work.

build.sh
```diff
@@ -28,7 +28,7 @@ fi
 
 # This installs the latest build of the dotnet targetting pack net461 via NuGet for linux desktop clr builds.
 # The package source can be removed once Microsoft.TargetingPack.NETFramework.v4.6.1 comes out of alpha on NuGet.
-dotnet restore ./buildscripts/BuildScripts.csproj -s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
+dotnet restore ./buildscripts/BuildScripts.csproj
```
stakx: (Specifying the package source here is no longer required, since the preceding `nuget restore Castle.Core.sln` (run by Travis CI) already downloaded the package from the alternate NuGet feed.)

jonorossi:
@fir3pho3nixx you added the other feed, have you been tracking the .NET Framework targeting packs threads to know the current status? It appears from the following 2 threads that there still isn't a supported mechanism.

* https://github.com/dotnet/sdk/issues/335
* https://github.com/dotnet/designs/pull/33

stakx:
I got everything working again the way it used to, but I'm seeing an error message in the [Travis build log](https://travis-ci.org/castleproject/Core/builds/370037254?utm_source=github_status&utm_medium=notification):

> ```
> /usr/share/dotnet/sdk/1.0.4/NuGet.targets(97,5): error : Unable to resolve
> 'Microsoft.DiaSymReader.Native (>= 1.4.1)' for '.NETCoreApp,Version=v1.1'.
> [/home/travis/build/castleproject/Core/buildscripts/BuildScripts.csproj]
> ```

Not quite sure where that is coming from or whether it needs to be fixed at all... it doesn't appear to affect the build as a whole.

.travis.yml
```diff
@@ -2,7 +2,10 @@ language: csharp
 
 solution: Castle.Core.sln
 
-mono: 4.6.1
+install:
+    - # Automatic package restore disabled. Packages will be restored by the build script.
```
stakx: This is kind of hackish. A simple one-liner `install:` will be ignored by Travis, so we need either an empty line, or perhaps some kind of `echo` informational message.

jonorossi:
Yer I've noticed that `Microsoft.DiaSymReader.Native` error recently too, don't worry about it for now.

> Just as a btw., there's a Stack Overflow answer that suggests that starting with Mono 5.x there's now another solution that the one we're currently using. (I'm happy to give it a try.)

Which option were you considering? 1 or 2? I'd love if we could rid ourselves of that unsupported dotnet.myget.org feed. Please do give it a try.

stakx:
> Which option were you considering? 1 or 2?

I'm currently looking at option 1, which I find neater because it doesn't require any hard-coded path in an environment variable.

The good news is that we can indeed get rid of `<FrameworkPathOverride>` (and the whole `BuildScripts.csproj` project along with it). All we need to do is turn invocations of `dotnet` into `msbuild`.

The bad news is that this appears to break NUnit's `[Platform(Exclude = "Mono", ...)]`. Those tests will now execute when they shouldn't. One solution would be to revert the `[Platform]` attributes back to `#if __MonoCS__` (which according to #227 is exactly the wrong thing to do).

I'll investigate further later on.

jonorossi:
> The bad news is that this appears to break NUnit's [Platform(Exclude = "Mono", ...)]. Those tests will now execute when they shouldn't. One solution would be to revert the [Platform] attributes back to #if __MonoCS__ (which according to #227 is exactly the wrong thing to do).

If you can't work out a way to reinstate the Platform attribute working correctly, then I'm way more interested to get rid of this cruft than the Platform attribute. The reason I was pushing towards the Platform attribute is it gives a good reason in the output for why it was ignored and pushes us towards not having to compile unit tests twice, but that is likely never going to ever be clean on Mono.

stakx:
> I'd love if we could rid ourselves of that unsupported dotnet.myget.org feed.

I think I've got it: While I didn't succeed with option 1 (using `msbuild` instead of `dotnet`), option 2 (`export FrameworkPathOverride=$(dirname $MONOPATH)/../lib/mono/4.6.1-api/` in `build.sh`) appears to work. I've picked that relative path based on information [over at TheAngryByrd/dotnet-mono](https://github.com/TheAngryByrd/dotnet-mono#a-word-on-frameworkpathoverride).)

> If you can't work out a way to reinstate the Platform attribute working correctly [...].

Please disregard the warning I gave about `[Platform]`. As it turned out, this was due to my local setup. (I tested the Unix build with the Ubuntu on Windows 10 subsystem instead of with a real Linux box. That'll teach me.)

fir3pho3nixx:
@stakx this is the right way to go. This has been hanging around in the back of my mind for a while. I have been busy with the ASP.NET stuff I think the Castle.Core needs a build refresh. I am so sorry you had to deal with it here.

fir3pho3nixx:
I will move back up and clean the build once we have done V5 on Windsor. Sorry again.

stakx:
@fir3pho3nixx: No worries, mate. I admit that I hadn't quite expected how far that first one-line commit would take me, but I've learned a lot. It's all good. 😁

jonorossi:
> I admit that I hadn't quite expected how far that first one-line commit would take me, but I've learned a lot.

@stakx me either, I half forgot all the crap that was done to work around the half broken .NET CLI tooling.

PR #348, opened 2018-04-24
Rectify test project's directory and namespace structure

stakx:
(This is fairly low priority:)

In the past, when adding tests to Castle.Core, I've often found that it's hard to decide where files should go and in which namespaces to put new test types. Today I've reserved some time to clean up the test project.

* Bring namespaces and directory structure in line with one another (or at least more than was previously the case).
* Bring namespaces in line with those of the main project. Tests for stuff in `X` always go in a `X.Tests` sub-namespace.
* Minimize the amount of duplicate namespaces.

Regarding the namespaces, here's a visual overview of what this PR does. In the middle, you see the namespaces in Castle.Core. The namespaces in Castle.Core.Tests are on the left (before PR) and on the right (after PR). The lines indicate the namespace mapping:

namespace-matching-before-after


jonorossi:
> I've often found that it's hard to decide where files should go and in which namespaces to put new test types.

LOL, I always wondered how we ended up with such a mess when projects got merged, and have had this on my Castle Trello board for years but just never got there. I'll take a look tomorrow to confirm I'm happy with the moves.

jonorossi:
> If we want VS to automatically generate the right namespaces for new code files, we'd have to rename directories: Core.Tests\Internal → Core\Internal.Tests and so on.

Let's look at this after we get this merged. If you fix the conflict I'm happy with the changes. Some of the namespaces are a little strange because it doesn't follow the usual convention for unit tests, but since Castle.Core.dll has things in so many namespaces this makes the most sense.

> The various namespaces for test types (.Classes, .InterClasses, .GenClasses, .Interfaces, .GenInterfaces) are to some extent arbitrary, I'm not sure they really add any real value over a single namespace (e.g. .Types).

I'd have to take a closer look but you are likely right. I also suspect now with things a little more organised that we'd probably be able to merge a bunch of test types.

stakx:
> If you fix the conflict [...]

Done. Rebased on top of the latest merges, should be all good to go.

PR #351, opened 2018-04-30
Remove __MonoCS__ conditional compilation symbol

### LINKED ISSUE #227: Change usages of __MonoCS__ to the NUnit PlatformAttribute

stakx:
This resolves #227 and ticks off (I believe) several ignored tests in #94.

`__MonoCS__` is gotten rid of through a combination of three approaches:

1. Most tests that have been disabled for Mono can simply be re-enabled, because it turns out that they actually pass just fine. `__MonoCS__` can be removed without replacement.

2. Some usages of `__MonoCS__` are due to PEVerify not being available on Unix systems. These instances are replaced with a new, dedicated conditional compilation symbol, `FEATURE_PEVERIFY`.

3. Some few remaining uses of `__MonoCS__` involve tests that really do fail on Mono. In these cases, it is replaced with a `[ExcludeOnMono]` custom attribute that includes the reason why the test fails, often linking to a GitHub issue over at [mono/mono](https://github.com/mono/mono). (Why `[ExcludeOnMono]` and not `[Platform]`? See https://github.com/castleproject/Core/issues/227#issuecomment-385208608.)

This PR also replaces all existing occurrences of `[Platform(Exclude = "Mono", ...)]` with `[ExcludeOnMono(...)]`.


src/Castle.Core.Tests/BasicClassProxyTestCase.cs
```diff
@@ -78,18 +78,15 @@ public void Caching()
...
 
 		[Test]
-#if __MonoCS__
-		[Ignore("Expected: Castle.DynamicProxy.Generators.GeneratorException, But was: System.ArgumentNullException")]
-#endif
 		public void ProxyForNonPublicClass()
 		{
 			// We need to use a type that is not from our assembly, because we are marked as internals visible to DynamicProxy2
-			var type = Type.GetType("System.__Canon"); // Don't specify the assembly name (it'll be either mscorlib or System.Private.CorLib)
+			var type = Type.GetType("NUnit.Framework.Internal.ActionsHelper, nunit.framework");
```
jonorossi: Was this necessary, I'm surprised Mono wouldn't also have a matching `System.__Canon` class. The reason for my query is that `__Canon` will be very stable as an internal class where as that NUnit class might change/move in the future.

https://github.com/mono/mono/blob/73866f6b16bb09a34b3caa0271bb9fa6ae9d8331/mcs/class/referencesource/mscorlib/system/object.cs#L230-L237

README.md
```diff
@@ -91,6 +92,7 @@ Symbol                              | NET35              | NET40              |
 * `FEATURE_LEGACY_REFLECTION_API` - provides a shim for .NET 3.5/4.0 that emulates the `TypeInfo` API available in .NET 4.5+ and .NET Core.
 * `FEATURE_LISTSORT` - enables support for features that make use of System.ComponentModel.ListSortDescription.
 * `FEATURE_NETCORE_REFLECTION_API` - provides shims to implement missing functionality in .NET Core that has no alternatives.
+* `FEATURE_PEVERIFY` - enables verification of dynamic assemblies using PEVerify during tests. (Only defined on Windows builds since Windows is currently the only platform where PEVerify is available.)
```
jonorossi: Since this is a unit test only symbol could we name it `FEATURE_TEST_PEVERIFY` to make that fact clear, which would group it with the other 2 below.

Also each column should have a tick mark rather than a cross.

src/Castle.Core.Tests/BasicClassProxyTestCase.cs
stakx: The Mono repo contains only very few [mentions of `__Canon`](https://github.com/mono/mono/search?q=__Canon). The file you linked to is one, but it appears to be relatively new. Also, I'm not sure why, but `System.__Canon` simply does not end up in Mono's `mscorlib.dll`, and I suppose that's why a call to `Type.GetType("System.__Canon")` returns `null` on Mono.

One notable internal type that both runtimes have is `System.__ComObject`. However, if you try to proxy that even on the CLR, you'll get a `FatalExecutionEngineError`. 😆

Not sure what to do about this. Should the test just be disabled on Mono?

stakx: Nice find! :+1: I've updated the test accordingly. Works fine!

jonorossi:
Looks like you've made all the review changes. I should have got this merged before the one moving unit tests around, could you fix this one up.

floydpink:
Good work @stakx - thank you! 😄

PR #353, opened 2018-05-02
Add tests documenting a CLR bug in System.Reflection.Emit

stakx:
As requested in https://github.com/castleproject/Core/issues/339#issuecomment-385906426.


jonorossi:
Thanks, merged.

Issue #341, opened 2018-04-05
Class proxies lose custom attributes on constructor parameters


gliljas:
The proxy generation tries hard to keep and clone attributes, but not on the constructor parameters. This causes problems for us, since our IoC container uses attributes to resolve certain values. When combining this with AOP proxying for e.g logging, we can no longer inspect the custom attributes on the resolved type, but instead have to use reflection to find the base class constructor.

Also, is there something that would prohibit keeping the parameter names? Convention based value injection could be nice, and that will also be tricky right now.

jonorossi:
@gliljas could you provide a code example or failing unit test of what you are doing/expecting. Are you attributes on the parameters set to inherit? Are you seeing constructor parameters with no names or just generic names?

gliljas:
See test in PR. I don't think inheritance comes into play.

jonorossi:
This should now be fixed by #355.

PR #355, opened 2018-05-02
Replicate custom attributes on constructor parameters

stakx:
This is an attempt at resolving #341.

The gist of this PR is to change code in the following location where custom attribute reproduction happens (currently only for `ParamArrayAttribute`):

https://github.com/castleproject/Core/blob/980752ca2c309ab86278b0100fa8f2e52de00efd/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs#L240-L249

This gets changed to replicate *all* custom attributes of all parameters.

There's a catch, however: Because constructors in the generated proxy type have different signatures (additional parameters e.g. for the interceptor array), the constructors' signatures do not match with those in the base class. This means that custom attribute inheritance would never work with these constructor parameters, which is why we need to disregard `[AttributeTargets(Inherited = true)]` and reproduce even those attributes.

Can you think of any scenario(s) where replicating additional, "inheritable" custom attributes on ctors could cause negative side effects in downstream libraries?

Finally, take note that AFAIK even the updated code **does not yet reproduce custom modifiers** (modopts, modreqs) because that would have to happen in a different code location and because that's not what #341 is about. Eventually, we should reproduce custom modifiers as well, otherwise we might get confronted with something like #339, but for ctors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Enhancements:
 - Added .NET Standard/.NET Core support for NLog (@snakefoot, #200)
 - Added .NET Standard/.NET Core support for log4net (@snakefoot, #201)
 - DynamicProxy supported C# `in` parameter modifiers only on the .NET Framework up until now. Adding .NET Standard 1.5 as an additional target to the NuGet package makes them work on .NET Core, too (@stakx, #339)
+- Replicate custom attributes on constructor parameters in the generated proxy type constructors to fulfill introspection of constructors. This does not change the proxying behavior. (@stakx, #341)
 - Improve performance of InvocationHelper cache lookups (@tangdf, #358)
 
 Bugfixes:
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ConstructorWithAttributesOnParametersTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ConstructorWithAttributesOnParametersTestCase.cs
@@ -1,0 +1,57 @@
+// Copyright 2004-2018 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests
+{
+	using System.Linq;
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Tests.Classes;
+
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class ConstructorWithAttributesOnParametersTestCase : BasePEVerifyTestCase
+	{
+		[Test]
+		public void CustomAttributesOnParametersAreCopiedToProxiedObjectConstructor()
+		{
+			var requiredObj = generator.CreateClassProxy(typeof(ClassWithAttributesOnConstructorParameters), new object[] { 10,"" });
+			var constructor = requiredObj.GetType().GetConstructors().First();
+
+			// There should be exactly two parameters with a `RequiredAttribute`:
+			var paramWithRequiredAttributeCount =
+				constructor.GetParameters().Where(p => p.IsDefined(typeof(RequiredAttribute), true)).Count();
+			Assert.AreEqual(2, paramWithRequiredAttributeCount);
+
+			// There should be exactly one parameter named `intParam`:
+			var intParam = constructor.GetParameters().SingleOrDefault(p => p.Name == "intParam");
+			Assert.NotNull(intParam);
+
+			// That parameter should have exactly one `RequiredAttribute` carrying the value 1.
+			var intParamRequiredAttrs = intParam.GetCustomAttributes<RequiredAttribute>(inherit: true);
+			Assert.AreEqual(1, intParamRequiredAttrs.Count());
+			Assert.AreEqual(1, intParamRequiredAttrs.Single().DefaultValue);
+
+			// There should be exactly one parameter named `stringParam`:
+			var stringParam = constructor.GetParameters().SingleOrDefault(p => p.Name == "stringParam");
+			Assert.NotNull(intParam);
+
+			// That parameter should have exactly one `RequiredAttribute` carrying the value "test".
+			var stringParamRequiredAttrs = stringParam.GetCustomAttributes<RequiredAttribute>(inherit: true);
+			Assert.AreEqual(1, stringParamRequiredAttrs.Count());
+			Assert.AreEqual("test", stringParamRequiredAttrs.Single().DefaultValue);
+		}
+	}
+}
--- a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
@@ -239,12 +239,14 @@ namespace Castle.DynamicProxy.Generators
 			var constructor = emitter.CreateConstructor(args);
 			if (baseConstructorParams != null && baseConstructorParams.Length != 0)
 			{
-				var last = baseConstructorParams.Last();
-				if (last.ParameterType.GetTypeInfo().IsArray && last.IsDefined(typeof(ParamArrayAttribute)))
+				var offset = 1 + fields.Length;
+				for (int i = 0, n = baseConstructorParams.Length; i < n; ++i)
 				{
-					var parameter = constructor.ConstructorBuilder.DefineParameter(args.Length, ParameterAttributes.None, last.Name);
-					var info = AttributeUtil.CreateInfo<ParamArrayAttribute>();
-					parameter.SetCustomAttribute(info.Builder);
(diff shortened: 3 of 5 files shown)


src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -187,6 +187,26 @@ public static IEnumerable<CustomAttributeInfo> GetNonInheritableAttributes(this
...
 		public static IEnumerable<CustomAttributeInfo> GetNonInheritableAttributes(this ParameterInfo parameter)
+		{
+			return parameter.GetCustomAttributesToBeReproduced(ignoreInheritance: false);
+			//                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
+			// The consequence of this is that custom attributes with `[AttributeUsage(Inherited = true)]` are not
+			// returned, because it is assumed that `parameter` will be able to inherit the custom attribute from
+			// a base type's corresponding method parameter.
+		}
+
+		internal static IEnumerable<CustomAttributeInfo> GetCustomAttributesToBeReproducedOnProxyCtor(this ParameterInfo parameter)
```
stakx: (Let's keep this `internal` in order to not expand the public API surface unnecessarily).

src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
```diff
@@ -239,12 +239,14 @@ protected void EnsureOptionsOverrideEqualsAndGetHashCode(ProxyGenerationOptions
...
-				var last = baseConstructorParams.Last();
-				if (last.ParameterType.GetTypeInfo().IsArray && last.IsDefined(typeof(ParamArrayAttribute)))
+				var offset = 1 + fields.Length;
+				for (int i = 0, n = baseConstructorParams.Length; i < n; ++i)
 				{
-					var parameter = constructor.ConstructorBuilder.DefineParameter(args.Length, ParameterAttributes.None, last.Name);
-					var info = AttributeUtil.CreateInfo<ParamArrayAttribute>();
-					parameter.SetCustomAttribute(info.Builder);
+					var parameterBuilder = constructor.ConstructorBuilder.DefineParameter(offset + i, baseConstructorParams[i].Attributes, baseConstructorParams[i].Name);
+					foreach (var attribute in baseConstructorParams[i].GetCustomAttributesToBeReproducedOnProxyCtor())
```
stakx: For regular methods, we'd call `baseMethodParams[i].GetNonInheritableAttributes()`, but we cannot do that here because no attributes would ever be returned by it due to the non-matching ctor signatures. We therefore need a new method that does all the work `AttributeUtil.GetNonInheritableAttributes()` does, *except* for excluding inheritable custom attributes.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -153,7 +153,7 @@ public static IEnumerable<CustomAttributeInfo> GetNonInheritableAttributes(this
 #else
 				var attributeType = attribute.AttributeType;
 #endif
-				if (ShouldSkipAttributeReplication(attributeType))
+				if (ShouldSkipAttributeReplication(attributeType, ignoreInheritance: false))
```
stakx: `ignoreInheritance: false` will result in the same behavior that was there before this parameter was added.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -204,7 +224,7 @@ public static IEnumerable<CustomAttributeInfo> GetNonInheritableAttributes(this
 				var attributeType = attribute.AttributeType;
 #endif
 
-				if (ShouldSkipAttributeReplication(attributeType))
+				if (ShouldSkipAttributeReplication(attributeType, ignoreInheritance))
```
jonorossi: Could you use `(parameter.Member as ConstructorInfo) != null` rather than those new methods.

Not sure why I wrote that inheritance could be the problem in the issue, because obviously constructors are not part of an class inheritance contract and must be redefined, but maybe I was thinking it could be this code excluding them because they are marked inheritable 😉.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Enhancements:
 - Added .NET Standard/.NET Core support for NLog (@snakefoot, #200)
 - Added .NET Standard/.NET Core support for log4net (@snakefoot, #201)
 - DynamicProxy supported C# `in` parameter modifiers only on the .NET Framework up until now. Adding .NET Standard 1.5 as an additional target to the NuGet package makes them work on .NET Core, too (@stakx, #339)
+- Replicate custom attributes on constructor parameters in the generated proxy type constructors (@stakx, #341)
```
jonorossi: Maybe add something like:

"... to fulfil introspection of constructors, this does not change the proxying behaviour"

So it doesn't sound like a defect.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
stakx: I don't quite understand. Could you please clarify which new methods you'd like to avoid and where that condition should go?

* `ShouldSkipAttributeReplication` needs an additional parameter so that it knows whether or not to bypass the check of `AttributeUsageAttribute.Inherited`. It's only given a `Type` but no `ParameterInfo`, after all, so a `parameter.Member is ConstructorInfo` check cannot go inside this method.

* `parameter.Member is ConstructorInfo` can however be placed in the original `GetNonInheritableAttributes` to avoid the two new forwarding methods `GetCustomAttributesToBeReproduced` and `GetCustomAttributesToBeReproducedOnProxyCtor`. However, doing so means that `GetNonInheritableAttributes` will in some circumstances return inheritable attributes, which goes directly against its name and *might* seem like a bug to someone coming across that code for the first time.

CHANGELOG.md
stakx: Good idea. Done.

fir3pho3nixx:
@stakx @jonorossi I am behind on my gh stuff because I am very busy at work, can you give me one more day to look into this properly? I will check it out tomorrow. Promise.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
jonorossi: Yep, I would like to avoid the 2 new forwarding methods as they seem unnecessary. I was thinking you could determine if you are handling a constructor in `GetNonInheritableAttributes`, pretty much thinking:

```csharp
bool ignoreInheritance = (parameter.Member is ConstructorInfo);
```

Since constructors in .NET are not inherited (not that you'd want them to be), I don't see any problem returning all attributes via this named method because no constructor parameters can ever be inherited.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -243,6 +263,11 @@ private static bool ShouldSkipAttributeReplication(Type attribute)
 				return false;
 			}
 
+			if (ignoreInheritance)
```
jonorossi: It might be best to wrap the actual inheritance check with `if (!ignoreInheritance)` instead in case another check gets added to this method since at the moment the ordering doesn't really matter.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
stakx: Ah, gotcha. Makes sense now. Done.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
stakx: Done.

src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -243,13 +245,18 @@ private static bool ShouldSkipAttributeReplication(Type attribute)
...
+				}
+
+				return true;
 			}
 
-			return true;
+			return false;
 		}
 
 		private static bool SpecialCaseAttributeThatShouldNotBeReplicated(Type attribute)
```
jonorossi: While we are here cleaning things up could you inline this method, it just complicates things for no reason, there is only the one call site for `SpecialCaseAttributeThatShouldNotBeReplicated` anyway.

If you want to rebase anything nows the time and we'll get this one merged.

stakx: I was thinking the same thing. Will do.

stakx:
Squashed a little and rebased on top of current master, should be good to go.

jonorossi:
All good, merging.

src/Castle.Core.Tests/DynamicProxy.Tests/ConstructorWithAttributesOnParametersTestCase.cs
```diff
@@ -0,0 +1,57 @@
...
+			Assert.NotNull(intParam);
+
+			// That parameter should have exactly one `RequiredAttribute` carrying the value 1.
+			var intParamRequiredAttrs = intParam.GetCustomAttributes<RequiredAttribute>(inherit: true);
+			Assert.AreEqual(1, intParamRequiredAttrs.Count());
+			Assert.AreEqual(1, intParamRequiredAttrs.Single().DefaultValue);
+
+			// There should be exactly one parameter named `stringParam`:
+			var stringParam = constructor.GetParameters().SingleOrDefault(p => p.Name == "stringParam");
+			Assert.NotNull(intParam);
```
gliljas: Should be "stringParam"

stakx: Well spotted, thanks! Feel free to send a PR -- otherwise I'll probably get around to it next week or so.

Issue #291, opened 2017-07-25
Nullable parameters on controller methods


Worthy7:
I have explained the issue quite well here:
aspnetboilerplate/aspnetboilerplate#2355 (comment)

Basically, if a method uses nullable parameters (int?) then this causes Castle.Windsor to error because it doesn't seem to want to use null, and instead replaces it with Missing.

jonorossi:
@Worthy7 just in case you weren't aware this is the Castle Core issue tracker not the Castle Windsor one. I see you logged https://github.com/castleproject/Windsor/issues/306 which you then closed, however here is likely the right place since it sounds Castle DynamicProxy related.

Can you provide a reproduction (i.e. sample code) of how Castle DynamicProxy is failing/working incorrectly using just Castle Core without ASP.NET Boilerplate or ASP.NET MVC. `System.Reflection.Missing.Value` is a valid value for performing reflection calls with optional parameters, this may be a Castle DynamicProxy defect in this case or something ASP.NET MVC isn't handling.

Worthy7:
Oh I thought this was core related not Windsor related, because the exception is Castle.Proxies

I'll reopen that issue.

Worthy7:
I see it was actually reported a long long time ago by Abp owner himself,
castleproject/Core#45

His code I assume is still a valid example

jonorossi:
The example in #45 is still reproducible thanks, you'll see why that issue didn't go anywhere since @hikalkan closed it. #35 is an earlier one on the same problem, however I think the fix in 17c1596b1e7f547eaa15479ada0b793a306ab991 (back in 2013) might not have actually fixed the problem and resulted in the default value being left as `System.Reflection.Missing` (I've not confirmed the `ParameterBuilder` behaviour).

DynamicProxy used to set the default value on the generated type prior to 4.0, but that was causing all sorts of problems because `ParameterBuilder.SetConstant` is buggy, see #141 and you'll see a web of linked issues. We decided for 4.0 that optional parameters are 99% of the time just a compile time syntactic sugar feature and doesn't really affect DynamicProxy at runtime, however the default value obviously does go along with the metadata and developers can access the metadata of the proxy types using reflection.

[...]

PR #356, opened 2018-05-06
Add back ability to reproduce default parameter values

stakx:
@jonorossi - This is in response to https://github.com/castleproject/Core/issues/291#issuecomment-359459169 and low-priority. If you've changed your mind since January and would rather not merge this, that's no problem.

Reproduction of default parameter values has caused a lot of trouble in the past and was eventually (in #149) disabled completely, because `ParameterBuilder.SetConstant` was deemed too buggy. At that time, it wasn't clearly understood what bugs exactly caused the problems.

I've spent some time discovering and documenting the relevant bugs that are present in current versions of the CLR, CoreCLR, and on Mono. This commits adds back the ability to reproduce default parameter values, but this time with more precise error detection.

This means that on recent runtimes, default parameter values will stand a good chance of being copied much more faithfully.

:warning: I've tested this only on .NET 3.5, .NET 4.7.1, .NET Core 1.1, .NET Core 2.0, a private (dev) build of .NET Core 2.1, and Mono 5.10.1.47. It is possible that on older runtime versions (esp. of Mono and .NET 4-4.7.0) , this *may* cause bugs if merged. Testing for older versions isn't always easy: For instance, I can't test locally for e.g. .NET 4.0 since 4.x versions are in-place updates and I have 4.7.1.

~~This is work in progress and shouldn't be considered for merging until the following have been ticked off:~~

* [x] Systematically add unit tests that exercise default value reproduction for any and all possible types.
* [x] Remove or rewrite existing unit tests that are covered by the above, but were based on a less precise understanding of the framework bugs.
* [x] Add a user documentation page explaining why default values aren't always reproduced correctly ([see it rendered in my fork](https://github.com/stakx/Castle.Core/blob/default-value-replication/docs/dynamicproxy-optional-parameter-value-limitations.md))

/cc #35 + #36, #45, #87 + #91, #141 + #149, #291


src/Castle.Core/DynamicProxy/Generators/Emitters/MethodEmitter.cs
```diff
@@ -160,6 +160,117 @@ private void DefineParameters(ParameterInfo[] parameters)
...
+			Debug.Assert(from != null);
+			Debug.Assert(to != null);
+			Debug.Assert((from.Attributes & ParameterAttributes.HasDefault) != 0);
+
+			object defaultValue;
+			try
+			{
+				defaultValue = from.DefaultValue;
+			}
+			catch (FormatException) when (from.ParameterType == typeof(DateTime))
```
stakx: `.DefaultValue` may also throw a `FormatException` for (closed generic) enum parameter types; see https://github.com/dotnet/corefx/issues/29570. Another `catch` dedicated for this issue might be necessary here.

docs/dynamicproxy-default-parameter-values.md
```diff
@@ -0,0 +1,37 @@
...
+
+   Note that if reflection reports a default value of `null` in such cases, this is not an error, but normal `System.Reflection` behavior that is to be expected. In this case, you may also safely assume `default(SomeStruct)` to be the correct default value.
+
+   For .NET Core, the underlying cause of this problem has been documented in [dotnet/corefx#26164](https://github.com/dotnet/corefx/issues/26164). A bugfix has been merged, so this issue may disappear sometime soon after .NET Core 2.1.
+
+* **Optional parameters of some nullable `enum` type `SomeEnum?` having a non-`null` default value.** If reflection reports (via `ParameterInfo.[Raw]DefaultValue`) a default value of `Missing.Value` for such parameters, the only thing you may safely assume is that the actual default value is not `null`.
+
+   You can only find out the correct default value by double-checking the default value in the original method of the proxied type.
+
+   For .NET Core, the underlying cause of this problem has been documented in [dotnet/coreclr#17893](https://github.com/dotnet/coreclr/issues/17893).
```
stakx: Add another bullet point for the (closed generic) enum case mentioned in the other review comment.

stakx: Done. 🏁

src/Castle.Core/DynamicProxy/Generators/Emitters/MethodEmitter.cs
stakx: Done. 🏁

jonorossi:
@stakx amazing work here, especially with the runtime teams! It appears you've covered all bases, but I'll have another look tomorrow at a sensible hour.

src/Castle.Core.Tests/ExcludeOnClrAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+		private string reason;
+
+		public ExcludeOnClrAttribute(string reason)
+		{
+			this.reason = reason;
+		}
+
+		public void ApplyToTest(Test test)
+		{
+			if (test.RunState != RunState.NotRunnable && test.RunState != RunState.Skipped && !IsRunningOnMono())
```
stakx: This is actually a bit hacky. Just because we're *not* running on Mono doesn't mean that we're running on the CLR. This is good enough for the current build setup, but not ideal.

jonorossi: Oh yer, I saw that and thought this needs to be better, will give it some thought tomorrow.

stakx: I was looking at dotnet/buildtools' `[SkipOnTargetFramework]` for inspiration, and asked myself whether it would be an option for Castle.Core to drop all magic OS & runtime guessing and have the necessary parameters passed in from the command line / build script. (There would have to be some reasonable default, which might be tricky. Some people might forget to pass the parameters. Or someone might execute the tests using a different runner.)

stakx: Just discovered [`RuntimeInformation.FrameworkDescription`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription), which appears to work on .NET, .NET Core, and Mono. How about reimplementing a single `[ExcludeOnRuntime(<runtime id>)]` based on that? In a first iteration, it would perhaps be version-agnostic, that could still be added later on.

Apart from that, BenchmarkDotNet might be another good place to find runtime / OS detection logic.

And finally, there's always the option of dropping all runtime / OS detection magic and having the required parameters passed in from the build script (but that seems tricky for a couple reasons).

jonorossi: I agree that it would be preferred to do at runtime rather than pass a parameter into the runner so we don't run into problems with using different test runners.

BenchmarkDotNet treats "CLR" as just the full/desktop CLR, and not also the CoreCLR. A single attribute looks like a good option as we might want to enable some on the CoreCLR once things get fixed.

Regarding the attribute name though, should it be "runtime" or "framework"? For example, the runtime is CoreCLR but heaps of the code we depend on is in CoreFX, so in general the framework is .NET Core? If we go with "runtime" we could probably do something like this, maybe even drop "runtime" from the attribute name as you'd always have the enum:

```
[ExcludeOnRuntime(Runtime.Mono)]
[ExcludeOnRuntime(Runtime.Clr)]
[ExcludeOnRuntime(Runtime.CoreClr)]
[ExcludeOnRuntime(Runtime.Clr | Runtime.CoreClr)]
[ExcludeOnRuntime(Runtime.ClrAndCoreClr)] // I think I prefer the bitwise one
```

With respect to determining which runtime/framework the tests are running on I don't mind requiring .NET 4.7.1 (supports Windows 7 SP1) for the unit tests, that is what BenchmarkDotNet is using for their shipping code anyway, and falling back to conditional compilation.

docs/dynamicproxy.md
```diff
@@ -18,6 +18,7 @@ If you're new to DynamicProxy you can read a [quick introduction](dynamicproxy-i
 * [Use proxy generation hooks and interceptor selectors for fine grained control](dynamicproxy-fine-grained-control.md)
 * [SRP applies to interceptors](dynamicproxy-srp-applies-to-interceptors.md)
 * [Behavior of by-reference parameters during interception](dynamicproxy-by-ref-parameters.md)
+* [Some default parameter values may not be reproduced correctly in proxy types](dynamicproxy-default-parameter-values.md)
```
jonorossi: It might be better if the docs page title was something shorter like "Optional parameter value limitations".

src/Castle.Core.Tests/ExcludeOnClrAttribute.cs
stakx: > Regarding the attribute name though, should it be "runtime" or "framework"?

Of the two, I'm starting to tend more towards "framework", but I'm still undecided. I'll happily defer to your wise decision.

Perhaps, it doesn't matter too much, since any particular framework usually comes with its own runtime, so the two are tied together. I don't know of any two CLI runtimes that share a common framework library. Framework is the more general term of the two, so probably less restrictive if we wanted to enhance that `[Exclude...]` attribute later on.

dotnet/corefx for one prefers the term "framework": see e.g. its [`[SkipOnTargetFramework]` attribute](https://github.com/dotnet/corefx/blob/ccdb5a48a49d3e932aa00d02749777464f4c8d5b/Documentation/project-docs/developer-guide.md#skipontargetframeworkattribute) and the [`TargetFrameworkMoniker` flags enum](https://github.com/dotnet/buildtools/blob/4918a5245e955480db77a071390b91698e0c820b/src/xunit.netcore.extensions/TargetFrameworkMonikers.cs). Not sure I like it the mix-up of runtime and version into a single enumeration, though.

"Platform" would be another possibility. It's somewhat vague but that could actually be good for an attribute that we might want to expand with additional constraints later on.

> ```csharp
> [ExcludeOnRuntime(Runtime.ClrAndCoreClr)] // I think I prefer the bitwise one
> ```

Agreed, a flags enum would be nicer.

docs/dynamicproxy-default-parameter-values.md
```diff
@@ -0,0 +1,43 @@
...
+On Mono (up to and including at least version 5.10.1.47), DynamicProxy may not be able to correctly reproduce default parameter values in the proxy type for...
+
+* **Optional parameters of any nullable type `Nullable<T>`.** Reflection will likely report (via `ParameterInfo.[Raw]DefaultValue`) a default value of `null` for such parameters, regardless of whether the actual default value in the proxied method was `null` or not.
+
+   Therefore, you cannot trust the reported default value, *unless* you know you're running on a version of Mono where the underlying issues have been resolved, or when you've double-checked the default value in the original method of the proxied type.
+
+   The underlying causes have been documented in [mono/mono#8504](https://github.com/mono/mono/issues/8504) and [mono/mono#8597](https://github.com/mono/mono/issues/8597).
+
+
+## When your code runs on .NET or .NET Core
```
jonorossi: ".NET" should be ".NET Framework" (there are a few mentions below), it is important to identity the product especially when you then refer to ".NET Core" in the same sentence.

src/Castle.Core.Tests/DynamicProxy.Tests/ParameterDefaultValuesTestCase.cs
```diff
@@ -0,0 +1,406 @@
...
+		{
+			// This class has methods with optional parameters having all kinds of default values.
+			//
+			// Method name nomenclature used: `{parameter type}[_nullable]_{kind of default value}`
+			//
+			// The methods must return the default value that reflection is expected to report for the
+			// parameter. Typically this matches the declared default value, but there are exceptions!
+			//
+			// This method list is intended to be exhaustive, i.e. it should cover *all* possible
+			// kinds of default values. If you find a case that is missing, please add it. (Some lines
```
jonorossi: I'd pull the sentence in parenthesis beginning "Some lines" into its own paragraph so it is more obvious.

src/Castle.Core.Tests/ExcludeOnClrAttribute.cs
jonorossi: Let's go with framework, the framework is the distribution of a runtime and BCL. For example, UWP's runtime is CoreCLR but is a different framework.

Which option do you like for the attribute name? `[Framework]` would work more like NUnit's `[Platform]`?

```csharp
[ExcludeOnFramework(Framework.Mono)]
[ExcludeOn(Framework.Mono)]
[Framework(Exclude = Framework.Mono)]

[Framework(Framework.Mono)] // could be Mono only

[Framework(~Framework.Mono)] // this should be doable, but is it clear?
```

stakx: I'd go with the first one iff the `Framework` parameter is positional and not an optional, named one. Otherwise, I'd go with your second suggestion, which would allow for more future extensibility (e.g. `[ExcludeOn(OS = OS.Unix, Framework = Framework.NetCore, ...)]`

(The third suggestion suffers from `Exclude` being a named parameter and therefore IIRC optional. What effect would `[Framework()]` have? The last two suggestions suffer from unclear intent: Does a `[Framework]` attribute imply inclusion or exclusion?)

jonorossi: Good points. I'm also heavily leaning on the first one now too, very explicit and we don't really have (or want) unit tests that can't be run everywhere, obviously except for framework defects.

Would you now support multiple declarations on the same member (like the example below), or just merge text into one?

```csharp
[ExcludeOnFramework(Framework.NetCore, "Blah blah see dotnet/coreclr/issues/1")]
[ExcludeOnFramework(Framework.NetFramework, "Other see msconnect/2")]
public void MyTest() { ... }
```

stakx:
I cleaned up and squashed all these commits a bit.

The new `[ExcludeOnFramework]` attribute appears to be doing its job, judging from the build logs. I also added a few tests for it, just in case.

jonorossi:
All good, merging.

PR #357, opened 2018-05-11
Replace Dictionary to HashSet

tangdf:


src/Castle.Core/DynamicProxy/Generators/MethodFinder.cs
```diff
@@ -78,17 +78,14 @@ private static MethodInfo[] MakeFilteredCopy(MethodInfo[] methodsInCache, Bindin
...
-			var uniqueInfos = new Dictionary<MethodInfo, object>(MethodSignatureComparer.Instance);
+			var uniqueInfos = new HashSet<MethodInfo>(MethodSignatureComparer.Instance);
 			foreach (var info in infos)
 			{
-				if (!uniqueInfos.ContainsKey(info))
-				{
-					uniqueInfos.Add(info, null);
-				}
+				uniqueInfos.Add(info);
 			}
```
stakx: This `foreach` loop probably isn't needed at all when using the `HashSet( IEnumerable<T> collection, IEqualityComparer<T> comparer)` ctor instead.

src/Castle.Core/DynamicProxy/Generators/MethodFinder.cs
```diff
@@ -78,17 +78,14 @@ private static MethodInfo[] MakeFilteredCopy(MethodInfo[] methodsInCache, Bindin
 
 		private static object RemoveDuplicates(MethodInfo[] infos)
```
stakx: (Unrelated to this change, and I haven't looked at the call sites... but I wonder why this method's declared return type is `object`, instead of `MethodInfo[]`? Do call sites have to cast the return value back to `MethodInfo[]`?)

jonorossi: They get stored into a field `Dictionary<Type, object> cachedMethodInfosByType`, so yes this field should be updated too since there is only one call site of this private method.

src/Castle.Core/DynamicProxy/Generators/MethodFinder.cs
jonorossi: Seeing the call site it looks like this method could be removed and the call site just replaced with:

```
type.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
    .Distinct(MethodSignatureComparer.Instance)
    .ToArray();
```

jonorossi:
@tangdf you've merged your cache key changes into this branch (because you used your master) which updated this pull request, so it now conflicts.

tangdf:
I'm so sorry,I ignored the comments.

fir3pho3nixx:
Don’t worry @tangdf we have all been there. Keep it going! ;)

jonorossi:
Thanks @tangdf, merged.

PR #358, opened 2018-05-14
Improve performance of InvocationHelper cache lookups

tangdf:
`KeyValuePair<Key, Value>` not implemented interface `IEquatable<>`.


jonorossi:
Thanks @tangdf, good find, that should improve the performance of that cache lookup quite a bit. If we didn't support .NET 3.5 we could have also used a tuple.

jonorossi:
Merged. I've updated the changelog.

PR #359, opened 2018-05-26
Fix wrong `type` being passed to `IInterceptorSelector.SelectInterceptors` for class proxies

### LINKED ISSUE #74: Interceptor Selectors under CreateClassProxy are provided a System.RuntimeType rather than a System.Type
When writing an implementation of an IInterceptorSelector that is dependent on the proxied type I have found that the type argument given to the selector is incorrect; it is being given System.RunTimeType.

I have looked through the code a fair bit but unfortunately have no experience with code emitters so could not pinpoint the source, however the accepted answer on [this SO question](http://stackoverflow.com/questions/19011818/using-castle-windsor-to-register-an-interceptor-for-only-the-derived-class-not) provides a good minimal reproduction of the issue.

stakx:
This resolves #74.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
@@ -32,16 +32,26 @@ namespace Castle.DynamicProxy.Generators
 	{
 		private readonly IInvocationCreationContributor contributor;
 		private readonly GetTargetExpressionDelegate getTargetExpression;
+		private readonly GetTargetExpressionDelegate getTargetTypeExpression;
 		private readonly Reference interceptors;
 		private readonly Type invocation;
 
 		public MethodWithInvocationGenerator(MetaMethod method, Reference interceptors, Type invocation,
 		                                     GetTargetExpressionDelegate getTargetExpression,
 		                                     OverrideMethodDelegate createMethod, IInvocationCreationContributor contributor)
+			: this(method, interceptors, invocation, getTargetExpression, null, createMethod, contributor)
+		{
+		}
+
+		public MethodWithInvocationGenerator(MetaMethod method, Reference interceptors, Type invocation,
+		                                     GetTargetExpressionDelegate getTargetExpression,
+		                                     GetTargetExpressionDelegate getTargetTypeExpression,
+		                                     OverrideMethodDelegate createMethod, IInvocationCreationContributor contributor)
 			: base(method, createMethod)
 		{
 			this.invocation = invocation;
 			this.getTargetExpression = getTargetExpression;
+			this.getTargetTypeExpression = getTargetTypeExpression;
 			this.interceptors = interceptors;
 			this.contributor = contributor;
 		}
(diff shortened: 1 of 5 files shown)


src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
```diff
@@ -147,11 +157,19 @@ private Expression SetMethodInterceptors(ClassEmitter @class, INamingScope namin
...
+			{
+				targetTypeExpression = new MethodInvocationExpression(null, TypeUtilMethods.GetTypeOrNull, getTargetExpression(@class, MethodToOverride));
+			}
+
 			var emptyInterceptors = new NewArrayExpression(0, typeof(IInterceptor));
 			var selectInterceptors = new MethodInvocationExpression(selector, InterceptorSelectorMethods.SelectInterceptors,
-			                                                        new MethodInvocationExpression(null,
-				                                                        TypeUtilMethods.GetTypeOrNull,
-				                                                        getTargetExpression(@class, MethodToOverride)),
+			                                                        targetTypeExpression,
```
stakx: The simplest possible bugfix would be to wrap the result of `getTargetExpression(...)` with a call to `TypeUtil.GetTypeOrNull` only if it is not a `TypeTokenExpression` already. I discarded this simpler solution because it feels like a hack. (It requires out-of-band knowledge about the kind of `Expression` that class proxies will produce inside their version of `getTargetExpression`.) The fix proposed in this PR is more elaborate, but arguably cleaner.

src/Castle.Core/DynamicProxy/IInterceptorSelector.cs
```diff
@@ -26,7 +26,7 @@ public interface IInterceptorSelector
 		/// <summary>
 		///   Selects the interceptors that should intercept calls to the given <paramref name = "method" />.
 		/// </summary>
-		/// <param name = "type">The type declaring the method to intercept.</param>
+		/// <param name = "type">The type of the target class.</param>
```
stakx: This is in response to https://github.com/castleproject/Core/issues/74#issuecomment-69142139:

> Unfortunately we've also got a mismatch between behaviour and documented behaviour via XML comments. `type` is implemented as target type, and `method` is implemented as documented, you can get the type declaring the method to intercept via `method.DeclaringType`.

src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
```diff
@@ -32,16 +32,26 @@ public class MethodWithInvocationGenerator : MethodGenerator
...
+		private readonly GetTargetExpressionDelegate getTargetTypeExpression;
 		private readonly Reference interceptors;
 		private readonly Type invocation;
 
 		public MethodWithInvocationGenerator(MetaMethod method, Reference interceptors, Type invocation,
 		                                     GetTargetExpressionDelegate getTargetExpression,
 		                                     OverrideMethodDelegate createMethod, IInvocationCreationContributor contributor)
+			: this(method, interceptors, invocation, getTargetExpression, null, createMethod, contributor)
+		{
+		}
```
stakx: Adding the new parameter to the existing (publicly visible) ctor would be a breaking change, hence a new ctor.

jonorossi: No one should be using these classes anyway as they really are internal implementation, but happy for this to stay as is for now and can be removed later.

PR #361, opened 2018-05-27
Re-enable four skipped tests by moving them to weak-named test project

stakx:
Now that we have a weak-named project, we can re-enable four tests that verify whether proxy types end up in the right module (weak vs strong named).

### DIFF
--- a/src/Castle.Core.Tests.WeakNamed/DynamicProxy.Tests/BasicClassProxyTestCase.cs
+++ b/src/Castle.Core.Tests.WeakNamed/DynamicProxy.Tests/BasicClassProxyTestCase.cs
@@ -1,0 +1,91 @@
+﻿// Copyright 2004-2018 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


src/Castle.Core.Tests.WeakNamed/DynamicProxy.Tests/BasicClassProxyTestCase.cs
```diff
@@ -0,0 +1,91 @@
...
+
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class BasicClassProxyTestCase : BasePEVerifyTestCase
+	{
+		[Test]
+		public void ProxyForBaseTypeFromUnsignedAssembly()
+		{
+			Assert.False(TestAssemblySigned());
```
jonorossi: Could we move this into a test fixture set up and add a reason message parameter, or just get rid of the assertion completely as it is no longer really needed since this whole project has this requirement now?

stakx: Done. Let's just drop these assertions, then. Like you said, the test project itself already guarantees weak-named-ness. Even if a future maintainer were tempted to move these tests back to the original fixture (now that he explicit assertion is no longer there to prevent this), the tests would still start failing, so that should be enough of a safeguard.

PR #370, opened 2018-06-11
Prevent interceptors from being able to modify `in` parameters

### LINKED ISSUE #369: Castle allows to rewrite read-only parameters (`in` modifier)
Consider the following code sample:

```c#
[Fact]
public void TestProxy()
{
    var proxy = new ProxyGenerator().CreateInterfaceProxyWithoutTarget<IInterfaceWithIn>(new RewritingInterceptor());
    
    var val = new MutableStruct { Value = 24 };
    proxy.DoNotMutate(val);

    Assert.Equal(24, val.Value);
}

public struct MutableStruct
{
    public int Value { get; set; }
}

public interface IInterfaceWithIn
{
    void DoNotMutate(in MutableStruct value);
}

public class RewritingInterceptor : IInterceptor
{
    {
        if (invocation.Method.Name == "DoNotMutate")
        {
            invocation.SetArgumentValue(0, new MutableStruct {Value = 42});
        }
    }
}
```

The issue happens because value is passed by ref, so technically it's possible to rewrite the value. However, like Roslyn, Castle should not allow to rewrite the value via API.

Likely, the fix should be placed somewhere [here](https://github.com/castleproject/Core/blob/39457da1bd1cd835bfca4df8946fb9dcbe996c5e/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs#L25).

/cc: @stakx @blairconrad JFYI as our mock libraries contain tooling for this feature invocation.

stakx:
This should fix #369.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Prevent interceptors from being able to modify `in` parameters (@stakx, #370)
+
 ## 4.3.0 (2018-06-07)
 
 Enhancements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 Bugfixes:
 - Prevent interceptors from being able to modify `in` parameters (@stakx, #370)
 
+Deprecations:
+- `Castle.DynamicProxy.Generators.Emitters.ArgumentsUtil.IsAnyByRef` (@stakx, #370)
+
 ## 4.3.0 (2018-06-07)
 
 Enhancements:
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
@@ -1,0 +1,121 @@
+﻿// Copyright 2004-2018 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests
+{
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class InParamsTestCase : BasePEVerifyTestCase
+	{
+		[Test]
+		public void By_value_parameter_cannot_modify_argument_var()
+		{
+			var x = new ReadOnlyStruct(1);
+			var original = new ReadOnlyStruct(x.Value);
+
+			var different = new ReadOnlyStruct(x.Value + 100);
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByValue>(new SetArgumentValueInterceptor(different));
+			proxy.Method(x);
+
+			Assert.AreEqual(original.Value, x.Value);
+		}
+
+		[Test]
+		public void By_referece_In_parameter_cannot_modify_argument_var()
+		{
+			var x = new ReadOnlyStruct(1);
+			var original = new ReadOnlyStruct(x.Value);
+
+			var different = new ReadOnlyStruct(x.Value + 100);
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByReadOnlyRef>(new SetArgumentValueInterceptor(different));
+			proxy.Method(in x);
+
+			Assert.AreEqual(original.Value, x.Value);
+		}
+
+		[Test]
+		public void By_reference_Ref_parameter_can_modify_argument_var()
+		{
+			var x = new ReadOnlyStruct(1);
+			var original = new ReadOnlyStruct(x.Value);
+
+			var different = new ReadOnlyStruct(x.Value + 100);
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByRef>(new SetArgumentValueInterceptor(different));
+			proxy.Method(ref x);
+
+			Assert.AreNotEqual(original.Value, x.Value);
+		}
+
+		[Test]
+		public void By_reference_Out_parameter_can_modify_argument_var()
+		{
+			var x = new ReadOnlyStruct(1);
+			var original = new ReadOnlyStruct(x.Value);
+
+			var different = new ReadOnlyStruct(x.Value + 100);
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IOut>(new SetArgumentValueInterceptor(different));
+			proxy.Method(out x);
+
+			Assert.AreNotEqual(original.Value, x.Value);
+		}
+
+		public readonly struct ReadOnlyStruct
+		{
+			private readonly int value;
+
+			public ReadOnlyStruct(int value)
+			{
+				this.value = value;
+			}
+
+			public int Value => this.value;
+		}
+
+		public interface IByValue
+		{
+			void Method(ReadOnlyStruct arg);
+		}
+
+		public interface IByReadOnlyRef
+		{
+			void Method(in ReadOnlyStruct arg);
+		}
+
+		public interface IByRef
+		{
+			void Method(ref ReadOnlyStruct arg);
+		}
+
+		public interface IOut
+		{
+			void Method(out ReadOnlyStruct arg);
+		}
+
+		public sealed class SetArgumentValueInterceptor : IInterceptor
+		{
+			private object value;
+
+			public SetArgumentValueInterceptor(object value)
+			{
+				this.value = value;
+			}
+
+			public void Intercept(IInvocation invocation)
+			{
+				invocation.SetArgumentValue(0, value);
+			}
+		}
+	}
+}
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
@@ -14,6 +14,12 @@
 
 namespace Castle.DynamicProxy.Tests
 {
+	using System;
+	using System.Linq;
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Tests.Interceptors;
+
 	using NUnit.Framework;
 
 	[TestFixture]
@@ -26,7 +32,7 @@ namespace Castle.DynamicProxy.Tests
 			var original = new ReadOnlyStruct(x.Value);
 
 			var different = new ReadOnlyStruct(x.Value + 100);
-			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByValue>(new SetArgumentValueInterceptor(different));
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByValue>(new SetArgumentValueInterceptor(0, different));
 			proxy.Method(x);
 
 			Assert.AreEqual(original.Value, x.Value);
@@ -39,7 +45,7 @@ namespace Castle.DynamicProxy.Tests
 			var original = new ReadOnlyStruct(x.Value);
 
 			var different = new ReadOnlyStruct(x.Value + 100);
-			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByReadOnlyRef>(new SetArgumentValueInterceptor(different));
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByReadOnlyRef>(new SetArgumentValueInterceptor(0, different));
 			proxy.Method(in x);
 
 			Assert.AreEqual(original.Value, x.Value);
@@ -52,7 +58,7 @@ namespace Castle.DynamicProxy.Tests
 			var original = new ReadOnlyStruct(x.Value);
 
 			var different = new ReadOnlyStruct(x.Value + 100);
-			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByRef>(new SetArgumentValueInterceptor(different));
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IByRef>(new SetArgumentValueInterceptor(0, different));
 			proxy.Method(ref x);
 
 			Assert.AreNotEqual(original.Value, x.Value);
@@ -65,10 +71,54 @@ namespace Castle.DynamicProxy.Tests
 			var original = new ReadOnlyStruct(x.Value);
 
 			var different = new ReadOnlyStruct(x.Value + 100);
-			var proxy = generator.CreateInterfaceProxyWithoutTarget<IOut>(new SetArgumentValueInterceptor(different));
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IOut>(new SetArgumentValueInterceptor(0, different));
 			proxy.Method(out x);
 
 			Assert.AreNotEqual(original.Value, x.Value);
+		}
+
+		// The next four tests verify some of DynamicProxy's assumptions about how the compiler emits code.
+		// If any of them start failing, we need to review the above tests, as well as the implementation of
+		// `Castle.DynamicProxy.Generators.GeneratorUtil.CopyOutAndRefParameters`.
+
+		[Test]
+		[TestCase(typeof(IByReadOnlyRef))]
+		public void By_reference_In_parameter_has_ParameterAttributes_In_set(Type type)
+		{
+			var parameter = type.GetMethod("Method").GetParameters()[0];
+
+			Assert.True(parameter.Attributes.HasFlag(ParameterAttributes.In));
+		}
+
+		[Test]
+		[TestCase(typeof(IByValue))]
+		[TestCase(typeof(IByRef))]
+		[TestCase(typeof(IOut))]
+		public void Parameter_other_than_by_reference_In_does_not_have_ParameterAttributes_In_set(Type type)
+		{
+			var parameter = type.GetMethod("Method").GetParameters()[0];
+
+			Assert.False(parameter.Attributes.HasFlag(ParameterAttributes.In));
+		}
+
+		[Test]
+		[TestCase(typeof(IByReadOnlyRef))]
+		public void By_reference_In_parameter_has_IsReadOnlyAttribute(Type type)
+		{
+			var parameter = type.GetMethod("Method").GetParameters()[0];
+
+			Assert.True(parameter.GetCustomAttributes().Any(a => a.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute"));
+		}
+
+		[Test]
+		[TestCase(typeof(IByValue))]
+		[TestCase(typeof(IByRef))]
+		[TestCase(typeof(IOut))]
+		public void Parameter_other_than_by_reference_In_does_not_have_IsReadOnlyAttribute(Type type)
+		{
+			var parameter = type.GetMethod("Method").GetParameters()[0];
+
+			Assert.False(parameter.GetCustomAttributes().Any(a => a.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute"));
 		}
 
 		public readonly struct ReadOnlyStruct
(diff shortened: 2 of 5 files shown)


src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -33,13 +35,55 @@ public static class GeneratorUtil
 
 			var arguments = StoreInvocationArgumentsInLocal(emitter, invocation);
 
+			// Extracted from the following `for` loop as an optimization.
+			//
+			// The comparison by name is intentional; any assembly could define that attribute.
+			// See explanation in comment below.
+			Func<object, bool> isIsReadOnlyAttribute =
```
zvirja: Create a local method for this - it will allow compiler to cache the value better (only once, rather than per each invocation).

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -33,13 +35,55 @@ public static class GeneratorUtil
...
+				//    use of a `GetCustomAttributes` enumeration instead of a faster `IsDefined`).
+				//
+				//  * A required custom modifier `System.Runtime.InteropServices.InAttribute`
+				//    is always present in those cases relevant for DynamicProxy (proxyable methods),
+				//    but not all targeted platforms support reading custom modifiers. Also,
+				//    support for cmods is generally flaky (at this time of writing, mid-2018).
+				//
+				// The above points inform the following detection logic: First, we rely on an IL
+				// `[in]` modifier being present. This is a "fast guard" for non-`in` parameters:
+				if ((parameters[i].Attributes & ParameterAttributes.In) != 0)
```
zvirja: As to me, this method started to have too many responsibilities. Better create a standalone method like `ArgumentsUtil.IsReadOnlyByRefArg()` or similar and put all the detection logic there.

Also it might make sense to create `ArgumentsUtil.IsAnyWritableByRef` or similar and use it instead of `ArgumentsUtil.IsAnyByRef`, so we don't generate e.g. redundant `try/catch` when we are not going to write something.

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
```diff
@@ -0,0 +1,165 @@
...
+			var different = new ReadOnlyStruct(x.Value + 100);
+			var proxy = generator.CreateInterfaceProxyWithoutTarget<IOut>(new SetArgumentValueInterceptor(different));
+			proxy.Method(out x);
+
+			Assert.AreNotEqual(original.Value, x.Value);
+		}
+
+		[Test]
+		[TestCase(typeof(IByReadOnlyRef))]
+		public void By_reference_In_parameter_has_ParameterAttributes_In_set(Type type)
```
zvirja: We are mixing facade (or integration) tests here and our expectations about compiler. I'd remove all the tests about expectations, as:
- it's unclear what to do when they start to fail;
- facade tests will start to fail anyway, so these are redundant.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
stakx: I've run a few experiments to see the compiler automatically cache instantiated delegates, but I haven't been able to observe such behavior (neither in a debug nor release builds). It'll quite literally do what it's told, so the important thing is to take the hidden delegate instantiation out of the loop... it's a quick and easy win. Writing the lambda in-place instead of in a local function or static method produces less line noise. IMO we can suffer that one delegate instantiation per invocation&mdash;if not, we'd possibly have to refactor *a lot* of other code in DynamicProxy!

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
stakx: Given that the method is called `CopyOutAndRefParameters`, I think the additional 6 lines of code are still in scope.

That said, I do agree that it could be re-written to make better use of small helper methods to somewhat raise the level of abstraction.

You'll note there's already a method `AttributeUtil.IsAnyByRef`, so `AttributeUtil` might be a good home for additional similar methods such as `bool IsIn(ParameterInfo)`, `bool IsRef(ParameterInfo)`, `bool IsOut(ParameterInfo)` or just one `ParameterByRefKind GetParameterByRefKind(ParameterInfo)`. But once we start adding methods there, we also have to talk about accessibility for consistency's sake. The present-day `AttributeUtil.IsAnyByRef` is `public` which I see as a mistake. So either we expand Core's public API further, or we add methods with inconsistent accessibility, or we add the method in a different place, which is also somewhat chaotic.

This is a design problem I'd rather not solve today. May I leave it up to you to submit a PR that cleans this method up a little?

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
stakx: Partially agree. I was thinking about whether it would be better to shell out such compiler expectation tests into a separate fixture, but decided against it because it should be obvious that they are there because of, and belong with, the other unit tests. It is also true that if the compiler expectation tests start failing, the actual unit tests will fail, too. In that regard, they are superfluous. However, their added value lies in giving us an additional, very direct hint at *why* the unit tests might be failing. We've done that kind of thing in a few other places recently.

I'll happily remove them but I'd like to leave the final decision to @jonorossi (who was previously in favor of these kinds of tests).

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: Playing with it more I found this issue: https://github.com/dotnet/roslyn/issues/19431. That's the reason whey are not cached (and this might change in future). If you pass anonymous non-capturing lambda and target e.g. `net462` you'll see the caching in action:
```c#
foreach (var parameter in parameters)
{
  if (parameter.attributes.Any(t => t.FullName == "EEEE"))
  {
    // Do smth
  }
}
```
Compiled:
```c#
foreach (System.ValueTuple<string, Type[]> parameter in parameters)
{
    if (!parameter.Item2.Any(<>c.<>9__0_0 ?? (<>c.<>9__0_0 = <>c.<>9.<CycleWithLocalMethod>b__0_0)))
        continue;
}
```

Later they might enable the similar optimization for the local functions - that's why I suggested it.
Also it looks more natural with local methods. But as this a matter of tastes - it's up to you 😉

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: Well, if you don't want to produce too much noise, then I'd suggest to simply extract all the `IsReadOnlyRef` logic to a static method in the same class. Each method should have single responsibility only, while currently `CopyOutAndRefParameters` method orchestrates parameters handling and knows the particularity of how `in` modifier is implemented.

stakx:
@zvirja, thanks for your review. You know what, I think you were right. I've rewritten the method to make use of local functions, and I've marked `ArgumentsUtil.IsAnyByRef` as `[Obsolete]`. Things look a little cleaner now (but @jonorossi, I can always revert if you'd prefer the previous closer-to-the-bare-metal approach).

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
jonorossi: I'm definitely in favour of keeping these additional tests that are sort of like proofs, if it happens the implementation of the runtime changes or we misunderstood how it works we'll know straight away. They have heaps of value for DynamicProxy where they wouldn't in most libraries because we are so low in the stack and make so many assumptions about implementation details that a lot of the time aren't documented in the ECMA specifications.

The only thing that would be good is a "heading-like" comment like you've done in the others to separate them from the actual unit tests.

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.False(parameter.Attributes.HasFlag(ParameterAttributes.In));
+		}
+
+		[Test]
+		[TestCase(typeof(IByReadOnlyRef))]
+		public void By_reference_In_parameter_has_IsReadOnlyAttribute(Type type)
+		{
+			var parameter = type.GetMethod("Method").GetParameters()[0];
+
+			Assert.True(parameter.GetCustomAttributes().Any(a => a.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute"));
```
jonorossi: I thought these classes were public classes, I assume they aren't visible in .NET Standard 1.3?

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
```diff
@@ -0,0 +1,165 @@
...
+		{
+			void Method(ref ReadOnlyStruct arg);
+		}
+
+		public interface IOut
+		{
+			void Method(out ReadOnlyStruct arg);
+		}
+
+		public sealed class SetArgumentValueInterceptor : IInterceptor
```
jonorossi: This class should probably sit alongside `Castle.DynamicProxy.Tests.Interceptors.SetReturnValueInterceptor`, and also accept an index.

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
stakx: Sure, I'll add such a comment!

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -35,12 +37,62 @@ public static class GeneratorUtil
...
+					// be saving much time anyway.
+					if (parameter.GetCustomAttributes(false).Any(IsIsReadOnlyAttribute))
+					{
+						return true;
+					}
+				}
+
+				return false;
+
+				bool IsIsReadOnlyAttribute(object attribute)
```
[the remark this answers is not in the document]
stakx: 😆, I almost thought that might be your reaction!

The idea here is to make the scope of helper functions as local as possible, so that it's clear where they are actually needed. Contrast this with a "global" helper function like `ArgumentsUtil.IsAnyByRef` which, as it turned out, was only used in one single location&mdash;but you couldn't tell by taking a casual glance at its source.

Options:

1. Leave it as it is now (nested local functions).
2. Bring `IsIsReadOnlyAttribute` to the same level as the other local functions to reduce nesting.
3. Turn the three local functions into `private static` methods in the same class (`GeneratorUtil`).
4. Revert to how things were before (i.e. don't have local functions at all).

What's your preference?

src/Castle.Core.Tests/DynamicProxy.Tests/InParamsTestCase.cs
stakx: Done.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Prevent interceptors from being able to modify `in` parameters (@stakx, #370)
+
+Deprecations:
+- `ArgumentsUtil.IsAnyByRef` (@stakx, #370)
```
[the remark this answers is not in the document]
stakx: Done.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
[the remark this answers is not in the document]
stakx: @jonorossi: 😆, I almost thought that might be your reaction!

The idea here is to make the scope of helper functions as local as possible, so that it's clear where they are actually needed. Contrast this with a "global" helper function like `ArgumentsUtil.IsAnyByRef` which, as it turned out, was only used in one single location&mdash;but you couldn't tell by taking a casual glance at its source.

Options:

1. Leave it as it is now (nested local functions).
2. Bring `IsIsReadOnlyAttribute` to the same level as the two other local functions to reduce nesting.
3. Turn the three local functions into `private static` methods in the same class (`GeneratorUtil`).
4. Revert to how things were before (i.e. don't have local functions at all).

What's your preference?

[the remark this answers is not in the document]
jonorossi: Let's just leave it as is, I didn't really have a preference other than being a little surprised. I think if `CopyOutAndRefParameters` was non-void it would look strange because you'd have to try to match up returns.

stakx:
@jonorossi - thanks for entrusting me with the reigns at this point. :heart: Unfortunately, Core's `master` branch is currently read-only to me. (Ironically, I *do* have write access on other Castle projects.) 🤔

jonorossi:
> Unfortunately, Core's master branch is currently read-only to me

Fixed. Castle Core and Windsor have "protected" master branches from when there were other people working on a single .NET Core branch that weren't committers.

zvirja:
Hi @stakx,

I'm trying to consume this change in NSubstitute and have complications with the way we perform the check.
Everything works fine for the interface proxy. However, to support delegate proxy NSubstitute uses an approach similar to Moq:
1. Generate dynamic interface with a single `Invoke` method, matching the delegate's signature.
2. Obtain proxy for that interface from Castle.
3. Create delegate from proxy's `Invoke` method.

The complication is that for .NET Framework the compiler generates the `IsReadOnly` attribute as an internal type in the module where the delegate is defined. Therefore, when I'm generating an interface (point 1 from above), I cannot simply re-use the existing `IsReadOnly` attribute - it's internal and is not visible by the dynamic assembly. As result, the interface's `Invoke` method doesn't contain the `IsReadOnly` attribute and later (at point 2) Castle doesn't make the method read-only.

You'll see the same issue with Moq if you decide to guarantee that mock for delegate with `in` modifier is indeed read-only.

I see a way to fix that (generate `IsReadOnly` attribute in dynamic assembly), however it looks overcomplicated. Therefore, I wanted to clarify your initial motivation for having the check like this:
https://github.com/castleproject/Core/blob/1ba6c894ea4b011b524782f2eec6e89b6245627c/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs#L77-L85

Why can't we simply check for the `In` required modifier here? It's fully supported by .NET Standard 1.5+, so it looks like a reliable way to do that. I believe we shouldn't worry about .NET Standard 1.3 as `in` modifiers are not supported there neither. This way it will be much simpler to make everything work, as `In` modifier type is already a part of .NET Standard.

Thanks. Sorry for not testing this before the release - I was almost confident that everything must work fine... 😖

stakx:
@zvirja:

Thank you for following up on this.

I should have realised in time that the compiler-generated `System.Runtime.CompilerServices.IsReadOnlyAttribute` shim is non-public and thus often non-copyable by DynamicProxy. So we have the situation that DynamicProxy recognizes C# `in` parameters by the presence of that custom attribute, but may generate proxy types that fail the very same test.

What a pity I didn't realise this earlier. To be extra-super-correct, DynamicProxy would have to follow the compilers' lead and emit the `IsReadOnlyAttribute` into the dynamic assembly on the fly if it doesn't already exist in an accessible location. Alas, let's wait and see if anyone actually needs this. Right now, I'm not too keen on implementing this to be honest.

> Everything works fine for the interface proxy. 

I take it that this is because you're using a `in` parameter detection logic based on custom modifiers (cmods) instead of the custom attribute?

> Why can't we simply check for the `In` required modifier here?

https://github.com/castleproject/Core/blob/1ba6c894ea4b011b524782f2eec6e89b6245627c/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs#L70-L73

Like you're saying, we *could* do this. However, I've spent quite a bit of time with cmods in the past few months, and it wasn't super enjoyable. None of the major runtimes (Mono, .NET Framework, .NET Core) support cmods very well when it comes to Reflection and Reflection.Emit. I think it entirely possible that yet more cmod-related bugs in these facilities will come to the fore as C# continues to rely on them more and more. Hence I figured it would be best to avoid them if possible, and decided against relying on cmods for the detection logic.

Add to this the fact that the *primary* characteristic of C# `in` parameters is the "duck-typed" custom attribute anyway, so it was a fairly easy decision to rely on *that* instead.

So, in conclusion, for the time being at least, you'll have to emit a `IsReadOnlyAttribute` yourself (it needs to be visible to DynamicProxy) and add it to the `Invoke` parameters that have the `InAttribute` cmod.

> BTW, it looks like a typo in spec, as they use `System.Runtime.InteropServices.InAttribute` modifier type

(Agree! I've taken the liberty to ask over at [dotnet/csharplang](https://github.com/dotnet/csharplang/issues/1655) whether this is a typo or not. Let's see what they say.)

zvirja:
> None of the major runtimes (Mono, .NET Framework, .NET Core) support cmods very well when it comes to Reflection and Reflection.Emit.

Could you please be more specific here? What exact issues did you see? I'm just curios, as I tested a few versions of both .NET Framework and .NET Core and it worked well there. Probably, not the generics (as you already reported), but apart from that it worked fine.

> Given that we have a fast-guard check for ParameterAttributes.In, we could probably afford to do an additional check for the cmod.

Looks like a great trade-off! If method has `in` modifier, we could test for _either_ custom modifier or custom attribute. If any of them is available - we consider parameter as a read-only.

It will add an additional robustness to the detection logic and support non-perfect code-gen scenarios. And would also simplify the NSubstitute code as it will be enough to emit custom modifier only 😊 I could raise a PR for that if you wish.

> (Agree! I've taken the liberty to ask over at dotnet/csharplang whether this is a typo or not.

Thanks for registering that issue 👍 Likely I should be that proactive to do it by myself... 😖

stakx:
>> None of the major runtimes [...] support cmods very well [...]
>
> Could you please be more specific here? [...] Probably, not the generics (as you already reported) [...]

Fair question. I should be more specific, but in truth, I cannot. I guess it comes down to being a trust issue for me. Consider:

Custom modifiers have historically been used very little outside C++/CLI, whose community was relatively small (compared to e.g. that of C#). Add to this that the typical use case for C++/CLI was to be a fast and easy interop bridge between managed and unmanaged/native code, *not* to do heavy-duty reflection stuff like DynamicProxy does. So if there *are* in fact cmod-related bugs in .NET's managed Reflection API, that would likely have gone unnoticed by most people for the longest time. Not even Microsoft's compiler team might have noticed since the compilers were native programs using the CLR's native APIs.

C# has started to use cmods for its own purposes only quite recently, and almost right away we walk into bugs/limitations with Reflection. (You can look up at some of the work I did in DynamicProxy, or some of the issues I raised over at the Mono, CoreCLR, and CoreFX repos. The CoreCLR issues are also representative for the CLR.) That's not exactly reassuring. Who knows what else we might discover.

On the other hand, custom attributes have long been used by the C# community, so they are much more "battle-tested". This is why I trust in them a lot more than I trust in Reflection's ability to handle cmods properly.

It doesn't help that one of the code owners of Reflection / Reflection.Emit admits ([here](https://github.com/dotnet/coreclr/pull/17881#issuecomment-387114966)) that Reflection has fundamental limitations (even though they might not matter much in practice) when it comes to cmods:

   > This is a long standing defect in how Reflection thinks about custom modifiers and it's not likely you could address this [...]

That all being said, you're of course right that the very scenario we're looking at doesn't appear to be affected.

I'll let you judge for yourself whether it's right to be cautious about cmods or not.

> Just tested the check like this: [...]

Have you checked on Mono, as well?

> Looks like a great trade-off! If method has `in` modifier [sic], we could test for either custom modifier or custom attribute. If any of them is available - we consider parameter as a read-only.

(I guess you meant "if the parameter has the `ParameterAttributes.In` flag set, we could...")

Yes, precisely.

> I could raise a PR for that if you wish.

Sounds great! I feel a little saturated with cmod-related work right now, so I'd be totally happy to defer to your initiative!

One suggestion, though: It would be good to make sure that the faster of the two checks is made first. I'm suspecting that the `modreq` check will generally outperform a custom attributes-based check (but it would be awesome if you could actually benchmark this using a variety of different scenarios):

* Parameters are generally likely to have less cmods than custom attributes, so `parameter.GetRequiredCustomModifiers()` should return an array that tends to be smaller (giving us less things to search through) than `parameter.GetCustomAttributes(false)`.

* `parameter.GetCustomAttributes(false)` requires us to look for an object of a type which we compare by name. `parameter.GetRequiredCustomModifiers()` OTOH returns `Type[]`, and assuming that Roslyn actually encodes the right type as `modreq` (we need to wait for confirmation of this!), it's one that's always available in the BCL, so we can compare using simple `.Equals` instead of by name.

zvirja:
@stakx Thanks for the verbose reply! 😉 

>  This is why I trust in them a lot more than I trust in Reflection's ability to handle cmods properly.

I see your point now. Likely you are right that cmods support is quite limited and indeed your possibilities are quite limited. However, probably, for our case it doesn't matter.

In any case if we tests for both modifiers and attributes, than I suppose we play safely 😉 

> Sounds great! I feel a little saturated with cmod-related work right now, so I'd be totally happy to defer to your initiative!

Sure. Will do after [this one](https://github.com/dotnet/csharplang/issues/1655) is closed.

> One suggestion, though: It would be good to make sure that the faster of the two checks is made first.

Sure :)

---
In the meanwhile I've adjusted the NSubstitute to generate the `IsReadOnly` attribute in runtime if needed. If you could take a look at https://github.com/nsubstitute/NSubstitute/pull/420 it would be awesome (code is not very specific to NSubstitute there) as you might want to do the same in Moq.

PR #373, opened 2018-06-13
DynamicProxy documentation: Add Entity Framework Core to list of users

stakx:
It's good to know who is using DynamicProxy further downstream.

One question about etiquette: Should we ask someone from the EF Core team first whether they're OK with being listed in our docs?


jonorossi:
> It's good to know who is using DynamicProxy further downstream.

That's news to me, nice! I always just assumed EF had lazy loading implemented using their own code, but it looks like EF Core only got lazy loading in 2.1 a few months ago.

> One question about etiquette: Should we ask someone from the EF Core team first whether they're OK with being listed in our docs?

I see no need, it is a public project and you've just told everyone here about it so it is no secret 😉.

stakx:
Yes, I was quite surprised, too! (Found out about it via https://github.com/dotnet/corefx/issues/29365#issuecomment-396919752 btw.)

Issue #374, opened 2018-06-13
System.Reflection.Emit package has been delisted


jonorossi:
See https://github.com/dotnet/corefx/issues/29365#issuecomment-385102487.

According to that thread the package was never cleanly implemented as a .NET Standard 1.x package and had a hack that it could see the internals of the runtime's classes, it was then recently decided for .NET Standard 2.0 not to continue that hack (because CoreRT doesn't have the hack). No .NET Standard 2.x package was published and recently on the 2018-04-28 `System.Reflection.Emit` and `System.Reflection.Emit.Lightweight` were delisted on NuGet.org.

That now leaves us [using a delisted package while targeting .NET Standard 1.3 and 1.5](https://github.com/castleproject/Core/blob/7ef5e620693e9548dbc0246adc91c3f9ba734b3d/src/Castle.Core/Castle.Core.csproj#L35-L45). We need to follow this thread, however it likely means we'll need to drop our .NET Standard 1.3 and 1.5 targets and move to `netcoreappx.x` (.NET Core) targets, obviously resulting in our downstream users doing the same. The .NET Framework targets (`net35;net40;net45`) wouldn't change.

@stakx is this your understanding?

stakx:
> [It] likely means we'll need to drop our .NET Standard 1.3 and 1.5 targets and move to `netcoreappx.x` (.NET Core) targets, obviously resulting in our downstream users doing the same. The .NET Framework targets (`net35;net40;net45`) wouldn't change.

Yes, @jonorossi, I'm drawing the same conclusion.

Reflection.Emit has always been a problematic part of .NET Standard because some platforms don't actually implement that facility while still nominally declaring .NET Standard compatibility. So your .NET Standard library ends up getting bug reports saying that it doesn't work properly&mdash;but there's nothing you can do about it. (Been there.)

Reflection.Emit might still be brought back to .NET Standard. But even then, Reflection.Emit will simply throw `NotSupportedException` or `NotImplementedException` on those platforms. So we'd still be pretending to support platforms that we cannot in fact run on.

This is one of several ways in which .NET Standard is a fundamentally broken standard, and I'm personally much in favor of abandoning it and targeting instead the actual platforms that really *do* support Reflection.Emit. It's simply the most transparent and accurate thing to do, and also what Microsoft appears to recommend these days.

That said, it would be difficult to get downstream libraries to abandon .NET Standard because they'd likely have to carry the burden, i.e. to face a public outcry from their user base.

[...]

PR #378, opened 2018-06-14
Cross target netstandard2.0

iamcarbon:
Cross targeting netstandard2.0 eliminates all dependencies on modern runtimes.

### DIFF
--- a/buildscripts/common.props
+++ b/buildscripts/common.props
@@ -95,6 +95,14 @@
 		<DefineConstants>$(NetStandard15Constants)</DefineConstants>
 	</PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


stakx:
> Cross targeting netstandard2.0 eliminates all dependencies on modern runtimes.

Hi @iamcarbon as you've noticed, that's not quite accurate: Your build fails because System.Reflection.Emit, which is the foundation on which DynamicProxy is built, is not part of .NET Standard, but provided via a (now delisted) NuGet package; see the current discussion in #374. :wink:

iamcarbon:
@stakx Agreed. Closing this PR and will keep a lookout on these discussions.

Issue #377, opened 2018-06-13
Odd locking strategy in AreInternalsVisibleToDynamicProxy


TimLovellSmith:
I can't figure out 
  a) why it tests `internalsVisibleToDynamicProxy.ContainsKey()` again after Upgrading the lock.
  b) why it doesn't do the first codepath with a non-upgradeable read lock, similar to BaseProxyGenerator, although maybe a reasonable answer could be they have different expectations on how likely they are to need to upgrade the lock...

```
internal static bool AreInternalsVisibleToDynamicProxy(Assembly asm)
		{
			using (var locker = internalsVisibleToDynamicProxyLock.ForReadingUpgradeable())
			{
				if (internalsVisibleToDynamicProxy.ContainsKey(asm))
				{
					return internalsVisibleToDynamicProxy[asm];
				}

				locker.Upgrade();

				if (internalsVisibleToDynamicProxy.ContainsKey(asm))
				{
					return internalsVisibleToDynamicProxy[asm];
				}
```

TimLovellSmith:
InvocationHelper's GetMethodOnType() is another case where it rechecks the cache after upgrading the lock, though I don't know why it is needed.

fir3pho3nixx:
@stakx I think this is reader writer locks. There was also a slim version but I am not sure we are using this.

TimLovellSmith:
@stakx Sorry, the point of the issue is not only that it looks funny but it seems like a good chance this could be rewritten for better perf: either because it is doing redundant call to GetFromCache which can be eliminated (since nobody else should be able to write to it while you have an upgradeable read lock), or because it is not avoiding contention on the cache lock by doing a non-upgradeable-read-lock while checking the cache the first time through.

stakx:
**P.S.: I've originally posted the following in your PR. I'm reposting it here because this is meant more for further discussion than for as a suggestion to implement.*

Your PR looks good to me. Sorry for initially missing the point of your enquiry.

The following is just an additional idea; feel free to disregard. (I'm not a multi-threading expert at all, so perhaps this is naïve.)

It seems that the locks are mostly used in combination with `Dictionary<,>`. I was wondering whether it might make sense to push the locking that's currently strewn all over the place down into a new dictionary-like type, say:

```csharp
// using System.Collections.Generic;
// using System.Threading;

internal sealed class SynchronizedDictionary<TKey, TValue> : IDisposable
{
    private Dictionary<TKey, TValue> dictionary;
[...]

PR #379, opened 2018-06-14
More consistent locking code across InvocationHelper, ProxyUtil, Base…

TimLovellSmith:
…ProxyGenerator. In answer to issue #377

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
@@ -415,9 +415,10 @@ namespace Castle.DynamicProxy.Generators
 				var proxyType = factory.Invoke(name, Scope.NamingScope.SafeSubScope());
 
 				// Upgrade the lock to a write lock. 
-				locker.Upgrade();
-
-				AddToCache(cacheKey, proxyType);
+				using (locker.Upgrade())
+				{
+					AddToCache(cacheKey, proxyType);
+				}
 				return proxyType;
 			}
 		}
--- a/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
@@ -48,6 +48,15 @@ namespace Castle.DynamicProxy.Internal
 
 			Debug.Assert(proxiedMethod.DeclaringType.IsAssignableFrom(type),
 						 "proxiedMethod.DeclaringType.IsAssignableFrom(type)");
+			using (var locker = @lock.ForReading())
+			{
+				var methodOnTarget = GetFromCache(proxiedMethod, type);
+				if (methodOnTarget != null)
+				{
+					return methodOnTarget;
+				}
+			}
+
 			using (var locker = @lock.ForReadingUpgradeable())
 			{
 				var methodOnTarget = GetFromCache(proxiedMethod, type);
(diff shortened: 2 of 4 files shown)


stakx:
Your PR looks good to me. Sorry for initially missing the point of your enquiry.

The following is just an additional idea; feel free to disregard. (I'm not a multi-threading expert at all, so perhaps this is naïve.)

It seems that the locks are mostly used in combination with `Dictionary<,>`. I was wondering whether it might make sense to push the locking that's currently strewn all over the place down into a new dictionary-like type, say:

```csharp
// using System.Collections.Generic;
// using System.Threading;

internal sealed class SynchronizedDictionary<TKey, TValue> : IDisposable
{
    private Dictionary<TKey, TValue> dictionary;
    private ReaderWriterLockSlim readerWriterLock;

    // constructor and Dispose omitted for brevity

    public TValue GetOrAdd(TKey key, Func<TKey, TValue> valueFactory)
    {
        TValue value;

        this.readerWriterLock.EnterReadLock();
        try
        {
            if (this.dictionary.TryGetValue(key, out value))
            {
                return value;
            }
        }
        finally
        {
            this.readerWriterLock.ExitReadLock();
        }

        this.readerWriterLock.EnterUpgradeableReadLock();
        try
        {
            if (this.dictionary.TryGetValue(key, out value))
            {
                return value;
            }

            this.readerWriterLock.EnterWriteLock();
            try
            {
                value = valueFactory.Invoke(key);
                this.dictionary.Add(key, value);
                return value;
            }
            finally
            {
                this.readerWriterLock.ExitWriteLock();
            }
        }
        finally
        {
            this.readerWriterLock.ExitUpgradeableReadLock();
        }
    }
}
```

This could have two advantages:

1. Raise the level of abstraction in the remaining code, which just needs a `SynchronizedDictionary<,>` instead of a combination of dictionary plus lock that must be remembered to always be used together.
2. Reducing overhead by using .NET's `ReaderWriterLockSlim` directly instead of Core's `Lock` abstraction that's built on top of it.

You'll notice that the `GetOrAdd` API is identical to that in the BCL's `ConcurrentDictionary<,>`. That however doesn't mean that `ConcurrentDictionary<,>` would be a suitable alternative (besides the fact that it isn't available for .NET 3.5): Using a completely lock-free collection might be the wrong thing to do in `BaseProxyGenerator.ObtainProxyType`&mdash;there should be no concurrent Reflection.Emit-ting to the same dynamic module, since Reflection.Emit's official thread-safety guarantees are very weak.

src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
```diff
@@ -415,9 +415,10 @@ protected Type ObtainProxyType(CacheKey cacheKey, Func<string, INamingScope, Typ
 				var proxyType = factory.Invoke(name, Scope.NamingScope.SafeSubScope());
 
 				// Upgrade the lock to a write lock. 
-				locker.Upgrade();
-
-				AddToCache(cacheKey, proxyType);
+				using (locker.Upgrade())
```
jonorossi: For those reading at home, `SlimReadWriteLock.ForReadingUpgradeable` returns a `SlimUpgradeableReadLockHolder`, `SlimUpgradeableReadLockHolder.Upgrade` creates and holds on to a `SlimWriteLockHolder`. If you don't dispose the writer lock holder the upgradeable read lock holder will do it for you (resulting in `ReaderWriterLockSlim.ExitWriteLock` being called), so this change semantically has no change but makes sense because it is more explicit to scope the write lock's lifetime.

src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
```diff
@@ -48,22 +48,29 @@ public static MethodInfo GetMethodOnType(Type type, MethodInfo proxiedMethod)
 
 			Debug.Assert(proxiedMethod.DeclaringType.IsAssignableFrom(type),
 						 "proxiedMethod.DeclaringType.IsAssignableFrom(type)");
-			using (var locker = @lock.ForReadingUpgradeable())
+			using (var locker = @lock.ForReading())
```
jonorossi: This is a great change, as you suspected this has got to be what the author of the code wanted. ProxyUtil was probably copy-and-pasted from this code and inherited the same flaw.

PR #383, opened 2018-06-18
Synchronize read/write accesses to `NamingScope.GetUniqueName`

stakx:
`NamingScope.GetUniqueName` maintains and updates counters in an unsynchronized `Dictionary<,>`. This writing method is currently called from a location that allows several concurrent readers.

This changes `BaseProxyGenerator.ObtainProxyType` so that it goes from a read lock straight to a write lock. There seems little point in taking an upgradeable lock first if we're going to need a write lock anyway for all of the work. (Taking an upgradeable read lock first would only benefit the second, repeated cache check.)

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
@@ -396,9 +396,9 @@ namespace Castle.DynamicProxy.Generators
 			}
 
 			// This is to avoid generating duplicate types under heavy multithreaded load.
-			using (var locker = Scope.Lock.ForReadingUpgradeable())
+			using (var locker = Scope.Lock.ForWriting())
 			{
-				// Only one thread at a time may enter an upgradable read lock.
+				// Only one thread at a time may enter a write lock.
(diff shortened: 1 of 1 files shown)


src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
```diff
@@ -396,9 +396,9 @@ protected Type ObtainProxyType(CacheKey cacheKey, Func<string, INamingScope, Typ
 			}
 
 			// This is to avoid generating duplicate types under heavy multithreaded load.
```
TimLovellSmith: nit, you could probably move the comment 'This is to avoid generating duplicate types under heavy multithreaded load'  to just after 'See if an earlier lock holder populated the cache.'

TimLovellSmith:
@stakx Looks beautiful to me!

src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
stakx: Actually, I think the comment `// See if an earlier lock holder populated the cache.` should be moved up to the first cache check.

Issue #376, opened 2018-06-13
Can Lock abstraction be simplified by removing waitForLock overload?


TimLovellSmith:
I've been reading 'Lock' and its lone implementation 'SlimReadWriteLock'.

There is something scary to me about the idea of a Lock class API with operations like 'ForReading', 'ForWriting', 'ForReadingUpgradable' where it will return you a *holder* of the lock, which I think implies that you own the lock and will thereby release it by Disposing the holder, but that only *optionally* waits to acquire the lock first (since it has an overload with a (bool waitForLock) signature.

I guess the intention here was that we're going to be able to use this Lock class to write some very optimal code where we don't wait for the lock if we know we already acquired it, or something clever like that right?

Well you may or may not be surprised to learn I could find no such extremely clever code anywhere in Castle.Core... in fact it seems that waitForLock is ALWAYS true. So I was scared for nothing.

So really this issue is just to point out this seems like a great opportunity for some YAGNI-aligned code simplification...

jonorossi:
> ... its lone implementation 'SlimReadWriteLock'.

Today we only have a single implementation, .NET 2.0 and Silverlight don't have `ReaderWriterLockSlim` so this class hid another implementation using `Monitor` on .NET 2.0 and Silverlight when we used to support those runtimes.

> ... but that only optionally waits to acquire the lock first (since it has an overload with a (bool waitForLock) signature.
>
> I guess the intention here was that we're going to be able to use this Lock class to write some very optimal code where we don't wait for the lock if we know we already acquired it, or something clever like that right?
>
> Well you may or may not be surprised to learn I could find no such extremely clever code anywhere in Castle.Core... in fact it seems that waitForLock is ALWAYS true. So I was scared for nothing.

With a quick search on GitHub I see one use of `ForReading(false)` in Windsor, but it doesn't look like a great use case. Yes, it looks like this might have been a little YAGNI. I didn't check any of the other old projects like MonoRail or ActiveRecord.

> So really this issue is just to point out this seems like a great opportunity for some YAGNI-aligned code simplification...

We could probably make the API clearer by changing the `ForReading` overload that takes a `waitForLock` parameter, and add a `TryForReading` or something method. However this would be a breaking change.

stakx:
> Today we only have a single implementation, .NET 2.0 and Silverlight don't have `ReaderWriterLockSlim` so this class hid another implementation using `Monitor` on .NET 2.0 and Silverlight when we used to support those runtimes.

[...]

PR #391, opened 2018-06-21
Deprecate `Lock` and `ModuleScope`'s public type cache API

stakx:
This is in response to https://github.com/castleproject/Core/issues/376#issuecomment-398688610:

> [We] should definitely consider changing the code to use [`ReaderWriterLockSlim`] directly as it'll also remove all the allocations of the "disposable holder" objects.

This PR deprecates `Lock` and directly related types. This is achieved by replacing internal uses either with `ReaderWriterLockSlim` (upon which it is based), or with a new but simpler abstraction, `SynchronizedDictionary<,>`.

This PR also deprecates `ModuleScope`'s public API for accessing the proxy type cache. This is necessary because that API consists of a `Lock` (which we're deprecating) along with a few thread-unsafe methods.

Publicly visible members marked for deprecation:

* `Castle.Core.Internal.Lock` (class)
* `Castle.Core.Internal.ILockHolder` (interface)
* `Castle.Core.Internal.IUpgradeableLockHolder` (interface)
* `Castle.DynamicProxy.Generators.BaseProxyGenerator.AddToCache` (method)
* `Castle.DynamicProxy.Generators.BaseProxyGenerator.GetFromCache` (method)
* `Castle.DynamicProxy.Generators.CacheKey` (class)
* `Castle.DynamicProxy.ModuleScope.Lock` (property)
* `Castle.DynamicProxy.ModuleScope.GetFromCache` (method)
* `Castle.DynamicProxy.ModuleScope.RegisterInCache` (method)
* `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
* `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)


fir3pho3nixx:
We need a test harness that batters the absolute hell out of this. I think something that I can track memory usage and call site hotspots. 

I would like to run some metrics on before and after changes here.

stakx:
@fir3pho3nixx - One small note regarding resource usage: I noticed that the `Lock`-related APIs never `Dispose` of anything. See e.g. `SingletonDispenser`'s use of wait handles, they are allocated but apparently never released. That doesn't seem right to me. (The fact that `ReaderWriterLockSlim` in DynamicProxy are never disposed is fine since they are `static`, i.e. they live as long as the process / `AppDomain`.)

fir3pho3nixx:
> See e.g. SingletonDispenser's use of wait handles, they are allocated but apparently never released

OK, this is really weird. 

https://github.com/stakx/Castle.Core/blob/lock/src/Castle.Core/Components.DictionaryAdapter/Xml/Internal/Utilities/SingletonDispenser.cs#L53

`TryGetExistingItem` default allocates a manual reset event

https://github.com/stakx/Castle.Core/blob/lock/src/Castle.Core/Components.DictionaryAdapter/Xml/Internal/Utilities/SingletonDispenser.cs#L87

`WaitForCreate` will `WaitOne`

https://github.com/stakx/Castle.Core/blob/lock/src/Castle.Core/Components.DictionaryAdapter/Xml/Internal/Utilities/SingletonDispenser.cs#L104

`Create` does the work of Set. 

This is bonkers, the thing dishing out the reset events is not disposable or ever disposing them.

Anything we can learn from [here](http://kozmic.net/2011/01/05/lock-free-thread-safe-initialisation-in-c/)? I have not look at this in infinite detail.

fir3pho3nixx:
> Anything we can learn from [here](http://kozmic.net/2011/01/05/lock-free-thread-safe-initialisation-in-c/)? I have not looked at this in infinite detail.

I am a fan of atomic locking(Interlocked) not sure how this would apply to the calling code.

stakx:
@fir3pho3nixx - I think the dangling wait handle issue could be resolved with `Lazy<T>(factory, LazyThreadSafetyMode.PublicationOnly)`. Unfortunately we don't have that available for .NET 3.5. We could implement a trimmed down `Lazy<T>` shim class that only implements that one particular thread-safety mode (which internally uses `Interlocked.CompareExchange` instead of `Monitor`, I believe). CoreFX could provide a starting point, if licensing allows us to reuse their implementation.

That said, I'd like to keep this PR focused solely on obsolescing `Lock` (and the baggage that comes with this). Do you think it fair enough to improve upon `SingletonDispenser` in a separate PR?

src/Castle.Core/Components.DictionaryAdapter/Xml/Internal/Utilities/SingletonDispenser.cs
```diff
@@ -18,12 +18,11 @@ namespace Castle.Components.DictionaryAdapter.Xml
 	using System;
 	using System.Collections.Generic;
 	using System.Threading;
-	using Castle.Core.Internal;
 
 	public class SingletonDispenser<TKey, TItem>
```
TimLovellSmith: If this is an internal utility why is it public? Should it be [Obsolete] too?

fir3pho3nixx:
@stakx 

> I think the dangling wait handle issue could be resolved with Lazy<TItem>(factory, LazyThreadSafetyMode.PublicationOnly). (I believe that thread-safety mode uses Interlocked instead of Monitor, 
 
Correct. https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Lazy.cs#L356

> the downside is that factory might be invoked more than once

Not sure about this. Please see: https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization#thread-safe-initialization

`Specifying LazyThreadSafetyMode.PublicationOnly allows multiple threads to attempt to initialize the Lazy<T> instance. Only one thread can win this race, and all the other threads receive the value that was initialized by the successful thread`

> We could implement a trimmed down Lazy<> shim class. CoreFX could provide a starting point, if licensing allows us to reuse their implementation

Not sure we can do this but nothing is stopping us from using the framework :)

> That said, I'd like to keep this PR focused solely on obsolescing Lock (and the baggage that comes with this). Do you think it fair enough if we improved SingletonDispenser in a separate PR?

Yep no worries! Just trying to validate your concerns you raised here: https://github.com/castleproject/Core/pull/391#issuecomment-398946120

src/Castle.Core/Components.DictionaryAdapter/Xml/Internal/Utilities/SingletonDispenser.cs
fir3pho3nixx: Or it should be `internal`, like as in the csharp modifier we all know and love :)

fir3pho3nixx:
So ... 

What are we doing about that test harness then? 

I could run dotmem and dottrace over it.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -174,11 +189,94 @@ public Type GetFromCache(CacheKey key)
...
+			}
+			finally
+			{
+				this.typeCacheLock.ExitReadLock();
+			}
+
+			this.typeCacheLock.EnterWriteLock();
+			try
+			{
+				if (this.typeCache.TryGetValue(key, out value))
```
jonorossi: Can the guts here just call `GetFromOrRegisterInCache`?

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -174,11 +189,94 @@ public Type GetFromCache(CacheKey key)
...
+		///   a write lock. The function must not access this scope's type cache.
+		/// </remarks>
+		public Type GetFromOrRegisterInCache(CacheKey key, Func<CacheKey, Type> valueFactory)
+		{
+			Type value;
+
+			Debug.Assert(key != null);
+			Debug.Assert(valueFactory != null);
+
+			if (this.typeCache.TryGetValue(key, out value))
```
jonorossi: No need to prefix all field access with `this.`.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -174,11 +189,94 @@ public Type GetFromCache(CacheKey key)
...
+		/// </summary>
+		/// <param name="key">The key to be looked up in the cache.</param>
+		/// <param name="valueFactory">A function producing the type to be registered in the cache.</param>
+		/// <returns>The type from this scope's type cache matching the key, or null if the key cannot be found</returns>
+		/// <remarks>
+		///   This method synchronizes accesses to this scope's type cache using read/write locks.
+		///   If the specified key cannot be found, the factory function is invoked while a write lock is held.
+		///   The function must not access this scope's type cache.
+		/// </remarks>
+		public Type GetFromOrRegisterInCacheSynchronized(CacheKey key, Func<CacheKey, Type> valueFactory)
```
jonorossi: I would have thought this method would be named `GetFromOrRegisterInCache` as the go to one (like how `AddCacheMappings` has built-in synchronisation), and the previous one would be suffixed to indicate it is the special one, e.g. `GetFromOrRegisterInCacheHoldingLock` or something.

On a separate note how do users use the first one without synchronisation without using the deprecated `Lock` property?

Maybe `GetOrAddToCache` would be a simpler name.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -174,11 +189,94 @@ public Type GetFromCache(CacheKey key)
...
+		/// </summary>
+		/// <param name="key">The key to be looked up in the cache.</param>
+		/// <param name="valueFactory">A function producing the type to be registered in the cache.</param>
+		/// <returns>The type from this scope's type cache matching the key, or null if the key cannot be found</returns>
+		/// <remarks>
+		///   This method does not synchronizes accesses to this scope's type cache using read/write locks.
+		///   Only use this method when you know for sure that a method further up in the call stack already holds
+		///   a write lock. The function must not access this scope's type cache.
+		/// </remarks>
+		public Type GetFromOrRegisterInCache(CacheKey key, Func<CacheKey, Type> valueFactory)
```
jonorossi: Since methods like `RegisterInCache` are being deprecated, are you making this public for a good reason?

I don't see why we'd want to allow users to add their own types to the cache. We've already got `LoadAssemblyIntoCache` for loading a DP generated assembly which is a special case.

src/Castle.Core/DynamicProxy/ModuleScope.cs
stakx: Sorry, that's a (bad?) habit I picked up from a recent coworker. 😄 Will remove.

src/Castle.Core/DynamicProxy/ModuleScope.cs
stakx: @jonorossi:

> On a separate note how do users use the first one without synchronisation without using the deprecated `Lock` property?

If they need locking, they use the other, synchronized method. If they don't need locking, then they don't need `Lock` in the first place.

The question here is probably, who are the users you're thinking of? Why should anyone even try to update the proxy type cache? Why is the proxy type cache, along with `CacheKey`, even exposed publicly?

> Maybe `GetOrAddToCache` would be a simpler name.

I tried to stay close to what was there before the changes, but I agree.

src/Castle.Core/DynamicProxy/ModuleScope.cs
stakx: > I don't see why we'd want to allow users to add their own types to the cache.

Neither do I.

> Since methods like `RegisterInCache` are being deprecated, are you making this public for a good reason?

If we agree that the proxy type cache shouldn't be exposed, I'd be more than happy to make this `internal`. I just didn't know the full story behind *why* the cache is exposed, so I didn't want to just reverse it without further discussion.

jonorossi: > If we agree that the proxy type cache shouldn't be exposed, I'd be more than happy to make this internal. I just didn't know the full story behind why the cache is exposed, so I didn't want to just reverse it without further discussion.

The old methods will be gone in the next major version, so we'll hear from someone if they think they have a valid use case (doubt it), but let's not make new stuff public unless we actually want to.

src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
```diff
@@ -179,19 +179,10 @@ private Type GetDelegateType(MetaMethod method, ClassEmitter @class, ProxyGenera
...
 				null);
 
-			var type = scope.GetFromCache(key);
-			if (type != null)
-			{
-				return type;
-			}
-
-			type = new DelegateTypeGenerator(method, targetType)
+			return scope.GetOrAddToCacheWithoutTakingLock(key, _ =>
```
TimLovellSmith: I see all these calls to WithoutTakingLock and I wonder - how are you reasoning about whether you hold the lock already or not?

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+   - `Castle.Core.Internal.Lock` (class)
+   - `Castle.Core.Internal.ILockHolder` (interface)
+   - `Castle.Core.Internal.IUpgradeableLockHolder` (interface)
+- The proxy type cache in `ModuleScope` should no longer be accessed directly. For this reason, the members listed below have been deprecated. (@stakx, #391)
+   - `Castle.DynamicProxy.ModuleScope.Lock` (property)
+   - `Castle.DynamicProxy.ModuleScope.GetFromCache` (method)
+   - `Castle.DynamicProxy.ModuleScope.RegisterInCache` (method)
+   - `Castle.DynamicProxy.Generators.BaseProxyGenerator.AddToCache` (method)
+   - `Castle.DynamicProxy.Generators.BaseProxyGenerator.GetFromCache` (method)
+
```
TimLovellSmith: Lets deprecate CacheKey too, since you're not supposed to use the cache directly!

src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
stakx: I don't. Someone else did that before me, I only refactored pre-existing code.

The general flow of execution appears to be from `IProxyGenerator` into `IProxyBuilder` into "generator" classes which, via their base class `BaseProxyGenerator`, take hold of a lock, then they use contributors to build a model of the proxy type, then they use "emitters" to actually generate code. Most of the calls you're seeing are in the contributors so if all goes well, the lock should be taken at that point.

CHANGELOG.md
stakx: I'm already on it. :wink:

src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
TimLovellSmith: Maybe we should have 'WithoutTakingLock' also do some kind of assertion that the lock is really held by the current thread then? (in debug mode)

Also a different idea - if you wanted the compiler to help analyze whether these assumptions are really correct, you can do an experiment - change 'WithoutTakingLock' to require an extra parameter 'ProofLockHeld', and see what build errors you get. Then fix them (by making callers pass in the parameter, which they must get from their own caller), and repeat - until you find yourself at an ultimate caller where you are taking the lock and acquiring 'ProofLockHeld' by doing so). Might be a painfully slow experiment though! 

And of course it would break all the apis so you can't really ship it this way, or keep it around to prove that nothing is broken ever, so it doesn't have much lasting value ;(

stakx: > Maybe we should have 'WithoutTakingLock' also do some kind of assertion that the lock is really held by the current thread then?

Perhaps, but I see the present PR purely as a refactoring to get rid of some no-longer-needed types and abstractions. I'd like to keep it focused on that.

Changes helping us resolve #193 don't belong here, in my opinion.

> Also a different idea [...]

It's a good idea as a one-time manual analysis, I might give this a try in a spare moment & separate branch! Thanks. :+1:

buildscripts/common.props
```diff
@@ -2,6 +2,7 @@
 	
 	<PropertyGroup>
 		<NoWarn>$(NoWarn);CS1591;CS3014;CS3003;CS3001;CS3021</NoWarn>
+		<NoWarn>$(NoWarn);CS0612;CS0618</NoWarn> <!-- TODO: Remove this line once `[Obsolete]` members have been dealt with. -->
```
stakx: I am not a fan of disabling `[Obsolete]`-related warnings CS0612 and CS0618 solution-wide, but this still seems better than sprinkling `#pragma warning disable`s and `#pragma warning restore`s across the whole code base. `[assembly: SuppressMessage]` would have been nice, but I couldn't get it to work with these warnings.

We should remove this again right before releasing Castle Core 5.0.0.

src/Castle.Core/Core/Internal/ILockHolder.cs
```diff
@@ -15,7 +15,10 @@
 namespace Castle.Core.Internal
 {
 	using System;
+	using System.ComponentModel;
 
+	[Obsolete("Consider using `System.Threading.ReaderWriterLockSlim` instead of `Lock` and related types.")] // TODO: Remove this type.
+	[EditorBrowsable(EditorBrowsableState.Never)]
```
stakx: I've put `// TODO` comments next to the most `[Obsolete]` attributes. They contain instructions how to deal with each obsolete member, so that the clean-up will be as simple as possible.

src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
```diff
@@ -18,16 +18,17 @@ namespace Castle.DynamicProxy.Internal
...
 	using Castle.DynamicProxy.Generators;
 
 	public static class InvocationHelper
 	{
 		private static readonly Dictionary<CacheKey, MethodInfo> cache =
 			new Dictionary<CacheKey, MethodInfo>();
 
-		private static readonly Lock @lock = Lock.Create();
+		private static readonly ReaderWriterLockSlim cacheLock =
+			new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
```
stakx: You often see a `Dictionary<,>` together with a `ReaderWriterLockSlim`. These could be wrapped up as an internal helper type similar to the BCL's `ConcurrentDictionary<>`, except we might only need a `GetOrAdd` method. Having such a type would mean we can replace all of the double-checked lock pattern that's currently being used in several places with a single `GetOrAdd` call. (This should work in `ProxyUtil`, `InvocationHelper`, `DictionaryAdapterFactory`; but probably not in `ModuleScope` nor `SingletonDispenser` due to the added synchronization requirements these two types have).

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -565,7 +657,7 @@ public void LoadAssemblyIntoCache(Assembly assembly)
 
 				if (loadedType != null)
 				{
-					RegisterInCache(mapping.Key, loadedType);
+					typeCache[mapping.Key] = loadedType;
```
stakx: Note the absence of any locking here. Proper locking should be added, I didn't do this yet because I wanted this PR to be mostly a refactoring.

src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterFactory.cs
```diff
@@ -132,21 +132,32 @@ public DictionaryAdapterMeta GetAdapterMeta(Type type, DictionaryAdapterMeta oth
...
 					return meta;
+				}
+			}
+			finally
+			{
+				interfaceToMetaLock.ExitReadLock();
 			}
 
-			using (var heldLock = interfaceToMetaLock.ForReadingUpgradeable())
+			interfaceToMetaLock.EnterWriteLock();
```
fir3pho3nixx: @stakx Just to confirm, we are happy with the performance hit here?

stakx: @fir3pho3nixx - You know what, I think you are right. There is no real need to make this functional change now. Why "optimize" something when (a) I haven't established that this really *is* an optimization, and (b) when this is supposed to be a refactoring-only PR anyway?

I've reinstated the upgradeable read locks. After this PR, it should be easy to change things because the locking pattern that DynamicProxy will be present in only two locations (`SingletonDispenser` and `SynchronizedDictionary<,>`).

Issue #390, opened 2018-06-21
Activator.CreateInstance(generatedType) in ProxyGenerator may be thread unsafe


TimLovellSmith:
Per #193 creating a new issue to track a hypothetical bug in ProxyGenerator calling Activator.CreateInstance without holding the lock that guards against races with Reflection.Emit.

>>Are you saying that the Activator.CreateInstance(generatedType) in ProxyGenerator must execute while a read lock is held?

>@stakx that might possibly be enough to fix it for some of the stacks we've seen, like this one:

```
at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithoutTarget(Type interfaceToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, IInterceptor[] interceptors)
at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithoutTarget(Type interfaceToProxy, IInterceptor interceptor)
at Castle.DynamicProxy.ProxyGenerator.CreateInterfaceProxyWithoutTarget[TInterface](IInterceptor interceptor)
at NetCoreRepro.ProxyFactory.CreateProxyT in C:\work\CastleCore193Repro\NetCore\NetCoreRepro\ProxyFactory.cs:line 13
at NetCoreRepro.ToDuplicate.Test697.DoSomething() in C:\work\CastleCore193Repro\NetCore\NetCoreRepro\Generated\Test697.cs:line 12
```

>but I do not feel hopeful that it is going to fix all the other cases, because from the stack it seems like maybe they are triggered by JIT, rather than type loading or instantiation.

I'll create an issue specifically around `Activator.CreateInstance(generatedType)` and we can try out what fixing it does.

stakx:
I should've added an explanation in #193 *why* I find it unlikely that this will be the solution (but by all means, do give this a try and report your findings! :-)

1. I have a branch in my clone of @BrunoJuchli's `NetCore/` repro where I've replaced DynamicProxy with plain Reflection.Emit code. Obviously I also need an `Activator.CreateInstance` call in there. The thing is, I've never been able to reproduce any errors once DynamicProxy is gone. So `Activator` doesn't appear to be the culprit or I'd still have seen the error.

[...]

PR #392, opened 2018-06-21
Theorized fix for theorized bug #390 option 1 - taking a Read lock 

TimLovellSmith:
while calling Activator.CreateInstance.

Don't accept this unless it works! I haven't figured out how to test this yet. 😊

### DIFF
--- a/src/Castle.Core/DynamicProxy/ProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/ProxyGenerator.cs
@@ -338,7 +338,10 @@ namespace Castle.DynamicProxy
 			                                                       options);
 
 			var arguments = GetConstructorArguments(target, interceptors, options);
-			return Activator.CreateInstance(generatedType, arguments.ToArray());
+			using (proxyBuilder.ModuleScope.Lock.ForReading())
+			{
+				return Activator.CreateInstance(generatedType, arguments.ToArray());
+			}
 		}
 
 		protected List<object> GetConstructorArguments(object target, IInterceptor[] interceptors,
@@ -624,7 +627,10 @@ namespace Castle.DynamicProxy
 				Debug.Assert(constructors.Length == 1, "constructors.Length == 1");
 				return constructors[0].Invoke(arguments.ToArray());
 			}
-			return Activator.CreateInstance(generatedType, arguments.ToArray());
+			using (proxyBuilder.ModuleScope.Lock.ForReading())
+			{
+				return Activator.CreateInstance(generatedType, arguments.ToArray());
+			}
 		}
 
 		/// <summary>
@@ -866,7 +872,10 @@ namespace Castle.DynamicProxy
 
 			var generatedType = CreateInterfaceProxyTypeWithoutTarget(interfaceToProxy, additionalInterfacesToProxy, options);
 			var arguments = GetConstructorArguments(null, interceptors, options);
-			return Activator.CreateInstance(generatedType, arguments.ToArray());
+			using (proxyBuilder.ModuleScope.Lock.ForReading())
+			{
+				return Activator.CreateInstance(generatedType, arguments.ToArray());
+			}
 		}
 
 		/// <summary>
@@ -1453,7 +1462,10 @@ namespace Castle.DynamicProxy
 		{
 			try
 			{
-				return Activator.CreateInstance(proxyType, proxyArguments.ToArray());
+				using (proxyBuilder.ModuleScope.Lock.ForReading())
+				{
+					return Activator.CreateInstance(proxyType, proxyArguments.ToArray());
+				}
 			}
 			catch (MissingMethodException)
 			{


stakx:
May I ask why a PR for this? You should be able to test locally; a PR won't really do you much good at this stage.

>  I haven't figured out how to test this yet. 😊

You'll need the following prerequisites:

* a computer with several physical cores
* Visual Studio JIT Debugging should be enabled / installed

Then proceed roughly as follows:

1. Clone e.g. [BrunoJuchli/Core](https://github.com/BrunoJuchli/Core) and check out the `193` branch.
2. Make your changes to DynamicProxy in the `src/Castle.Core` project.
3. Compile and publish the `src/NetCoreRepro` project to `src\NetCoreRepro\bin\Debug\netcoreapp2.1\win10-x64`.
4. Open several command line prompts (say, at least 4) and run `_193.bat` in each of them. This will execute the published project in a loop.
5. Let it run up to several hours until an error happens.

TimLovellSmith:
So far:
I did a very short run with the added write lock (not the read lock of this PR) and couldn't repro any crash yet.
I did another very short run with no locking and couldn't repro any of the exceptions from #193 per se, but did run into one app crash:

```
Application: NetCoreRepro.exe CoreCLR Version: 4.6.26515.7 Description: The  process was terminated due to an internal error in the .NET Runtime at IP  00007FFEDB9F9B04 (00007FFEDB650000) with exit code 80131506
```

According to StackOverflow answers this indicates an ExecutionEngineException. Quote

>The generic cause is corruption of the state of the garbage collected heap. Which in turn is invariably caused by unmanaged code

[or]

>bugs in concurrent garbage collection...

fir3pho3nixx:
@TimLovellSmith 

A long time ago when I used to muck around with methods like [ICorProfilerInfo::GetILFunctionBody](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo-getilfunctionbody-method) and [ICorProfilerInfo::SetILFunctionBody](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo-setilfunctionbody-method) I used to get this exception all the time. It was almost always because I cocked up the IL.

stakx:
@TimLovellSmith - I can confirm that putting a read lock around `Activator.CreateInstance` makes the error either disappear, or a lot less likely to happen&mdash;see https://github.com/castleproject/Core/issues/193#issuecomment-400206187.

However, that doesn't mean that putting a lock around `Activator.CreateInstance` is definitely the right thing to do. The error might just become a lot less likely. I stopped my test run after 8 hours (the error usually occurs after approx. 2 1/2 hours on my machine), but it might be necessary to keep the test running for several days (as in e.g. https://github.com/castleproject/Core/issues/193#issuecomment-399839323).

I suggest we keep this PR open for now, until at least next week, when we'll have test results from another test run by @BrunoJuchli.

TimLovellSmith:
Related to https://github.com/castleproject/Core/issues/193#issuecomment-400496224 maybe it is type instantiation, whether via 'new' or 'CreateInstance' which causes all the problems, and so if we just add one more Read lock around newing up the invocation objects, all bugs will be solved!

stakx:
@TimLovellSmith - no. If type instantiation is not thread-safe, then Microsoft needs to fix the runtime.

TimLovellSmith:
@stakx Well I don't disagree, but it seemed like a pretty fun experiment!

TimLovellSmith:
@stakx @BrunoJuchli Well, actually other reasons I mentioned it were 
- it seems to be a good unifying theory behind why we get BadImageFormatException during method invocations
- it might have something to do with why we are having trouble reproing the issue without DynamicProxy, but I am still not sure exactly what.

BrunoJuchli:
@TimLovellSmith 
I agree that if putting a read lock around all instantiations makes the issue vanish it's an indicator that the framework is not performing as it should (concurrency issue with instantiation & emitting).
I'm currently running a test like that.

The general problem with concurrency issues is that generally even a small, seemingly unrelated, changes to the code can have a big impact on the frequency with which errors occur.
So even if it ran for 2 weeks without producing a single error that would not be a guarantee that the change actually fixes the issue.
Here, proof can only work the other way round: If you run the code and it produces an error - it's guaranteed to be buggy.

Still, I'd say if the change means going from an error every 2.5hours to no error in 2 weeks, then I hope this would at least be enough for MS to have a close look.

stakx:
@TimLovellSmith, @BrunoJuchli: This PR is about to go to sleep, so before noone remembers anymore what it was about, I'd like to ask how we should proceed with it.

My personal vote is for closing this PR unmerged. I agree with @BrunoJuchli that (a) it's hard to prove that additional locks actually make the error go away completely vs. just making it much more improbable.; and that (b) the likeliest explanation for the threading issue we've been hunting is a runtime bug. And finally (c), these locks will negatively affect performance of Castle.Core, while providing benefits only in some very marginal scenarios. For these two reasons, I feel we shouldn't do this.

Do you have any objections if I close this PR?

stakx:
Just when I thought you were simply otherwise busy, you're telling me the real reason. 😆 What a crazy hot summer it is, indeed. 🌞

jonorossi:
> My personal vote is for closing this PR unmerged. I agree with @BrunoJuchli that (a) it's hard to prove that additional locks actually make the error go away completely vs. just making it much more improbable; and that (b) the likeliest explanation for the threading issue we've been hunting is a runtime bug (meaning we're not the ones who should "fix" things). And finally (c), these locks will negatively affect performance of DynamicProxy, while providing benefits only in some very marginal scenarios. For these reasons, I feel we shouldn't make this change.
> 
> Do any of you have objections against closing this PR?

I agree, this pull request should be closed unmerged. The performance implications of this change are too severe. Microsoft have already stated that they expect `Activator.CreateInstance` with dynamic types should not have any threading constraints, any required locking on internal runtime structures for mutations should be performed by the runtime.

PR #393, opened 2018-06-21
Theorized fix for theorized bug #390 option 2 - taking a Write lock 

TimLovellSmith:
while calling Activator.CreateInstance.

Don't accept this unless it works - I haven't figured out how to test this yet! 😊


fir3pho3nixx:
Please see: https://github.com/castleproject/Core/pull/392#issuecomment-399176721

stakx:
@TimLovellSmith - Can we close this PR? There's not much point in a stronger lock when a weaker lock already appears to accomplish what it's supposed to. (See #392, where I confirm that holding a read lock while `Activator.CreateInstance` is being called indeed appears to make a difference.)

TimLovellSmith:
@stakx Sure, read lock sounds good for now.

PR #395, opened 2018-06-22
Remove orphaned backing field from `ProxyGenerationOptions`

stakx:
Attention, baby PR ahead. 👶🍼

I noticed an unused field in `ProxyGenerationOptions` that should have been removed [in `692d82f`](https://github.com/castleproject/Core/commit/692d82fb917d8568da56601db6d45a891babe090#diff-ad3089af8109829f561062fe11a12d92), along with the property to which it originally belonged.

(...or was this intentionally kept around for serialization backwards compatibility?)


jonorossi:
> or was this intentionally kept around for serialization backwards compatibility?

Doubt it, `GetObjectData` doesn't use it.

PR #397, opened 2018-06-24
Remove redundant call to EmitCallEnsureValidTarget

### LINKED ISSUE #295: Duplicate invocation of EnsureValidTarget method
I've inspected the generated invocation class using Reflector and found 
that `EnsureValidTarget` method is invoked twice:

```c#
public override void InvokeMethodOnTarget()
{
      base.EnsureValidTarget(); // #1
      base.EnsureValidTarget(); // #2
      string str = (base.target as IEventHandler).Handle((EventArgs) base.GetArgumentValue(0));
      base.ReturnValue = str;
}
```

These method calls are generated here:
* [InvocationTypeGenerator, line 120](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs#L120-L123) — for proxies that can change target
* [CompositionInvocationTypeGenerator, line 63](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/Generators/CompositionInvocationTypeGenerator.cs#L63-L65) — for all proxies

It's not really a big deal, but it seems quite easy to avoid.
Perhaps by having a flag, something like `ensureValidTargetEmitted`.

stakx:
This fixes #295. See https://github.com/castleproject/Core/issues/295#issuecomment-399751655 for an explanation of the proposed change. (I'd say this doesn't deserve a changelog entry, it's basically just a small optimization.)

### DIFF
--- a/src/Castle.Core/DynamicProxy/Tokens/InvocationMethods.cs
+++ b/src/Castle.Core/DynamicProxy/Tokens/InvocationMethods.cs
@@ -36,7 +36,7 @@ namespace Castle.DynamicProxy.Tokens
 			                                             },
 			                                             null);
 
-		public static readonly MethodInfo EnsureValidTarget =
+		public static readonly MethodInfo CompositionInvocationEnsureValidTarget =
 			typeof(CompositionInvocation).GetMethod("EnsureValidTarget", BindingFlags.Instance | BindingFlags.NonPublic);
 
 		public static readonly MethodInfo GetArgumentValue =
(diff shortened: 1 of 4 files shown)


src/Castle.Core/DynamicProxy/Tokens/InvocationMethods.cs
```diff
@@ -89,7 +89,7 @@ public static class InvocationMethods
 		public static readonly MethodInfo SetReturnValue =
 			typeof(AbstractInvocation).GetMethod("set_ReturnValue");
 
-		public static readonly FieldInfo Target =
+		public static readonly FieldInfo CompositionInvocationTarget =
```
jonorossi: Just realised these fields are public, so technically breaking. I'm thinking if anyone is using them directly rather than via one of the supported interfaces (e.g. `IProxyTargetAccessor`) then bad luck to them, and we should make the new ones internal.

stakx: True! 🤔 Since we're already about to break something, how about we deprecate *all* those `…Methods` types in the `Tokens` folder? I could submit a separate PR for that to go before this one.

jonorossi: What about just keeping back compat like this then:

```csharp
public static readonly FieldInfo CompositionInvocationTarget = typeof...;
public static readonly FieldInfo Target = CompositionInvocationTarget;
```

And we worry about deprecating them with the other task.

stakx: Done.

jonorossi: > Shall I still submit a PR just for deprecating Tokens/, or would you prefer a larger PR that deprecates a lot of things all at once?

Let's just leave it to the full API review, one big release with all the deprecations should be easier to manage.

PR #401, opened 2018-06-26
Make AST construction more efficient

stakx:
**More to come. Please don't merge yet.**

The few commits in this PR make the construction of ASTs (which represent the code in intercepted method bodies) more efficient by getting rid of unnecessary object instantiations.

Two AST node types, `ConstReference` and `ExpressionStatement`, are rendered superfluous and therefore marked as obsolete.

There's almost certainly potential for further optimizations, those covered here are probably just the easiest ones to find.

See commit messages for further details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ Deprecations:
    - `Castle.DynamicProxy.Generators.CacheKey` (class)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
+- The following AST node types have been deprecated:
+   - `Castle.DynamicProxy.Generators.Emitters.SimpleAST.ConstReference` (class)
+   - `Castle.DynamicProxy.Generators.Emitters.SimpleAST.ExpressionStatement` (class)
 
 ## 4.3.1 (2018-06-21)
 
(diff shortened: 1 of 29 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Deprecations:
    - `Castle.DynamicProxy.Generators.CacheKey` (class)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
+- The following AST node types have been deprecated:
+   - `Castle.DynamicProxy.Generators.Emitters.SimpleAST.ConstReference` (class)
+   - `Castle.DynamicProxy.Generators.Emitters.SimpleAST.ExpressionStatement` (class)
```
stakx: (Actually, it would be really nice to deprecate the whole namespace. :-)

stakx:
Closing this for the moment. I did some profiling, since the ASTs likely all get garbage-collected in generation 0, avoiding allocations here might give a tiny speed up, but it's likely nothing in the face of what System.Reflection.Emit causes, so probably not noticeable. If I can determine by more profiling that these changes are in fact noticeable, I'll reopen.

(Other than that, some of these changes would still be nice from a "code hygiene" point of view.)

fir3pho3nixx:
@stakx I am not following closely as I should. What test harness are you actually using for this? 

I create separate repos with console apps that batter the bejesus out of these things and measure using dot trace and dot memory. 

How are you doing this?

stakx:
@fir3pho3nixx: Same! I have a console app that lets DynamicProxy create several thousand proxy types & invokes proxies, then I've measured with dotMemory. I do this for the unmodified `master` and for the modified branch, then I compare.

I've also used `BenchmarkDotNet` to avg. measure the time it takes to generate a proxy type.

I haven't seen any significant gains from these changes at all, it's essentially a micro-optimization that pales in relation to what Ref.Emit does. The GC is also really good at cleaning up these short-lived AST node objects it seems.

So while I think the changes would be good in principle (also from a code hygiene POV) they're still work for reviewers, so probably won't pay off overall. 

Feel free to measure also... I'm not super experienced at profiling so perhaps didn't interpret everything correctly.

Issue #399, opened 2018-06-25
Do we need a thread-safe alternative to `ModuleScope.DefineType`?


stakx:
@zvirja alerted me to the fact that after deprecating `Lock` (in #391), it seems no longer possible in the layers above DynamicProxy to use `ModuleScope.DefineType` in a thread-safe manner; see https://github.com/nsubstitute/NSubstitute/pull/420/files#r197765187.

I see three options to resolve this:

0. **Do nothing.**  
 That is, leave it up to downstream libraries to ensure that *if* they access `ModuleScope`'s `ModuleBuilder`s, they must not call into their `ProxyGenerator` at the same time.

1. **Deprecate `ModuleScope.DefineType` as well.**  
 Not very realistic because, for instance, NSubstitute and Moq both rely on this API in order to mock delegate types.

2. **Revert #391.**  
 I'm biased here, but I think it would be a pity because that PR went in the right direction of closing doors into DynamicProxy's internals.

3. **Expose the `ReaderWriterLockSlim` lock that `ModuleScope` uses internally.**  
 `ModuleScope` has an API that exposes the two `ModuleBuilder`s that it uses, which essentially means that once a user starts accessing `ModuleBuilder`s, all thread-safety guarantees are off. We can no longer expose a `Lock`, but we *could* expose the internal `ReaderWriterLockSlim` instead.

4. **Augment `ModuleScope` with a thread-safe alternative for `DefineType`.**  
 Apart from (0) above, this is my personal preference. This could be in the form of a new method on `ModuleScope` which performs the necessary locking internally, for example:

   ```diff
   +[Obsolete("This method is not thread-safe. Please use `ModuleBuilder.CreateType` instead.")]
    public TypeBuilder DefineType(bool inSignedModulePreferably, string name, TypeAttributes flags)
    {
        …
    }

   +public Type CreateType(bool inStrongNamedModulePreferably,
   +                       string name,
   +                       TypeAttributes flags,
   +                       Func<TypeBuilder, Type> build)
   +{
   +    TypeCache.Lock.EnterWriteLock();
   +    try
   +    {
   +        var module = ObtainDynamicModule(disableSignedModule == false && inStrongNamedModulePreferably);
   +        var typeBuilder = module.DefineType(name, flags);
   +        return build.Invoke(typeBuilder);
   +    }
   +    finally
   +    {
   +        TypeCache.Lock.ExitWriteLock();
   +    }
   +}
   ```

[...]

PR #403, opened 2018-08-04
Add new methods `ProxyGenerator.CreateDelegateProxy[WithTarget]`

stakx:
This is in response to https://github.com/castleproject/Core/issues/399#issuecomment-404473995:

>If both Moq and NSubstitute are doing the same thing for mocking delegates maybe this is something DP should support[.]

and https://github.com/castleproject/Core/issues/399#issuecomment-405902803:

> If we add delegates proxy support, NSubstitute will not need to use the emit anymore, so it would be a good feature. Likely, the Moq as well[.]

This PR would enable you to do this:

```csharp
ProxyGenerator proxyGenerator = …;
var action = proxyGenerator.CreateDelegateProxy<Action>(interceptors: …);
action();
```

I've taken a look at how this could be implemented in DynamicProxy, and lo and behold! Most of the machinery required for delegate proxying is already there (created by @kkozmic in d3ec3bf36bbd34af1469321461e74be79ba6b89a for Castle Windsor), just not hooked up to `ProxyGenerator` yet.

~~There's a lot of things that still need to be worked out, or worked on:~~

* [X] ~~Write unit tests to check whether delegate proxiess' behavior regarding `IInvocation.Target`, `.InvocationTargetMethod`,~~ `.GetConcreteMethod()`, `.GetConcreteMethodInvocationTarget()` is in line with that of other proxy types.

   Let's skip `.GetConcreteMethod()` and `.GetConcreteMethodInvocationTarget()` since those are like the properties but ensure the methods aren't open generic; since it's already impossible to build a delegate proxy for open generic delegate types, there's probably no need to test this.

* [X] ~~Write unit tests ensuring that delegate proxy types are cached.~~

* [X] ~~We need to add documentation (in `docs/`).~~

* [x] ~~Regarding the stability of `DelegateProxyGenerator`:  @fir3pho3nixx: Are you aware of any problems with the use of `DelegateProxyGenerator` in Windsor?~~

   I'll be assuming that this hasn't caused any trouble in Windsor until I hear anything to the contrary. See https://github.com/castleproject/Core/pull/403#issuecomment-411755700.

* [x] ~~We need to decide whether there should be a `CreateDelegateProxyWithTarget` as well. I have included it for now in a separate commit (which we can easily drop if we want).~~

   I'll keep that method in the PR for now. See https://github.com/castleproject/Core/pull/403#issuecomment-411764163 below.

* [x] ~~We might want to deprecate `ModuleScope.DefineType` and all methods that allow third parties to get hold of DynamicProxy's `ModuleBuilder`s.~~

   We'll deal with deprecations in a separate PR. See https://github.com/castleproject/Core/pull/403#issuecomment-411755700.

* [x] ~~The method name: `CreateDelegateProxy` is in line with the names of other methods on `ProxyGenerator`, but semantically, I'm not sure if it's correct. Should it be called `CreateDelegate` or `CreateProxyDelegate` instead?~~

   Let's use the name as currently proposed in this PR. See https://github.com/castleproject/Core/pull/403#issuecomment-411755700.

* [X] ~~I have no idea how stable `DelegateProxyGenerator` is. I've already found out that it cannot deal with `in` parameters properly (perhaps because it doesn't copy custom attributes and custom modifiers?).~~

   `DelegateProxyGenerator` appears to be reasonably stable, it reuses a lot of the same classes that regular class proxy generators use. I'm adding unit tests to verify the most common operations one would perform on a delegate proxy.

   ~~It might be worthwhile seeing whether Windsor is actually using it~~

   Yes, it is.

* [x] ~~We need to add XML documentation.~~
* [x] ~~We need to add argument validation.~~
* [x] ~~There are tests for `DelegateProxyGenerator`, but we'd need tests for `ProxyGenerator.CreateDelegateProxy`.~~

Reviewers:  This is at an early stage, feel free to add any input or additional requirements that you might have.

/cc @zvirja


src/Castle.Core/DynamicProxy/ProxyGenerator.cs
```diff
@@ -1595,5 +1596,136 @@ protected List<object> BuildArgumentListForClassProxy(ProxyGenerationOptions opt
...
+		///   As such, the caller should additionally expect any type of exception that the given <see cref="IProxyBuilder"/> implementation may throw.
+		/// </remarks>
+		public TDelegate CreateDelegateProxyWithTarget<TDelegate>(TDelegate target, params IInterceptor[] interceptors)
+		{
+			if (target == null)
+			{
+				throw new ArgumentNullException(nameof(target));
+			}
+
+			return (TDelegate)(object)CreateDelegateProxyImpl(typeof(TDelegate), (Delegate)(object)target, interceptors);
```
stakx: These ugly type casts could be avoided by adding a `where TDelegate : Delegate` constraint. This would require C# 7.3, which unfortunately Mono does not yet support, even in the current stable version (5.12).

src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs
```diff
@@ -78,6 +78,14 @@ public Type CreateClassProxyType(Type classToProxy, Type[] additionalInterfacesT
 			return generator.GetGeneratedType();
 		}
 
+		public Type CreateDelegateProxyType(Type delegateToProxy)
+		{
+			AssertValidType(delegateToProxy);
```
stakx: As an aside, I noticed a fair amount of code duplication in the methods used for validating arguments.

* For `DefaultProxyBuilder`, there's `AssertValidType` and `AssertValidTypes`.
* For `ProxyGenerator`, there is `CheckNotGenericTypeDefinition` and `CheckNotGenericTypeDefinitions`.
* In `ProxyGenerator`, I opted not to use the latter because they throw the wrong type of exception (`GeneratorException` when I needed `ArgumentException`), so I duplicated the same validation logic once again.

These all do very similar things. Perhaps we could merge them sometime.

src/Castle.Core/DynamicProxy/IProxyGenerator.cs
```diff
@@ -1029,5 +1029,70 @@ TClass CreateClassProxy<TClass>(ProxyGenerationOptions options, params IIntercep
...
+
+		/// <summary>
+		///   Creates a delegate that, when invoked, calls the specified <paramref name="interceptors"/>.
+		/// </summary>
+		/// <param name="delegateToProxy">Type of delegate which will be proxied.</param>
+		/// <param name="interceptors">The interceptors called during the invocation of the delegate.</param>
+		/// <returns>
+		///   New delegate of type <paramref name="delegateToProxy"/> that, when invoked, calls the specified <paramref name="interceptors"/>.
+		/// </returns>
+		/// <exception cref="ArgumentNullException">Thrown when the given <paramref name="delegateToProxy"/> object is a null reference (Nothing in Visual Basic).</exception>
```
stakx: I forget to mention "[...] or when `interceptors` is a null reference [...]". Same in other XML documentation comments. Going to rectify this soon.

src/Castle.Core/DynamicProxy/IProxyBuilder.cs
```diff
@@ -133,5 +133,22 @@ public interface IProxyBuilder
...
+		/// <returns>The generated proxy type.</returns>
+		/// <exception cref="GeneratorException">
+		///   Thrown when <paramref name="delegateToProxy"/> is a generic type definition or not public /
+		///   inaccessible to DynamicProxy. In the latter case, you can mark the offending type internal and
+		///   define <see cref="InternalsVisibleToAttribute"/> for Castle DynamicProxy's assembly, if appropriate.
+		/// </exception>
+		/// <remarks>
+		///   Implementers should return a proxy type that contains an <c>Invoke</c> method with the same signature as
+		///   the <c>Invoke</c> method of the given delegate type <paramref name="delegateToProxy"/>.
+		///   This <c>Invoke</c> method should delegate all executions to the interceptors passed to the proxy type's constructor.
```
stakx: This should also mention "[...] delegate all executions *to the target and* to the interceptors passed to the proxy type's constructor [...]". Going to add this soon.

stakx:
@jonorossi, thanks for reviewing.

>> Regarding the stability of DelegateProxyGenerator: [...]
>
> It's been there since Windsor 2.5 in 2010. I don't think any of the Windsor typed factory problems are related to this, mostly lifetimes and other things.

I'll tick this one off then... unless and until I hear anything else about it.


>>We need to decide whether there should be a CreateDelegateProxyWithTarget as well.
>
> I don't really foresee much use, however it might come in handy  [...]

I take this to mean that you're not actively against that method. Since I've already implemented it and it doesn't add a lot of extra code to DynamicProxy, I'd say let's include it.


>> We might want to deprecate [...]
>
> Yep, I think we do that one separately.

Sounds good.


> What was the plan for the Moq/NSubstitute interface delegates?

Generally speaking, we'd ideally get things working downstream with the new `CreateDelegateProxy` *before* we merge this PR.

I'm hoping @zvirja will chime in and say how he plans to proceed with NSubstitute.

Speaking for Moq, I'm still working out all the required changes caused by going from a custom-emitted container interface type to simply using DP's `CreateDelegateProxy`. Here's one problem I've encountered:
 
There's one subtle change that caused me trouble in Moq. When using the new `generator.CreateDelegateProxy`, Moq's interceptors receive an `IInvocation` whose `.Method` now points to the `Invoke` method in the original (proxied) delegate type (e.g. it points at `System.Func<,>.Invoke`). When Moq was generating its own container interface type for an `Invoke` method, and then using `CreateInterfaceProxyWithoutTarget` on that, `invocation.Method` would point at that container interface's `Invoke` method (e.g. `Castle.Proxies.Func``2Proxy.Invoke`). That can be worked around, but it's a functional change and I'm not sure it is to be expected, or whether DP's delegate proxies misbehave in this regard.

>>The method name [...]

If you're satisfied with the name `CreateDelegateProxy`, then so am I. We don't need to make a big deal out of this, I just thought I'd bring it up in case you also found the name weird. But I'm totally happy to put a tick that one off on my list.

jonorossi:
> There's one subtle change that caused me trouble in Moq. When using the new `generator.CreateDelegateProxy`, Moq's interceptors receive an `IInvocation` whose `.Method` now points to the `Invoke` method in the original (proxied) delegate type (e.g. it points at `System.Func<,>.Invoke`). When Moq was generating its own container interface type for an `Invoke` method, and then using `CreateInterfaceProxyWithoutTarget` on that, `invocation.Method` would point at that container interface's `Invoke` method (e.g. `Castle.Proxies.Func``2Proxy.Invoke`). That can be worked around, but it's a functional change and I'm not sure it is to be expected, or whether DP's delegate proxies misbehave in this regard.

That sounds like a defect in `CreateDelegateProxy`. Could you add a unit test for `IInvocation.Method` and `IInvocation.MethodInvocationTarget` (they could go into the test file for either delegates or Invocation methods). Probably also worth adding one for `IInvocation.GetConcreteMethod`, maybe the whole suite of properties and methods should be tested 😉.

stakx:
I actually just re-read my above post and came to the opposite conclusion: `CreateDelegateProxy` does the right thing. I am nevertheless going to add some tests for `IInvocation.Method` (and perhaps `.ConcreteMethod` or whatsitcalled) to make sure delegate proxies' behavior are in line with other proxy types.

src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs
stakx: I'm interpreting your :+1: as, "Yes, let's get rid of this code duplication (but in a separate PR)." Let me know if I misinterpreted. :)

stakx:
@jonorossi - Here's a summary of the latest changes:

* Fixed the glitches in XML documentation comments that I mentioned in review comments above.
* Added tests for `IInvocation.Method` and `IInvocation.MethodInvocationTarget`.
* For good measure, threw in some extra tests verifying type caching, as well as updated written documentation.

This PR looks more or less finished for me, let me know if you find anything else that needs changing.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationMethodInvocationTargetTestCase.cs
```diff
@@ -70,7 +70,7 @@ public void InterfaceProxyWithTarget_MethodInvocationTarget_should_be_methodOnTa
 		}
 
 		[Test]
-		public void InterfaceProxyWithTarget_MethodInvocationTarget_should_be_null()
+		public void InterfaceProxyWithoutTarget_MethodInvocationTarget_should_be_null()
```
stakx: Fixed a typo here; otherwise unrelated to this PR.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationMethodTestCase.cs
```diff
@@ -0,0 +1,118 @@
...
+	using Castle.DynamicProxy.Tests.Classes;
+	using Castle.DynamicProxy.Tests.GenClasses;
+	using Castle.DynamicProxy.Tests.Interceptors;
+	using Castle.DynamicProxy.Tests.InterClasses;
+	using Castle.DynamicProxy.Tests.Interfaces;
+
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class InvocationMethodTestCase : BasePEVerifyTestCase
```
stakx: This whole file is more or less a copy of `InvocationMethodInvocationTargetTestCase.cs` adjusted to `.Method` (instead of `.MethodInvocationTarget`). I removed a few non-delegate proxy tests where I'm simply not sufficiently knowledgeable to know what the expected behavior should be. We can always add these back at a later time.

zvirja:
@stakx @jonorossi Sorry guys for the full silence. I'm up, just was overwhelmingly busy last weeks 😖 

> I'm hoping @zvirja will chime in and say how he plans to proceed with NSubstitute.

I'm very positive in rewriting NSubstitute core to rely on this proxy generator. As far as I recall, we don't worry whether method container is a class or an interface, the only matter is performance (again, AFAIK the difference is negligible). I'm completely fine to test the code _before_ you merge, so I can provide you with feedback:
- whether API shape is suitable for NSub
- whether performance is fine
- whether functionally it doesn't break our tests.

@stakx Sorry, I didn't have enough time to read the PR - please ensure that you have tests for the `in` attribute support for the delegates, so that values are read-only. As far as I remember, substantial part (and the most clumsy one) of our generator is about this scenario support 😅 

Let me know when you are done with the development, so I can test. Hope to have more free time next weeks, so I can take a more precise look.

Thanks for the great work!

stakx:
@zvirja - 🚥 You can start testing now; I'm pretty much done here. I've added a test for `in` parameters for you, it passes fine. (Even though strictly speaking, it probably wasn't necessary; the read-only-ness of `in` parameters should be an issue completely orthogonal to the  proxy type. Either way, delegate proxies are really just a special case of class proxies.)

stakx:
#### Feedback from Moq

`CreateDelegateProxy` itself works like a charm. Unfortunately, during conversion to this new method, I've hit an unexpected architectural problem in Moq that I don't yet know how to solve with DynamicProxy's current API.

A little background: Moq keeps two distinct objects per mock: (1) The actual mock object, which is a DP-generated proxy. (2) An instance of `Mock<T>` which is used to configure the behavior and verify invocations of the former. Basically, a glorified DP interceptor. Moq's API allows you to navigate between these two objects: You can go from `Mock<T>` to the proxy by querying the `Mock<T>.Object` property. You can go from the proxy back to `Mock<T>` by calling the static `Mock<T> Mock.Get<T>(T proxyObject)`. Under the hood, this static method casts the proxy object to `IMocked<T>` and queries that interface's `Mock` property. Moq asks DP to implement this interface on every proxy type.

It is this last bit that causes problems: delegate proxies don't support additional interfaces, so the `Mock.Get` API can no longer be implemented (because we no longer have control over the "container type" for the generated `Invoke` method).

What I'd need to get this Moq API working again is any one of the following:

1. A means to query the proxy object's interceptor(s), i.e. something like `IInterceptorsAccessor` (similar to `ITargetAccessor`). (Not sure we want more stuff like that in DP, though.) This would be sufficient because `Mock<T>` *is* the sole interceptor for its accompanying proxy.

2. A means to attach arbitrary metadata to a DP proxy. Moq could then attach a reference to the `Mock<T>` to a delegate proxy.

3. Support in the `IProxyBuilder.CreateDelegateProxyType` method for additional interfaces to be implemented. Moq would then use this lower-level method instead of `IProxyGenerator.CreateDelegateProxy`, so it can exercise more fine-grained control over the "container type" for the `Invoke` method. (Not sure whether we want to keep `IProxyBuilder` as part of the public API, though.)

@jonorossi - do you have any favourite solutions here? I think (3) would be the cleanest solution *if* we're not going to make `IProxyBuilder` non-public in the next major DP version.

TimLovellSmith:
@stakx 

>2.	A means to attach arbitrary metadata to a DP proxy. Moq could then attach a reference to the Mock<T> to a delegate proxy.

You could probably use ConditionalWeakTable to do this?

zvirja:
Hi guys,

I've just reviewed the NSubstitute lib and the usages we have and found that new API lucks a few features:
- ability to add mixins to `delegateProxy.Target` object (we use them for the same purpose as Moq - retrieve proxy configuration object from the proxy instance)
- ability to emit custom attributes to the generated method. We need this to quickly get the original delegate type from the `methodInfo` instance we internally store for the each recorded call. Basically, we use this attribute also to understand that `methodInfo` is a delegate proxy, rather than a method from a class proxy, so we render the client error messages more precisely.

I'd suggest to add overload with the `ProxyGenerationOptions` we already have and apply the specified properties like attributes (to the method itself) and mixins (to the container class/interface). It should basically solve all the issues.

> You could probably use ConditionalWeakTable to do this?

While weak-references based dictionary is a working solution, I wouldn't recommend to follow it. It's quite hacky and provides poor memory, performance and GC experience. Better change the Caslte's design, so it provides us with the required capabilities out of the box (as this is technically possible).

stakx:
> I'd suggest to add overload with the ProxyGenerationOptions we already have and apply the specified properties like attributes (to the method itself) and mixins (to the container class/interface). It should basically solve all the issues.

I've thought about this, too. The problem with this is that it turns `CreateDelegateProxy` into a leaky abstraction: You, as a user of `CreateDelegateProxy`, shouldn't know that this method creates a container class type with an `Invoke` method behind the curtains; that's an implementation detail. You don't ask it for a type, after all&mdash;you ask it for a delegate, and that's exactly what you get. Most of the options in `ProxyGenerationOptions` don't conceptually apply to delegates... they apply to (class) types.

It's because of this that I mentioned option (3) in [my above feedback post](https://github.com/castleproject/Core/pull/403#issuecomment-412487679)&mdash;if you actually need to control the type that contains the `Invoke` method used for the delegate, then perhaps by using `ProxyGenerator.CreateDelegateProxy` you're operating at the wrong abstraction level, and `IProxyBuilder.CreateDelegateProxyType` would be more appropriate. Adding a `ProxyGenerationOptions` parameter there would be a much better fit.

zvirja:
> The problem with this is that it turns CreateDelegateProxy into a leaky abstraction

I clearly see your point and do agree with it completely. Probably the point about container attributes applies here entirely. I'm just thinking whether the mixin concept might work if we look at it from the different perspective 🤔 You indeed cannot extend the delegate interface somehow, however your mixins are a part of `delegate.Target` object, which might look like the API to tag an arbitrary data. Like a reasonable tradeoff.

But probably it's still enough leaking, as the only way for you to offer this API is to generate a class/interface proxy.

>  and IProxyBuilder.CreateDelegateProxyType would be more appropriate

From the NSubstitute perspective it doesn't matters a lot whether we use the 1st or the 2nd level of API. However, the usage part with `Activator` like below looks a bit advanced:
```c#
      Type delegateProxyType = ProxyBuilder.CreateDelegateProxyType(delegateToProxy);
      object instance = Activator.CreateInstance(delegateProxyType, new []
      {
        target,
        interceptors
      });
      return delegateProxyType.GetMethod("Invoke").CreateDelegate(delegateToProxy, instance);
```

Nevertheless, any approach we decide will be fine if it's suitable. Let's see what @jonorossi thinks about that.

zvirja:
@jonorossi Do you know whether you are going to have some free time in the nearby future to review this? Just want to align NSubstitute release plans with this feature - whether it worth waiting, or better proceed without 😊

jonorossi:
@zvirja sorry about the silence on this one, I'll work my way through all the code again this weekend.

Are all 3 mocking libraries now capable of using the `ProxyGenerator.CreateDelegateProxy` API, or is there still something else we need to add? I'd much prefer them using it over `ProxyBuilder.CreateDelegateProxyType`, not even sure I like that method being public since we don't have anyone new-ing up other proxy types and having to know what the proxy class constructor arguments are.

stakx:
@jonorossi:

> Are all 3 mocking libraries now capable of using the `ProxyGenerator.CreateDelegateProxy` API, or is there still something else we need to add?

For Moq, `ProxyGenerator.CreateDelegateProxy` works... but it's not ideal. Moq needs to work around the missing tweakability of the generated delegate proxy container type. ([Here are the code changes](https://github.com/moq/moq4/compare/master...stakx:createdelegateproxy) if you're interested).

The new API probably targets a use case somewhat different from what mocking libraries need, which is, "Make me a proxy type with a method having the same signature as that delegate. Oh, and let me add some mixins/interfaces to that type."

I cannot think of a clean, generally useful API that captures such a specific requirement (`ProxyGenerator.CreateClassProxyWithMethodCompatibleWithDelegate`?). `IProxyBuilder` would be at the right level of abstraction, but like you, I'm not sure whether it should be publicly used.

If noone else here can come up with a good alternate API, I'm willing to live with the limitations of the new API.

zvirja:
> If noone else here can come up with a good alternate API, I'm willing to live with the limitations of the new API.

Do you mean to keep as is? If so, than this API is not useful for the mocking libraries... :worried: As far as I remember, that was the main goal we were trying to achieve.

Basically, we need to have container mixins and ability to emit attributes (to attach static metadata like original delegate type). We would be not the developers, if we cannot find the API shape for such a concrete requirement :sweat_smile: You know, "There is no problem in IT that can’t be solved by adding an extra layer of abstraction” :grinning:

jonorossi:
> Basically, we need to have container mixins and ability to emit attributes (to attach static metadata like original delegate type)

What other things do you wish to attach? For your example I think we should add `ProxyUtil.GetUnproxiedDelegate` like `ProxyUtil.GetUnproxiedType`.

stakx:
> [...] not useful for the mocking libraries... 😟 As far as I remember, that was the main goal we were trying to achieve.

We're ultimately trying to remove the need for mocking libraries to use `ModuleScope.DefineType`. That's the larger goal behind this PR. It's up for debate how much inconvenience this may cause those libraries. (Hopefully, as little as possible.)

> We would be not the developers, if we cannot find the API shape for such a concrete requirement

Exploring other options before deciding on this one makes sense. What are your suggestions?

TimLovellSmith:
So are you just wanting more specific API proposals here? Quick and cheerful proposal: what does everyone think of this?

```
var action = proxyGenerator.CreateDelegateProxy<Action>(config => config.WithInterceptors(interceptors).WithAttributes(new FooAttribute()).WithBaseType<MockedDelegateBase>());
```
(where MockedDelegateBase is an IMocked or whatever).

TimLovellSmith:
@stakx Good brutal honesty. Helps me understand the problem better. :)

jonorossi:
Just so everyone is clear, I'm waiting for more details requested in https://github.com/castleproject/Core/pull/403#issuecomment-417589069 before reviewing the code in detail again.

zvirja:
> What other things do you wish to attach? For your example I think we should add ProxyUtil.GetUnproxiedDelegate like ProxyUtil.GetUnproxiedType.

I've spent some time reviewing the API we have and found that it isn't suitable a lot. The API allows to get delegate type from the `proxyInstance`, while in our API we are operating with the captured `MethodInfo` objects. The method in our code base is something like:
```c#
string GetDeclaringType(MethodInfo methodInfo)
{
    var delegateType = methodInfo.GetCustomAttribute<ProxiedDelegateTypeAttribute>()?.DelegateType;
    return = delegateType != null
        ? delegateType.GetNonMangledTypeName()
        : methodInfo.DeclaringType.Name;
}
```

And we get as output either:
- `MyProject.Factories.TestFactoryType` (for class/interface proxy)
- `MyProject.Factories.TestDelegateType` (for delegate proxy)

Later we concatenate that with method name and call arguments and get either:
- `MyProject.Factories.TestFactoryType.Create("42", true)` (for class/interface proxy)
- `MyProject.Factories.TestDelegateType.Invoke("42", true)` (for delegate proxy)

Currently the code works because `methodInfo` contains our custom attribute. We don't have the proxy instance in all the places and changing that would requite substantial rework of our internal architecture.

Apart from that we are not attaching any other information to the type.

Therefore, I believe, the most perfect API for our product would be (like @stakx briefly suggested) to have smth like `ProxyGenerator.CreateClassProxyWithMethodCompatibleWithDelegate(delegateType)`, so we can decorate the container method/class with any attributes we wish. Given that we can also use `methodInfo.DeclaringType.GetCustomAttribute<>`, it doesn't really matter whether we attach that attribute directly to the method, or to the containing type.

If you see other ways given the API usage we have - please advise 😉

stakx:
@zvirja - Thanks for the feedback.

The ability to attach custom attributes to either the `Invoke` method or the containing type would work for NSubstitute, but not do Moq any good. Moq doesn't rely on custom attributes, instead it needs to be minimally able to retrieve the interceptor of a delegate proxy.

@jonorossi, what are your thoughts on such a specific method as `ProxyGenerator.CreateClassProxyWithMethodCompatibleWithDelegate(delegateType, ProxyGenerationOptions, ...)`? Compared to the other, comparatively generic methods, it would probably stick out like a sore thumb... at the same time it comes much closer to what NSubstitute, Moq (and maybe others?) actually need than the `CreateDelegateProxy` proposed in this PR.

zvirja:
> Perhaps the method just needs a less bulky name to make it more agreeable. ProxyGenerator.CreateClassProxyWithInvokeMethod(Type delegateType, ...)?

Yep, something like this would work for perfectly for us. And will cover both the needs despite the internal architecture we have.

The only question is whether we want to assert that it's actually "Class" proxy. Does it matter? As far as I understand, the interface proxy suits perfectly fine as well.

We can also rename the `CreateDelegateProxy` method to e.g. `CreateContaineredDelegateProxy(Type proxyType, , ProxyGenerationOptions containerOptions, params IInterceptor[] interceptors)`, so we say nothing about container type. We might have a better name, but idea is simple - rename `CreateDelegateProxy` in a way it's leaking the implementation explicitly 😅 

P.S. BTW, @blairconrad Do you want to participate in the tread as well for the FakeItEasy? I [see](https://github.com/FakeItEasy/FakeItEasy/blob/c8b4e7a822905ba54e724dc147ddda19b7c80c0b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs#L49) that you use expression tree to make a delegate proxy, but [they are slow](https://github.com/nsubstitute/NSubstitute/issues/248), isn't it? 😉

blairconrad:
@zvirja, thanks for thinking of us (I've poked @thomaslevesque as well). There's a lot going on here!
Yes, we do use the expression tree to make the delegate proxy, and I bet it is slow. We've just never bothered profiling. I can grab the PR and see how it would work with FakeItEasy.

blairconrad:
Hey, all.  👋 

I tried it out, as you can see in https://github.com/FakeItEasy/FakeItEasy/pull/1438. and https://github.com/FakeItEasy/FakeItEasy/issue/1437.

As far as I can tell, the new functionality works very well with FakeItEasy. The only problem is that proxied Delegates would need to be visible to the generated dynamic proxy assembly. This matches all the other proxies that DynamicProxy creates, but is a regression, since currently FakeItEasy can fake even private delegates.

jonorossi:
> @jonorossi, what are your thoughts on such a specific method as `ProxyGenerator.CreateClassProxyWithMethodCompatibleWithDelegate(delegateType, ProxyGenerationOptions, ...)`? Compared to the other, comparatively generic methods, it would probably stick out like a sore thumb... at the same time it comes much closer to what NSubstitute, Moq (and maybe others?) actually need than the `CreateDelegateProxy` proposed in this PR.
> 
> Perhaps the method just needs a less bulky name to make it more agreeable. `ProxyGenerator.CreateClassProxyWithInvokeMethod(Type delegateType, ...)`?

If we want to expose all the power to the user then is there a reason we shouldn't just make `ProxyGenerator.CreateClassProxy`, `ProxyGenerator.CreateInterfaceProxy`, etc work with delegates specifically rather than treating them separately in the public API? This was the user can decide if they want a class or interface proxy, and can add all sorts of goodies via `ProxyGenerationOptions`.

Users (including Windsor) obviously shouldn't be using `Castle.DynamicProxy.Generators.DelegateProxyGenerator` and that class should be internal like the other generators.

jonorossi:
I should have mentioned I was thinking we could special case any types that inherit from `System.Delegate` through the various `CreateXXX` methods.

stakx:
@jonorossi, that's a neat idea! It would definitely have to be documented well, since something like the following might not make much sense at first glance / to the uninitiated:

```csharp
generator.CreateInterfaceProxyWithoutTarget<Action>(...);
// "Action is not an interface!?!
// ...and anyway, it's sealed?!?
// ...and don't all delegates have a "target" method!?!?!?"
```

I'm also wondering whether we should possibly introduce an extension method `CreateDelegateProxy(this IProxyGenerator generator)` (similar to what we have in this PR) that works on top of special-cased `generator.CreateClassProxy<TDelegate>();`, for those who don't need the more fine-grained control... it would be a fairly cheap addition.

We have a rainy weekend ahead of us over here, I might give this (an alternate PR) if others agree that this idea is worth following up on.

jonorossi:
> that's a neat idea! It would definitely have to be documented well

For sure, but I don't view DynamicProxy as a library people use without reading some of the docs first.

> generator.CreateInterfaceProxyWithoutTarget<Action>(...);
> // "Action is not an interface!?!
> // ...and anyway, it's sealed?!?
> // ...and don't all delegates have a "target" method!?!?!?"

Yer, thinking about it more today I'm now thinking that is a bad idea since it internally doesn't work the way the 4 other proxy methods work (it isn't a class or interface proxy), and using the delegate's Invoke method as the "contract" for the new proxy is not how the others work they actually return a class either inheriting or implementing the "contract" which we obviously can't do with delegates. I now think that would abuse those methods, and we really only need one way for delegates not 2 (class/interface).

- `System.Delegate` provides a `Method` and a `Target` property along with a bunch of methods including `GetInvocationList`.
- When using the `delegate` keyword the C# compiler will create a class inheriting `System.MulticastDelegate` with `Invoke`, `BeginInvoke` and `EndInvoke` methods.

See the IL below for `System.Action` (with some details removed including `__DynamicallyInvokableAttribute` annotations on each method):

```
.class public auto ansi sealed System.Action extends System.MulticastDelegate
{
  .method public hidebysig specialname rtspecialname instance void
      .ctor(object 'object', native int 'method') runtime managed
  {
  }

  .method public hidebysig newslot virtual instance void
      Invoke() runtime managed
  {
  }

  .method public hidebysig newslot virtual instance class System.IAsyncResult 
      BeginInvoke(class System.AsyncCallback callback, object 'object') runtime managed
  {
  }

  .method public hidebysig newslot virtual instance void
      EndInvoke(class System.IAsyncResult result) runtime managed
  {
  }
}
```

I think we need to take a step back and work out what the resulting proxy class will look like. It won't be an class proxy or an interface proxy, but we should continue using `CompositionInvocation` even though the "contract" is only weakly implemented by having the right methods.

I've generated a proxy type for `System.Action` using the existing code and cleaned up the disassembled code so it is readable so we can discuss what other features people want to add to it. I think only then can work out the public API.

```csharp
namespace Castle.Proxies
{
    [XmlInclude(typeof(Action))]
    [Serializable]
    public class ActionProxy : ISerializable, IProxyTargetAccessor
    {
        private static MethodInfo token_Invoke = (MethodInfo)MethodBase.GetMethodFromHandle(typeof(Action).GetMethod("Invoke").MethodHandle, typeof(Action).TypeHandle);
        private static ProxyGenerationOptions proxyGenerationOptions;

        private Action __target;
        private IInterceptor[] __interceptors;

        public virtual void Invoke()
        {
            var invocation = new Action_Invoke(__target, this, __interceptors, token_Invoke, new object[0]);
            invocation.Proceed();
        }

        // ISerializable

        public virtual void GetObjectData(SerializationInfo info, StreamingContext context)
        {
            // ... Serialize instance using ProxyObjectReference
        }

        // IProxyTargetAccessor

        public virtual object DynProxyGetTarget()
        {
            return this;
        }

        public virtual void DynProxySetTarget(object target)
        {
            throw new InvalidOperationException("Cannot change the target of the class proxy.");
        }

        public virtual IInterceptor[] GetInterceptors()
        {
            return __interceptors;
        }

        public ActionProxy(Action target, IInterceptor[] interceptors)
        {
            __target = target;
            __interceptors = interceptors;
        }
    }
}

namespace Castle.Proxies.Invocations
{
    [Serializable]
    public class Action_Invoke : CompositionInvocation
    {
        public Action_Invoke(Action target, object proxy, IInterceptor[] interceptors, MethodInfo proxiedMethod, object[] arguments)
            : base(target, proxy, interceptors, proxiedMethod, arguments)
        {
        }

        protected override void InvokeMethodOnTarget()
        {
            EnsureValidTarget();
            ((Action)target).Invoke();
        }
    }
}
```

- I see no reason we can't support most if not all of `ProxyGenerationOptions`, maybe even the `BaseTypeForInterfaceProxy` but should definitely be able to add additional interfaces so Moq can implement `IMocked<T>`.
- I think the implementation of `IProxyTargetAccessor` is wrong, the proxy instance isn't the target and it can be changed if we wanted, these work more like interface proxies than class proxies.
- Should we also be implementing `BeginInvoke` and `EndInvoke` since that is part of the delegate contract.
- Should we inherit from `System.Delegate` to match the contract better, that way the proxy's public members match the target exactly

I know I've thrown a lot in this comment so please bear with me, I think we should be able to rework this to end up with a set of `CreateDelegateProxy` overloads that will provide everyone just as much flexibility as they have today.

stakx:
@jonorossi: You've given us a lot to think about. I'm hoping we don't have to take such a large step back... but if we do, that's OK. I'll need to think about this carefully for some time.

In the meantime, another possible approach for discussion: we could introduce a `ProxyGenerationOptions.AdditionalMethods`. This would work similarly to `AdditionalInterfaces` and allow one to add *any* methods with the same signatures (name, return type, parameter list, calling convention) as some specified existing methods. For example:

```csharp
options.AdditionalMethods.Add(typeof(SomeDelegateType).GetMethod("Invoke"));
generator.Create...(..., options); // business as usual
```

That's the smallest change to the public API I can come up with, and very general. It would mean mocking libraries still need to call `Delegate.CreateDelegate` themselves for the additional method on the returned proxy object... but that should be fairly trivial.

If we went that way we might want to limit `AdditionalMethods` such that any possible conflict with existing methods causes an exception right away, otherwisewe might run into some tricky questions of semantics.

jonorossi:
> You've given us a lot to think about. I'm hoping we don't have to take such a large step back

I wasn't intending for it to be a step backwards but a step forward, we improve the existing `DelegateProxyGenerator` so we can get `ProxyGenerationOptions` into it and get a set of proper overloads on `ProxyGenerator` for delegates which handle the `Delegate.CreateDelegate` for you. I suspect I dumped a little too much info into a single comment, I sort of dumped all my notes while taking a look at where we could go.

> In the meantime, another possible approach for discussion: we could introduce a `ProxyGenerationOptions.AdditionalMethods`

Interesting, however I'm struggling to see a good use case for this other than using it for delegates here, we want users adding a contract (interface or mixin). This would still require you specify a base class or interface and how would you specify the target?

> If we went that way we might want to limit `AdditionalMethods` such that any possible conflict with existing methods causes an exception right away, otherwisewe might run into some tricky questions of semantics.

This is where I think it gets a little ugly, we'd be introducing something that we don't really want, and then asking all the mocking libraries to use it, and then we'd have to support it.

jonorossi:
If we take some of what I spoke about and did one step at a time. I'm thinking we want 4 methods something like these:

```csharp
TDelegate CreateDelegateProxy<TDelegate>(TDelegate target, params IInterceptor[] interceptors) where TDelegate : System.Delegate;
TDelegate CreateDelegateProxy<TDelegate>(TDelegate target, ProxyGenerationOptions options, params IInterceptor[] interceptors) where TDelegate : System.Delegate;

object CreateDelegateProxy(Type delegateToProxy, object target, params IInterceptor[] interceptors);
object CreateDelegateProxy(Type delegateToProxy, object target, ProxyGenerationOptions options, params IInterceptor[] interceptors);
```

Straight away the code in the pull request can do the first of the pairs. We just need to look at getting the `ProxyGenerationOptions` through (the proxy class already has a field for it) and beef up `DelegateProxyGenerator` to do our bidding. We can work through each property of the options class if you feel this is a good way forward.

stakx:
> I wasn't intending for it to be a step backwards but a step forward [...]

Sorry for not being more precise, I meant "taking a step back" in the sense of "looking at the big picture & at the fundamentals", not in the sense of "regressing to a worse state". Going back to look at the very basics and work forward from there just seems like a ton of work and I was hoping we wouldn't have to go back quite that far. But I agree it might become necessary eventually, and I'm ready for it.

> Interesting, however I'm struggling to see a good use case for this other than using it for delegates here

Same here. But it has the advantage of being a generic mechanism (as opposed to the earlier, overly specific `CreateDelegateWithInvokeMethodCompatibleWith`) that should mostly integrate fairly seamlessly with the rest of DP, and be fairly logical/non-contradictory (cf. below). I just thought I'd mention it.

> This is where I think it gets a little ugly, [...]

True. Probably best indeed to drop this idea. 😁

Regarding your latest comment, that brings us back to where we were before: `CreateDelegateProxy` being a leaky abstraction IMO. You ask for a delegate (something "invokable"), yet you hand over options for a type (something "instantiable"), thus revealing how delegate creation works internally.

Sure, we can follow this solution, if we're all fine with this (IMO) minor API design flaw. This approach *is* entirely practical, after all. I guess it wouldn't be too hard to propagate the `ProxyGenerationOptions` further, all the way to to type emitting code in DP.

jonorossi:
> Sorry for not being more precise

No problem. It just felt a little like we are going around in circles so I was trying to work out what we need the implementation to have to support the required scenarios.

> But it has the advantage of being a generic mechanism (as opposed to the earlier, overly specific CreateDelegateWithInvokeMethodCompatibleWith) that should mostly integrate fairly seamlessly with the rest of DP, and be fairly logical/non-contradictory (cf. below). I just thought I'd mention it.

Hmmm yes, there was something about the `CreateDelegateWithInvokeMethodCompatibleWith` proposal I didn't like but couldn't put my finger on, and I think that is it isn't generic especially that the method doesn't need to be called Invoke it can be named anything.

I'm definitely liking `ProxyGenerationOptions.AdditionalMethods` over `CreateDelegateWithInvokeMethodCompatibleWith`.

> Regarding your latest comment, that brings us back to where we were before: `CreateDelegateProxy` being a leaky abstraction IMO. You ask for a delegate (something "invokable"), yet you hand over options for a type (something "instantiable"), thus DP is revealing how delegate proxy creation works internally.

Very true, I guess we've come down to the point where there is no way to provide the control over the type without exposing the fact that the type exists in the first place.

> Btw. like with any other class proxy, I believe a target wouldn't be strictly required.

But class proxies always have targets, either you provide one and virtual members will go there or the proxy's base class is essentially the target.

> Sometimes you just want to create a delegate whose implementation is provided by the interceptors, without ever .Proceed()-ing to (i.e. invoking) some target delegate. So perhaps we'd want both CreateDelegateProxy, and a CreateDelegateProxyWith[out]Target (depends on what the default would be).

Your `AdditionalMethods` idea isn't looking too bad with how complicated this will get and this replicates the normal proxy functionality for something fairly uncommon in most applications.

I don't think `AdditionalMethods` can be of type `MethodInfo` because semantically it is a bit wrong as that belongs to another type and it'll cause problems like we used to have before changing `AdditionalAttributes` to `IList<CustomAttributeInfo>`.

I wonder if we can change mixins to support this rather than another property which works similar.

How do you see the API being used, just throwing something together at the moment:

```csharp
// Example: Moq mocking MyClass

// add any attributes
options.AdditionalMethods.Add(typeof(SomeDelegate).GetMethod("Invoke"));

Type t = typeof(IMocked<>).MakeGenericType(typeof(MyClass));
var proxy = generator.CreateInterfaceProxyWithoutTarget(t, options, interceptors);

var @delegate = Delegate.CreateDelegate(typeof(SomeDelegate), proxy, "Invoke");
@delegate.Invoke();
```

With `AdditionalMethods` it would always require the user to have a class or interface in the mix, where as the current `DelegateProxyGenerator` does not require either.

jonorossi:
> Btw. like with any other class proxy, I believe a `target` wouldn't be strictly required. Sometimes you just want to create a delegate whose implementation is provided by the interceptors, without ever `.Proceed()`-ing to (i.e. invoking) some `target` delegate. So perhaps we'd want both `CreateDelegateProxy`, and a `CreateDelegateProxyWith[out]Target` (depends on what the default would be).

I think we should just allow the target to be null rather than another set of overloads/methods. `CreateClassProxy` and `CreateClassProxyWithoutTarget` is very wordy, and they are actually implemented quite differently (inheritance vs composition).

blairconrad:
> FakeItEasy (if I understood correctly) wouldn't use either of those as they map types outside of the generated proxy type.

We'd started mapping types outside of the generated proxy type specifically for delegates, because our hand-rolled delegates didn't lend themselves to intrinsic mapping to the "fake manager". Previously we'd used a mechanism similar to `Moq`'s, if I understand its usage.

I'd have to confirm with @thomaslevesque, but I don't think FakeItEasy would reject changing the mechanism we use to link proxies to our managers.

thomaslevesque:
> I'd have to confirm with @thomaslevesque, but I don't think FakeItEasy would reject changing the mechanism we use to link proxies to our managers.

I have no objection to change it if it makes sense 👍

jonorossi:
Sorry for the delay, I too didn't know which way to go.

> > With `AdditionalMethods` it would always require the user to have a class or interface in the mix, where as the current `DelegateProxyGenerator` does not require either.
> 
> Right. Having a container type in the mix might give some clients more control than they need (FakeItEasy, Windsor), but at the same time doesn't overly restrict others (Moq, and to some degree NSubstitute) and won't cause anyone much additional work to actually create a delegate if that's all they need. (We could provide a convenience extension method for that.)

I was wrong, just realised there is no reason you couldn't just do `CreateClassProxy` with `typeof(object)`, no methods, no additional interfaces.

> > I don't think `AdditionalMethods` can be of type `MethodInfo`
> 
> Right. That was probably a little too simplistic. We'd perhaps need to define some kind of `AdditionalMethodInfo` type:

For some reason I was actually thinking we couldn't make it of type `MethodInfo` because that would cause problems because MethodInfo defines so much stuff that only relates to the method on another type, however we'd obviously just document the few things we read off the template (name, return parameter, return type, generic arguments, parameters) and what isn't supported (custom attributes), and throw for constructors, abstract and static. We can't read custom attributes from the MethodInfo because some can't be decoded from the reflection API (i.e. why we moved to our own CustomAttributeInfo.

> > I wonder if we can change mixins to support this
> 
> Perhaps by special-casing the mixin facility so it accepts delegate types as mixins? Delegate types don't implement any _explicit_ interface but they _do_ fulfill a well-known contract (of which we'd probably be forgiven for cherrypicking just`Invoke` and ignoring `[Begin|End]Invoke`).
> 
> (Btw., I don't have a clear favourite solution so far. They all have their distinct [dis]advantages.)

Having more of a look at mixins I don't think there will even be much special casing for `MulticastDelegate` ([see this article why you should pretend `System.Delegate` does not exist](https://blogs.msdn.microsoft.com/brada/2004/02/05/delegate-and-multicastdelegate/)) especially since reflection doesn't hide that delegates are just a class. We could even decide to call `AddMixinInstance` but it might be safer (and more discoverable) if we add a `AddMixinDelegate` or `AddDelegateMixin` method. Right now the `ProxyGenerationOptions` and `MixinData` classes don't do much checking, `MixinData` has one line we could change to populate the contract from the delegate type rather than the object instance's interface. Then come to the emitted proxy type, we could emit slightly different code depending on if there was a target to proceed to but ultimately the main structure of the mixin fields and the likes seem pretty close to what we want.

Here is a runnable test case with a mixin, along with a bunch of comments for what I'm thinking:

```csharp
using System;
using Castle.DynamicProxy;

public delegate int AdditionDelegate(int x, int y);

class Program
{
    static void Main(string[] args)
    {
        options.AddMixinInstance(new Addition()); // working mixin for running code
        //options.AddMixinInstance(typeof(AdditionDelegate));              // without target
        //options.AddMixinInstance(new AdditionDelegate((x, y) => x + y)); // with target

        // MixinData.MixinInterfaces[0] - the contract is the delegate type
        // MixinData.Mixins[0]          - the delegate instance if one, or null

        var proxy = generator.CreateClassProxy(typeof(object), options, new HelloInterceptor());

        // Obviously not required, but we can just document that all delegates used as mixins will
        // be given the method name "Invoke" since that is the name in the metadata of the class,
        // however would be nice to have a method so a user doesn't need to call CreateDelegate
        var @delegate = (AdditionDelegate)Delegate.CreateDelegate(
            typeof(AdditionDelegate), (object)proxy, "Invoke");
        //var @delegate = ProxyUtil.CreateDelegateToMixin<AdditionDelegate>(proxy);
        //var @delegate = ProxyUtil.CreateDelegateToMixin(proxy, typeof(AdditionDelegate));

        int z = @delegate.Invoke(10, 20);
        Console.WriteLine(z);
    }
}

// Working mixin implementation that we'd be able to do via just a delegate
public interface IAddition
{
    int Invoke(int x, int y);
}
public class Addition : IAddition
{
    public int Invoke(int x, int y)
    {
        Console.WriteLine("Calculating...");
        return x + y;
    }
}

public class HelloInterceptor : IInterceptor
{
    {
        Console.WriteLine("Before");
        invocation.Proceed();
        Console.WriteLine("After");
    }
}
```

It doesn't allow you to emit custom attributes on the proxy's methods (which sounded like a nice to have, not a requirement), but it does still allow custom attributes on the proxy type, as well as doing whatever you like with the container in terms of base class, additional interfaces, etc. The implementation of this should be quite easy because the InvokeMethodOnTarget will call on the mixin field typed after the delegate (rather than an interface like normal), which is much cleaner than having a new set of fields if we went with the `AdditionalMethods` option. I think this way will also prevent users misusing something like `AdditionalMethods` doing something undesirable.

Does this proposal fit everyone's requirements? If you are not sure, please just ask. Please let me know which bits you like or dislike.

stakx:
@jonorossi - apologies for being a bit passive myself regarding this issue. I'm still quite busy settling into my new daytime job, trying to be new best buddies with PHP. (It's challenging. 😃)

> Does this proposal fit everyone's requirements?

Speaking for Moq, this proposal should work.

> Please let me know which bits you like or dislike.

* I like `options.AddMixinInstance(new AdditionDelegate((x, y) => x + y))`.

* `options.AddMixinInstance(typeof(AdditionDelegate))` on the other hand, not so much.  Calling a `Type` object an "instance" seems contradictory. I'd prefer something without `Instance` in the method name, e.g. `options.AddMixin(typeof(…))` or `options.AddDelegateMixin(typeof(…))`. Perhaps delegate mixins should generally use a different method name for consistency.

* `ProxyUtil.CreateDelegateToMixin(proxy, typeof(AdditionDelegate))` sounds like a useful helper method to have, too, because most clients are going to have to do the `Delegate.CreateDelegate` step one way or another. Having such a helper method in DynamicProxy would relieve clients from having to map delegate signatures to the generated method.

The last point brings me to...

> ```csharp
> // Obviously not required, but we can just document that all delegates used as mixins will
> // be given the method name "Invoke" since that is the name in the metadata of the class,
> // however would be nice to have a method so a user doesn't need to call CreateDelegate
> ```

As noted above, *if* we added a method such as `ProxyUtil.CreateDelegateToMixin`, we wouldn't even need to document the name(s) of the generated delegate methods. DynamicProxy would take care of mapping a delegate type to the right method.

That being said, I believe it would be very reasonable to call all delegate mixins' methods `Invoke`, and officially document that, too. Only then would `ProxyUtil.CreateDelegateToMixin` be a truly optional helper method that one can choose to use or not use.

Calling all delegate mixin methods `Invoke` should work fine as long as noone tries to add mixins for two delegates having the same signature. Say, if someone does this:

```csharp
options.AddMixinInstance(new Func<object, bool>(obj => true));
options.AddMixinInstance(new Predicate<object>(obj => true));
```

We'd probably have to identify the signature collision and throw an `InvalidOperationException` (or similar) for the second call. (Alternatively, we could do the same that we do for interface methods and switch to an "explicit implementation" name whenever a name/signature collision occurs.)

> ```csharp
> // MixinData.MixinInterfaces[0] - the contract is the delegate type
> // MixinData.Mixins[0]          - the delegate instance if one, or null
> ```

I cannot comment on this bit yet, as I'm still somewhat unfamiliar with how mixins are kept track of internally. Will look into it.

stakx:
P.S. I am not sure whether the "with target" / "without target" mapping to `AddMixinInstance(delegate)` / `AddMixinInstance(delegateType)` is obvious enough. Up until now, it has always been possible to `Proceed` for invocations of mixed-in methods; that'll no longer be the case for `AddMixinInstance(delegateType)`. (Are we all comfortable with that change btw.?) Just saying this as another argument why we might possibly want to call that method overload something different.

thomaslevesque:
> Does this proposal fit everyone's requirements?

To be honest, I didn't follow very closely and I'm not sure what you're proposing exactly... With your proposal, how should we create a delegate proxy?

stakx:
@thomaslevesque, I'll attempt an answer by example. Following @jonorossi's proposal, if all you needed were a delegate proxy, then usage might look like this in the most basic case:

```csharp
// You'll need distinct `options` per delegate type. Specify additional options as needed:
options.AddMixinInstance(typeof(SomeDelegate));

// Then use it to generate a proxy. That proxy will have an `Invoke` method
// with the specified delegate type's signature. Base type can be anything:
var proxy = proxyGenerator.CreateClassProxy(typeof(object), options, new MyInterceptor(...));

// Either call `Delegate.CreateDelegate` for that `Invoke` method yourself, or use a
// helper method provided by DynamicProxy:
var someDelegate = (SomeDelegate)ProxyUtil.CreateDelegateToMixin(proxy, typeof(SomeDelegate));

// Invoking the returned delegate will trigger the interceptor(s) specified above:
someDelegate(...)
```

blairconrad:
@stakx, thanks for the summary (and @jonorossi for the proposal!). It seems like an easy enough recipe to follow, and I've no doubt that FakeItEasy could use it.

I've never used DynamicProxy outside of FakeItEasy, and have no experience with the mixin functionality, so maybe this is all idiomatic, but the `CreateDelegateToMixin` invocation sounds funny. What does `ToMixin` mean here? It sounds more like `FromMixin` to me…

thomaslevesque:
@stakx, thanks. I guess we could use this, but it doesn't seem very straightforward, especially compared to the `CreateDelegateProxy` that was suggested at some point.

blairconrad:
@thomaslevesque, it's true, but if I understand things, it does offer the ability to have the proxy implement an interface, which would be helpful if we want to use that mechanism for providing a link to the FakeManager…

jonorossi:
I going to start by explaining how the current mixin functionality works, I obviously assumed that was obvious which it isn't and isn't even a common or well documented feature.

DynamicProxy Mixins aren't very complicated. You add an object instance as a "mixin" during proxy creation, DP will look at the interfaces the class implements and it'll additionally implement those interfaces on the proxy class and pass calls to those methods through to the mixin object. Proxies that have a target obviously hold a reference to the target (in DP it is just a field of the proxy class), the same occurs with mixin "targets", it creates a field for each interface the mixin implements. You can't cast the proxy instance to the mixin class but you can to any interface the mixin implements.

My proposal is to do the same sort of thing with delegates. Since we can't explicitly expose the delegate's contract on a proxy class (a bit like the mixin isn't explicitly exposed) the proxy class would implement the delegate contract and optionally hold a reference to a target delegate if you do "Proceed" to the delegate. I think this proposal is cleaner than adding a completely different type of proxy for delegates.

@stakx I know I wrote a lot in my last comment, but yes I agree we'd want a different method name for adding the delegate type. I didn't mention it in my last comment but one feature of mixins that should have been supported was to specify which interfaces you want DP to implement for you rather than it just adding all, I think this would fit in well with the add mixin method we'd add for delegates.

@stakx Mixin support already requires the interfaces of mixins to be unique, i.e. two mixins can't implement the same interface; and we'd do the same thing here so you can't add two delegates with the same signature to the same proxy.

@blairconrad `CreateDelegateToMixin` or `CreateDelegateFromMixin`, could probably go either way, but my thinking is that it is creating a delegate pointing to a mixin.

@thomaslevesque At first it might seem more complicated than the `CreateDelegateProxy` proposal without knowing DP mixins, but this way there is no need to document a special "delegate matched container type". Does my explanation above help make the proposal clearer that this results in essentially the same proxy type because we would have had to hold on to an optional delegate target too, but this way you can generate the proxy just as you would a class or interface and the extra method(s) get mixed in?

thomaslevesque:
> Does my explanation above help make the proposal clearer that this results in essentially the same proxy type because we would have had to hold on to an optional delegate target too, but this way you can generate the proxy just as you would a class or interface and the extra method(s) get mixed in?

Yes, it's perfectly clear. Thanks @jonorossi!

blairconrad:
Ah, yes. That is very clear. Thanks for detailed explanation, @jonorossi. I'm sorry you had to put so much time into it!

stakx:
@jonorossi:

Apologies once again for the long delay. [Above](https://github.com/castleproject/Core/pull/403#issuecomment-423430574), you wrote:

>  one feature of mixins that should have been supported was to specify which interfaces you want DP to implement for you rather than it just adding all, I think this would fit in well with the add mixin method we'd add for delegates.

Let me attempt a sketch for a new mixin API:

```csharp
public void AddMixin(object instance, params Type[] mixedInTypes)
```

This should cover what you mentioned. `mixedInTypes` would have to be either:

1. one or more interface types, all of which must be implemented by `instance`'s runtime type (directly, or via any of its base classes).
2. a single delegate type, which must be identical to `instance.GetType()`.
3. possibly an empty array, in which case the method would assume that you want to mix in all interface types implemented by `instance`'s runtime type.

```csharp
public void AddMixin(object instance, Type mixedInType)
```

This wouldn't be strictly necessary, as it is a special case of the previous method. It would be especially useful when you only want to specify one type to be mixed in, which would likely be the case for delegate mixins. It could also prevent an unnecessary params array allocation in other cases.

```csharp
public void AddMixin(Type mixedInType)
```

This would be needed for delegate mixins when you do not want to specify an `instance` (aka target).

We could optionally relax constraints a little further and allow `mixedInType` to be an interface type, which would turn this API into an alternative to `AdditionalInterfaces` (but at the level of the proxy generation options vs. single `Create…Proxy…` calls). (This might be useful for mocking libraries such as Moq that let all their proxies implement a specific interface.)

```csharp
public void AddMixinInstance(object instance)
```

This is what we have today. We *could* deprecate it in favor of (reimplement it by delegating to) the first method listed above.

Finally, we could go a little further and have a `MixinCollection Mixins { get; }` in `ProxyGenerationOptions` so you could write `options.Mixins.Add(...)` instead of `options.AddMixin(...)`. That's mostly cosmetics, though.

Any thoughts? If this sounds about right, I could attempt an implementation for it.

stakx:
Closing in favour of the PR referenced above.

jonorossi:
@stakx apologies I didn't respond to your questions here on 2 Oct 2018, not sure how I missed it.

jonorossi:
@stakx were you planning to make any of your proposed changes on 2 Oct 2018 around being able to add a mixin and specifying the interfaces? If so, did you want to create another issue so we can look over the API changes.

stakx:
@jonorossi:

> were you planning to make any of your proposed changes on 2 Oct 2018 around being able to add a mixin and specifying the interfaces?

TBH, the mixin API in general isn't super-dear to me. I cannot say whether it gets used much at all, but I suspect not. So the time spent on improving it would perhaps be better invested elsewhere.

That being said, if you'd like, I'll open a new issue so we can track this properly.

jonorossi:
> TBH, the mixin API in general isn't super-dear to me. I cannot say whether it gets used much at all, but I suspect not. So the time spent on improving it (purely for its own sake) would perhaps be better invested elsewhere.
> 
> When I prepared my delegate mixin PR I decided on more specifically-named methods, because it's more obvious to the end user what can be mixed in. With the proposal I originally made above, you'd have had to check the XML documentation to figure out that you can use it to mix-in delegates.
> 
> That being said, if you'd like, I'll open a new issue so we can track this properly.

Good answer, that is what I expected. I'm happy for us to improve things in the future when/if there is demand.

PR #404, opened 2018-08-16
Add trace log level

pi3k14:
Modified logger to support Trace log in Log4Net & NLog (fallback to Debug for Serilog).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ Deprecations:
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
 
+## 4.3.2 (2018-09-03)
+
+Enhancements:
+ - Added Trace level to logging (maps to Verbose in Serilog)
+
 ## 4.3.1 (2018-06-21)
 
 Enhancements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+Enhancements:
+ - Added Trace level to logging (maps to Verbose in Serilog)
+
 Deprecations:
 - The API surrounding `Lock` has been deprecated. This consists of the members listed below. Consider using the Base Class Library's `System.Threading.ReaderWriterLockSlim` instead. (@stakx, #391)
    - `Castle.Core.Internal.Lock` (class)
@@ -17,10 +20,6 @@ Deprecations:
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
 
-## 4.3.2 (2018-09-03)
-
-Enhancements:
- - Added Trace level to logging (maps to Verbose in Serilog)
 
 ## 4.3.1 (2018-06-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 Enhancements:
- - Added Trace level to logging (maps to Verbose in Serilog)
+ - Added Trace level to logging, maps to Verbose in Serilog. (@pi3k14, #404)
 
 Deprecations:
 - The API surrounding `Lock` has been deprecated. This consists of the members listed below. Consider using the Base Class Library's `System.Threading.ReaderWriterLockSlim` instead. (@stakx, #391)
@@ -19,7 +19,6 @@ Deprecations:
    - `Castle.DynamicProxy.Generators.CacheKey` (class)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
-
 
 ## 4.3.1 (2018-06-21)
 
--- a/src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
+++ b/src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
@@ -36,10 +36,24 @@ namespace Castle.Core.Logging.Tests
 		}
 
 		[Test]
+		public void LevelTrace()
+		{
+			logger.Level = LoggerLevel.Trace;
+
+			Assert.IsTrue(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsErrorEnabled, "LevelFilteredLogger.IsErrorEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsFatalEnabled, "LevelFilteredLogger.IsFatalErrorEnabled is not returning true when the level is Debug");
+		}
+
+		[Test]
 		public void LevelDebug()
 		{
 			logger.Level = LoggerLevel.Debug;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
 			Assert.IsTrue(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning true when the level is Debug");
 			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Debug");
 			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Debug");
@@ -52,6 +66,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Info;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Info");
 			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Info");
 			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Info");
@@ -64,6 +79,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Warn;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Warn");
 			Assert.IsFalse(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning false when the level is Warn");
 			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Warn");
@@ -76,6 +92,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Error;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Error");
 			Assert.IsFalse(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning false when the level is Error");
 			Assert.IsFalse(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning false when the level is Error");
@@ -88,6 +105,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Fatal;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Fatal");
 			Assert.IsFalse(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning false when the level is Fatal");
 			Assert.IsFalse(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning false when the level is Fatal");
@@ -100,6 +118,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Off;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Off");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Off");
 			Assert.IsFalse(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning false when the level is Off");
 			Assert.IsFalse(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning false when the level is Off");
--- a/src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
+++ b/src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
@@ -40,12 +40,12 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Trace;
 
-			Assert.IsTrue(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning true when the level is Debug");
-			Assert.IsTrue(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning true when the level is Debug");
-			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Debug");
-			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Debug");
-			Assert.IsTrue(logger.IsErrorEnabled, "LevelFilteredLogger.IsErrorEnabled is not returning true when the level is Debug");
-			Assert.IsTrue(logger.IsFatalEnabled, "LevelFilteredLogger.IsFatalErrorEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning true when the level is Trace");
+			Assert.IsTrue(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning true when the level is Trace");
+			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Trace");
+			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Trace");
+			Assert.IsTrue(logger.IsErrorEnabled, "LevelFilteredLogger.IsErrorEnabled is not returning true when the level is Trace");
+			Assert.IsTrue(logger.IsFatalEnabled, "LevelFilteredLogger.IsFatalErrorEnabled is not returning true when the level is Trace");
 		}
 
 		[Test]
@@ -66,7 +66,7 @@ namespace Castle.Core.Logging.Tests
 		{
 			logger.Level = LoggerLevel.Info;
 
-			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Info");
 			Assert.IsFalse(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning false when the level is Info");
 			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Info");
 			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Info");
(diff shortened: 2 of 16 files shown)


jonorossi:
@pi3k14 this request is a bit out of nowhere. Do you intend to use it in one of the Castle libraries, or are you using Castle's `ILogger` in your project?

snakefoot:
Notice Serilog has a LogLevel called Verbose

pi3k14:
@jonorossi We are using `ILogger` in a project(library), and wan't to enable Trace possibilities from that.
@snakefoot Updating `SerilogLogger` to map Trace to Verbose.

src/Castle.Core/Core/Logging/ConsoleLogger.cs
```diff
@@ -32,7 +32,7 @@ public class ConsoleLogger : LevelFilteredLogger
 		///   set to <c>LoggerLevel.Debug</c> and the <c>Name</c>
 		///   set to <c>String.Empty</c>.
 		/// </summary>
-		public ConsoleLogger() : this(String.Empty, LoggerLevel.Debug)
+		public ConsoleLogger() : this(String.Empty, LoggerLevel.Trace)
```
jonorossi: Not convinced the console logger should default to trace logging.

stakx:
@jonorossi:

> I know we've spoken about reducing our dependence on Castle's ILogger which will take a while, but in the mean time I don't see any objection to this change. I'm assuming since you guys didn't jump in you don't mind either.

I don't mind the change.

(It seems to me a pity that we need to support every log level ever conceived, even when they *appear* redundant in some cases; feels to me as if it made the common logging abstraction fall apart. That being said, adding more log levels is probably much easier and more straightforward than investing time in devising some opaque & overly clever log level mapping.)

fir3pho3nixx:
I vehemently disagree with this. I am not happy. I won’t block this to help folks out but as maintainers we should not be in this position.

src/Castle.Core/Core/Logging/ConsoleLogger.cs
pi3k14: I reversed this change.

pi3k14:
@jonorossi I added an entry in the ChangeLog, but unsure about your regime for version number and dates, please have a look.

@fir3pho3nixx, a bit unsure about what "position" you don't like to be in. But maybe the Castle project should have a look at using [LibLog ](https://github.com/damianh/LibLog)(or EventSourceLogger, an ILogger implementation) and throw away your own logging abstraction?

Lastly - I don't know why most of the code in this project used constructs like
``` c#
public void Trace(Func<string> messageFactory)  
{  
  if (IsTraceEnabled)
  {
    Logger.Log(declaringType, Level.Trace, messageFactory.Invoke(), null);
  }
}
```
(I "normalized" all code to be similar)

my preferred version would be
```  c#
public void Trace(Func<string> messageFactory)
{
    Logger.Log(declaringType, Level.Trace, messageFactory.Invoke(), null);
}
```
and let only the logger implementations perform the check for enabled log level.

Give a hint, and I can change all to the latter.

snakefoot:
The idea is avoid the "expensive" Invoke, when loglevel is not active


Sent from my Samsung device


-------- Original message --------

stakx:
@pi3k14:

> unsure about your regime for version number and dates

Normally all changes are listed under "Unreleased" heading; then when we're ready to publish a release that title gets replaced with the version number and release date. This means you as a contributor don't need to worry about version number & date at all, this will be determined by the Castle team (@jonorossi to be precise).

I was just briefly skimming over the changes, two things I noticed:

* Is the change of `DiagnosticsLogger`'s new default level of `Trace` (instead of `Debug` as it was previously) intentional? Not complaining, nor requesting a change, just mentioning it in case you intended to revert that one too but overlooked it.

* Please check if there's an indentation issue in `SerilogLogger.cs`, looks like it (but hard to tell for sure on a mobile layout).

jonorossi:
> It seems to me a pity that we need to support every log level ever conceived, even when they appear redundant in some cases; feels to me as if it made the common logging abstraction fall apart.

> I vehemently disagree with this. I am not happy. I won’t block this to help folks out but as maintainers we should not be in this position.

> But maybe the Castle project should have a look at using LibLog (or EventSourceLogger, an ILogger implementation) and throw away your own logging abstraction?

@stakx @fir3pho3nixx @pi3k14 you can probably tell I haven't been quick to progress this pull request, the new log level won't be used by Castle projects so only by users, this is the first time I recall this log level being proposed. The problem we've got right now is that we've discussed deprecating Castle's logging abstraction (which predates newer .NET logging, its from around 2004-2005) but haven't done anything about formalising that proposal to let people know our position and encourage users to use other libraries directly. We've briefly discussed how we could support a new logging facility for Windsor that isn't tied to the current Castle logging abstraction and would give users more freedom and not force all logging to a lowest common denominator interface (i.e. loosing structured logging when using Serilog).

@pi3k14 what logging sink/adapter are you using under the Castle abstraction?

pi3k14:
> what logging sink/adapter are you using under the Castle abstraction?

@jonorossi for the time being we are using NLog (and Log4Net), but we are also in a process to find a better solution. Not sure yet what that is going to be ... [EventSourceLogger](https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs) and [EventFlow](https://github.com/Azure/diagnostics-eventflow) looks interesting though.

jonorossi:
> for the time being we are using NLog (and Log4Net), but we are also in a process to find a better solution

From my perspective, in Castle Project's prime (before 2010) users were building their applications with Windsor, MonoRail, ActiveRecord and other bits of the Castle stack so making use of the Castle logging abstraction wasn't a big thought as it just made sense; obviously most of these are now deprecated and other libraries have taken their place (i.e. ASP.NET MVC). We've been slimming down Windsor in the last year, and slowly deprecating bits dumped into what is today Castle Core.

I'm not trying to be rude just completely honest to learn how people are using it, but if you don't see the Castle logging abstraction in your project's future why are you continuing to use it today rather than just talking to NLog directly?

pi3k14:
@stakx Fixed some space/tab glitches, corrected placement in ChangeLog. As for the Diagnostics logger I "feel" :) this is correct, but in practise it doesn't make any difference because of the simple log level mappings.

pi3k14:
@jonorossi "Rome wasn't built in one day" :), and it isn't just our libraries/apps but also third parties (eg. ASP.NET Boilerplate) that we have custom branches of.
The problem is really with libraries/frameworks, we don't want to end up in a situation where we have to rewrite "a lot" of those because the apps switched logging (and NLog is probably not the future).

jonorossi:
> "Rome wasn't built in one day" :)

Very true 👍

> [...] it isn't just our libraries/apps but also third parties (eg. ASP.NET Boilerplate) that we have custom branches of.

Oh, I didn't realise ASP.NET Boilerplate recommended using Castle's logging: https://aspnetboilerplate.com/Pages/Documents/Logging.

Since the 3 major .NET logging libraries have a log level under debug for tracing then we should probably just add it and move on. We'll separately need to start putting together our roadmap for Castle's logging abstraction and adapters, things won't happen overnight, it'll take a while to unroll things and will need to start from the top to add guidance and a new logging facility.

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
```diff
@@ -113,13 +122,106 @@ public virtual Core.Logging.ILogger CreateChildLogger(String name)
...
+		}
+
 		/// <summary>
 		///   Logs a debug message.
 		/// </summary>
 		/// <param name="message"> The message to log </param>
 		public void Debug(string message)
 		{
-			Log(LogLevel.Debug, message);
+			if (IsDebugEnabled)
```
jonorossi: Why did you add all these `if (IsDebugEnabled)` type checks to existing log methods, the `IsXXXEnabled` check is only needed on log methods taking a `Func` to avoid invoking the function if not necessary, the log message will get dropped down the stack by another check.

It would also be best to keep this pull request focused on just adding the trace log level, so just keep the current code structure with the return statement even though it is a little verbose.

src/Castle.Services.Logging.SerilogIntegration/SerilogLogger.cs
```diff
@@ -14,356 +14,419 @@
...
-    using Serilog.Events;
+	using Serilog;
+	using Serilog.Events;
 
 #if FEATURE_SERIALIZATION
-    [Serializable]
+	[Serializable]
 #endif
-    public class SerilogLogger :
+	public class SerilogLogger :
```
jonorossi: You'll need to take another look at `SerilogLogger.cs` because it still shows as all changed.

CHANGELOG.md
```diff
@@ -17,6 +20,7 @@ Deprecations:
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.ApplyTo` (method)
    - `Castle.DynamicProxy.Serialization.CacheMappingsAttribute.GetDeserializedMappings` (method)
 
+
```
jonorossi: You've left an extra blank line here.

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
pi3k14: Reverted every change not directly related to implementation of Trace log level.

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
```diff
@@ -130,9 +232,8 @@ public void Debug(Func<string> messageFactory)
 		{
 			if (IsDebugEnabled == false)
```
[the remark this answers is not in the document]
pi3k14: Ugh, shame on me - fixed

src/Castle.Services.Logging.SerilogIntegration/SerilogLogger.cs
pi3k14: Reverted

CHANGELOG.md
pi3k14: Fixed

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+Enhancements:
+ - Added Trace level to logging (maps to Verbose in Serilog)
```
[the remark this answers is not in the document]
pi3k14: Fixed

src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
```diff
@@ -35,11 +35,25 @@ public void SetUp()
...
+		public void LevelTrace()
+		{
+			logger.Level = LoggerLevel.Trace;
+
+			Assert.IsTrue(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsDebugEnabled, "LevelFilteredLogger.IsDebugEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsInfoEnabled, "LevelFilteredLogger.IsInfoEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsWarnEnabled, "LevelFilteredLogger.IsWarnEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsErrorEnabled, "LevelFilteredLogger.IsErrorEnabled is not returning true when the level is Debug");
+			Assert.IsTrue(logger.IsFatalEnabled, "LevelFilteredLogger.IsFatalErrorEnabled is not returning true when the level is Debug");
```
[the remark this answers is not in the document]
pi3k14: (Blush) - to hasty with that ones. Fixed.

src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
```diff
@@ -52,6 +66,7 @@ public void LevelInfo()
 		{
 			logger.Level = LoggerLevel.Info;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
```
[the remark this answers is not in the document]
pi3k14: ditto

src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
```diff
@@ -64,6 +79,7 @@ public void LevelWarn()
 		{
 			logger.Level = LoggerLevel.Warn;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
```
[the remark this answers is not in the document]
pi3k14: ditto

src/Castle.Core.Tests/Core.Tests/Logging/LevelFilteredLoggerTests.cs
```diff
@@ -76,6 +92,7 @@ public void LevelError()
 		{
 			logger.Level = LoggerLevel.Error;
 
+			Assert.IsFalse(logger.IsTraceEnabled, "LevelFilteredLogger.IsTraceEnabled is not returning false when the level is Debug");
```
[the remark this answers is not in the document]
pi3k14: ditto

src/Castle.Core/Core/Logging/LevelFilteredLogger.cs
```diff
@@ -103,6 +103,99 @@ public String Name
...
+		/// <param name = "message">The message to log</param>
+		public void Trace(string message)
+		{
+			if (IsTraceEnabled)
+			{
+				Log(LoggerLevel.Trace, message, null);
+			}
+		}
+
+		public void Trace(Func<string> messageFactory)
```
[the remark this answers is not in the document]
pi3k14: Yeah, copy paste from other level implementation. Added comment.

src/Castle.Core/Core/Logging/LevelFilteredLogger.cs
```diff
@@ -103,6 +103,99 @@ public String Name
...
 
+		#region Trace
+
+		/// <summary>
+		///   Logs a trace message.
+		/// </summary>
+		/// <param name = "message">The message to log</param>
+		public void Trace(string message)
+		{
+			if (IsTraceEnabled)
```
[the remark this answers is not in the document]
pi3k14: fixed

pi3k14:
All comments should be resolved.

snakefoot:
Think you should keep the Level-check when you can avoid string.Format call or allocation of SystemStringFormat (Same overhead as calling Invoke)

This should be applied to all level-implementations


Sent from my Samsung device


-------- Original message --------

src/Castle.Core/Core/Logging/LevelFilteredLogger.cs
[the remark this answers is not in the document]
stakx: Ah, I see, the checks were already there in `master` for all overloads. Sorry, my bad.

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
```diff
@@ -113,6 +122,84 @@ public virtual Core.Logging.ILogger CreateChildLogger(String name)
...
+
+		/// <summary>
+		///   Logs a trace message.
+		/// </summary>
+		/// <param name="messageFactory"> Factory constructing lazily the message to log if the level is enabled </param>
+		public void Trace(Func<string> messageFactory)
+		{
+			if (IsTraceEnabled)
+			{
+				Log(LogLevel.Debug, messageFactory());
```
stakx: `LogLevel.Debug` should be `LogLevel.Trace`, I think.

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
```diff
@@ -113,6 +122,84 @@ public virtual Core.Logging.ILogger CreateChildLogger(String name)
...
+
+		/// <summary>
+		///   Logs a trace message.
+		/// </summary>
+		/// <param name="exception"> The exception to log </param>
+		/// <param name="format"> Format string for the message to log </param>
+		/// <param name="args"> Format arguments for the message to log </param>
+		public void TraceFormat(Exception exception, string format, params object[] args)
+		{
+			if (IsTraceEnabled)
```
stakx: Detail / question: You've removed the checks even for most of the `Trace`-specific methods... why not here, too?

stakx:
@jonorossi:

> The problem we've got right now is that we've discussed deprecating Castle's logging abstraction (which predates newer .NET logging, its from around 2004-2005) but haven't done anything about formalising that proposal to let people know our position and encourage users to use other libraries directly.

Should we open an issue for this? ("Future of Castle logging facilities"?)

Just briefly, :+1: to @pi3k14 for mentioning LibLog. I originally discovered it via [Nicholas Blumhardt's blog article "Good citizenship - logging from .NET libraries"](https://nblumhardt.com/2017/07/library-logging/) and, without having it taken for a spin yet, some aspects of it make good sense to me:

* **Not defining our own logging abstraction**, but instead using whatever the application above Castle decided on. As you said above, other logging libraries have by now surpassed Castle's own logging facilities, and there is perhaps little point in keeping it going or even inventing something brand new. My personal vote would go to deprecating Castle logging facilities, if we can find a good way to do so.

* **Avoiding hard static dependencies on any particular logging libraries through run-time discovery** using Reflection. Following this approach would seem to make things really easy for users of Castle libraries (*if* we take care to keep our use of LibLog up-to-date), and it would seem that we'd no longer need the various logging library integration projects. This in turn would perhaps simplify the versioning & release process quite a bit. (I vaguely recall an episode where @fir3pho3nixx and you had to look into this, and ended up releasing Core & all logging integration projects together and in version lockstep.)

Some things that I'm not sure about when it comes to `LibLog`:

* **Relying on run-time Reflection seems somewhat risky.** You don't know whether logging will work until you try it. Static dependencies give you more of a guarantee in that respect. This "risk" is however mitigated by the fact that the number of commonly used .NET logging libraries is probably quite small and run-time detection is entirely feasible. More arcane logging libraries would require more work to be wired up with Castle, however.

* **Performance.** I quickly checked, it seems that Reflection is only used for initial discovery & preparing logging delegates, but from then on logging is virtually as fast as using a logging library directly. If reflection were used for every single logging operation, that would obviously be a huge no-go.

* **Fallback when no logging library is detected.** Not sure what LibLog does when it fails to detect a logging library. Also not sure why it doesn't have a provider for the (admittedly horrible) logging facility built into the BCL (`System.Diagnostics`) but IMO this should be the fallback.

I'm by no means set on LibLog, but it does look like an interesting option to me.

pi3k14:
@jonorossi, @stakx, @snakefoot I am holding back the last check-in until we are in agreement on what "correct" check of log levels are. 
My current understanding is, use a IsTraceEnabled guard when Log function uses functors, creates new objects or use string.Format?
Only apply this rule to new methods created for Trace logging?

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
pi3k14: Fixing (same as above)

src/Castle.Services.Logging.NLogIntegration/NLogLogger.cs
pi3k14: Not sure how that happend - fixing.

snakefoot:
I think the checks should be on for all levels when it avoid allocations. When logging is not enabled.


Sent from my Samsung device


-------- Original message --------

jonorossi:
@stakx I've created #408, could you move your comment there.

@pi3k14 please go with whatever the other log levels do in that same scenario.

jonorossi:
> I think the checks should be on for all levels when it avoid allocations. When logging is not enabled.

@snakefoot `LevelFilteredLogger` already guards all log methods, it appears like most others do too.

jonorossi:
Thanks @pi3k14, apologises for so much back for forward.

@stakx before we merge, what are your thoughts on versioning for releasing this, this has a breaking change to `ILogger`, feels like we should be jumping another major version. If we decide to do that there are a bunch of other deprecations I've wanted to push out which we can do with that version too.

stakx:
> what are your thoughts on versioning for releasing this, this has a breaking change to `ILogger`, feels like we should be jumping another major version. If we decide to do that there are a bunch of other deprecations I've wanted to push out which we can do with that version too.

**TL;DR:** If we release a new major version (which seems reasonable to me), we should definitely include all deprecations with it to make that release a substantial one. (And if we decided in #408 to deprecate the logging facility as well, then there would possibly be *no* release with the new `Trace` stuff being added here.)

Now for some more detail.

My initial feeling was that this wasn't a big deal and would probably justify only a minor version bump&mdash;because, just how many people these days actually implement `ILogger` and will notice the breaking change?&mdash;but on second thought, I suppose you're right. (Like I said before, I don't know the logging facility well, I don't know Windsor well, so I'm perhaps not a good judge on the impact this change might have.)

If we wanted to turn this into a non-breaking change, we could probably go down the horrible COM route and move the new `Trace`-specific members to an `ILogger2` interface that inherits `ILogger`. I anticipate that you're not a big fan of that. :laughing:

Which brings us back to a breaking change. My understanding is that you prefer major version releases to feel "substantial", so we should definitely include more than just the `ILogger` change made here. So I agree that this would be a good opportunity to get rid of things we no longer want in the public API. We could probably just remove stuff right away (instead of just marking it as `[Obsolete]`).

If we do this, perhaps we shouldn't merge this PR just yet and wait until we're getting closer to the v5 release. Because once it's merged, it might block us from releasing a further 4.x minor version if someone requests such a one. (Say we get a request to release 4.3.2 because of #406. We'd then have to create separate release branches to exclude or revert this PR here.)

Alternatively, we could subscribe to semver's rather clinical view of major version bumps, which happen whenever there's a breaking change (however substantial or small), i.e. iterate major versions much more quickly than we have in the past. I'm not a big fan of this as humans give more meaning to version numbers than semver does, releasing new major versions every few months might just make Core look unstable.

jonorossi:
> And if we decided in #408 to deprecate the logging facility as well, then there would possibly be no release with the new Trace stuff being added here.

The logging facility is in Windsor, and the the logging abstraction isn't going away that quickly. We'd have to get everything else in place before even deprecating it (can't see it going anywhere in the next 12 months). I can guarantee there is a lot of use of these interfaces out there.

> My initial feeling was that this wasn't a big deal and would probably justify only a minor version bump—because, just how many people these days actually implement ILogger and will notice the breaking change? [...]

I was actually thinking the same thing, I'd be happy with 4.4.0 for these changes, that way we'd be able to release the changes without waiting for a big bang 5.0 release.

Yer, not going down the COM-style route, i.e. ILogger2 or ILoggerEx 😆.

> If we do this, perhaps we shouldn't merge this PR just yet and wait until we're getting closer to the v5 release. Because once it's merged, it might block us from releasing a further 4.x minor version if someone requests such a one.

We can always branch off a tag, but yes I'd prefer we don't have work sitting unreleased in master so long.
 
> [...] releasing new major versions every few months might just make Core look unstable.

I've always thought the same thing especially for libraries, I know people have held back on upgrades to JSON.NET major version bumps without even looking at the changes.

My view is that you can't just change your dependencies and ship to production anyway, because any and all changes can be breaking changes even if they don't change the public API, all code changes change behaviour, and especially with Castle libraries which have sort of defaulted to everything being public the chance is even higher.

stakx:
@jonorossi:

> [...] any and all changes can be breaking changes [...]

From that perspective, choosing the version number becomes more a question of how much of a breaking change this might be. I would then say that the impact of the changes made to `ILogger` is probaly quite small, so a major version bump wouldn't be justified, and 4.4.0 would be OK.

Here's three ideas:

* Let's do 4.4.0 instead of 5.0.0 so we can release sooner and don't build up lots of unreleased changes.
* Consider being more explicit, in `CHANGELOG.md`, about this change to `ILogger` being a breaking change for those who implement that interface.
* Consider mentioning, at the very beginning of `CHANGELOG.md`, that this project uses a versioning scheme close to semver, but doesn't adhere to it in a strict fashion. (See e.g. the changelog template on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) which includes a similar preamble.) Basically a disclaimer so people don't blame us when they blindly upgrade dependencies using tools that expect strict semver.

jonorossi:
> Let's do 4.4.0 instead of 5.0.0 so we can release sooner and don't build up lots of unreleased changes.

👍, glad to read you agree.

> Consider being more explicit, in CHANGELOG.md, about this change to ILogger being a breaking change for those who implement that interface.

This is something we've done in the past quite a bit, see the "Breaking Changes" heading under 4.1.0 last year for some minor but still breaking changes.

> Consider mentioning, at the very beginning of CHANGELOG.md, that this project uses a versioning scheme close to semver, but doesn't adhere to it in a strict fashion. (See e.g. the changelog template on Keep a Changelog which includes a similar preamble.) Basically a disclaimer so people don't blame us when they blindly upgrade dependencies using tools that expect strict semver.

Interestingly I don't recall anyone complaining about us making small breaking changes in non-major versions in the past. A preamble feels a little YAGNI, if someone is taking the time to read the changelog they'll see the breaking changes section.

I guess that means we are ready to merge this and get a release out. @stakx happy?

stakx:
@jonorossi, from my point of view, this is good to go.

jonorossi:
@stakx my bad, working through my Castle notifications today. Since moving back to Gmail from "Inbox by Gmail" in September, I've found myself not actioning notifications after reading them since I was using a bundle in my inbox before so always saw them, but now they get hidden, out of sight out of mind. For now I've changed the email rule not to skip my inbox.

jonorossi:
Merged. Thanks @pi3k14, I'm normally not so slack getting pull requests merged.

PR #406, opened 2018-08-30
Make readonly parameters detection more robust

### LINKED ISSUE #398: Consider `In` modifier when testing whether parameter is read-only
Currently [when testing](https://github.com/castleproject/Core/blob/f84d4fcebdc66b859c321b2eee5434aa41df19f5/src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs#L51) a parameter for the immutability we consider the following attributes:
- Parameter attribute flags
- `IsReadOnly` attribute presence.

As it has been discussed [here](https://github.com/castleproject/Core/pull/370#issuecomment-399370350) with @stakx we should additionally test for the required modifier presence to make the logic more robust. The parameter should be considered read-only if __either__:
- parameter attributes has `ParameterAttributes.In` flag _and_ `In` required modifier is present
- parameter attributes has `ParameterAttributes.In` flag _and_ `IsReadOnly` attribute is defined

We should tests the performance and add the quicker check first.

I'd suggest to postpone the improvement till the correct modifier type is clarified here: https://github.com/dotnet/csharplang/issues/1655.

zvirja:
Closes  #398

Now we check the `In` required modifier, so the detection logic is more robust.

Additional improvements:
- exclude Rider tmp files from git
- do not store invocation arguments if there are no writable parameters (optimization).

I've tested that it's much quicker to test for the custom modifiers, than for attributes.

Test body:
```c#
[ClrJob, CoreJob]
public class MyBench
{
    private readonly ParameterInfo Parameter = typeof(ReadOnlyHolder).GetMethod(nameof(ReadOnlyHolder.Do)).GetParameters().Single();

    [Benchmark]
    public bool TestInModifier() => (Parameter.Attributes & ParameterAttributes.In) == ParameterAttributes.In;

    [Benchmark]
    public bool HasInReqMod() => Parameter.GetRequiredCustomModifiers().Any(x => x == typeof(InAttribute));

    [Benchmark]
    public bool HasIsReadOnlyAttribute() => Parameter.GetCustomAttributes(false).Any(x => x.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute");
}

public abstract class ReadOnlyHolder
{
    public abstract void Do(in int value);
}
```

Result:
```
BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i7-6700K CPU 4.00GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=3914069 Hz, Resolution=255.4886 ns, Timer=TSC
.NET Core SDK=2.1.400
  [Host] : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3132.0
  Core   : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


                 Method |  Job | Runtime |         Mean |      Error |     StdDev |
----------------------- |----- |-------- |-------------:|-----------:|-----------:|
         TestInModifier |  Clr |     Clr |     1.255 ns |  0.0093 ns |  0.0077 ns |
            HasInReqMod |  Clr |     Clr |   219.762 ns |  1.4106 ns |  1.3194 ns |
 HasIsReadOnlyAttribute |  Clr |     Clr | 1,623.902 ns | 20.4394 ns | 18.1190 ns |
         TestInModifier | Core |    Core |     1.225 ns |  0.0027 ns |  0.0021 ns |
            HasInReqMod | Core |    Core |   195.929 ns |  1.3431 ns |  1.1906 ns |
 HasIsReadOnlyAttribute | Core |    Core | 1,143.240 ns |  9.8433 ns |  8.2196 ns |
```

CC @stakx


src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -74,25 +74,23 @@ bool IsReadOnly(ParameterInfo parameter)
...
-					// The comparison by name is intentional; any assembly could define that attribute.
-					// See explanation in comment above.
-					return attribute.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute";
-				}
+                // The comparison by name is intentional; any assembly could define that attribute.
+                // See explanation in comment above.
+				if (parameter.GetCustomAttributes(false).Any(x => x.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute"))
+					return true;
+
+				return false;
```
stakx: Is there a reason&mdash;beyond personal coding style preferences&mdash;for rewriting so much of this code (e.g. rewriting `IsIsReadOnlyAttribute` as an inline lambda; removing one level of indentation by inverting the first `if` condition)?

Unless I'm mistaken, this diff here could be much smaller: You could simply insert the new check (from `#if FEATURE_EMIT_CUSTOMMODIFIERS` up to and including `#endif`) above the code comment line that reads, _"Here we perform the actual check [...]"_ (which would have to be rephrased to keep it accurate), the rest could stay as it is.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -74,25 +74,23 @@ bool IsReadOnly(ParameterInfo parameter)
...
-					// Here we perform the actual check. We don't rely on cmods because support
-					// for them is at current too unreliable in general, and because we wouldn't
-					// be saving much time anyway.
-					if (parameter.GetCustomAttributes(false).Any(IsIsReadOnlyAttribute))
-					{
-						return true;
-					}
-				}
+				if ((parameter.Attributes & (ParameterAttributes.In | ParameterAttributes.Out)) != ParameterAttributes.In)
+					return false;
```
stakx: I know it was me who originally added such a check, but I am no longer certain whether it is a terribly good idea to use `ParameterAttributes` as a fast guard against non-`in` parameters. The fact that `in` parameters are marked as `ParameterAttributes.In` isn't (AFAIK) [specified](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/readonly-ref.md#metadata-representation-of-in-parameters), after all, so perhaps we shouldn't rely on it...?

(I am not requesting a change. I just wanted to mention that I am having second thoughts here, in case anyone else feels the same way about this check.)

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
stakx: (Besides, there's some bad indentation of the code comments above the last `if`.)

.gitignore
```diff
@@ -4,6 +4,7 @@
 .vs
 _ReSharper*/
 *.csproj.ReSharper
+.idea/
```
stakx: I know single-line PRs can seem silly to some, but I would personally have preferred this to be a separate PR as it is completely unrelated to the rest of your proposed changes. (That said, it's a detail that's easily reviewed so I don't mind too much if it stays.)

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -74,25 +74,23 @@ bool IsReadOnly(ParameterInfo parameter)
...
 
-				return false;
+#if FEATURE_EMIT_CUSTOMMODIFIERS
```
stakx: (I cannot recall exactly whether `#if FEATURE_EMIT_CUSTOMMODIFIERS` is specifically for situation where custom modifiers are *emitted*, or whether it's also necessary to guard places where one just wants to *read* them. If you've already ensured that this `#if` is necessary, then all is well. Otherwise, perhaps you could quickly check this?)

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: > (Besides, there's some bad indentation of the code comments above the last if.)

I'll fix it, sorry about that. Forgot to run R#'s reformat code 😊 

> Is there a reason—beyond personal coding style preferences—for rewriting so much of this code ... ?

The main reason was readability. The less level of indentation we have, the easier it's read the code. If you find it worse - I can revert.

> e.g. rewriting IsIsReadOnlyAttribute as an inline lambda;

This is optimization. With previous implementation the delegate was never cached:
image

With new one it is:
image

> Unless I'm mistaken, this diff here could be much smaller

Well, my effort was to left the readable code after myself, rather then make as small diff as possible. However, if you find some changes redundant - I'm fine to refactor them, just point the direction 😉

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: Yep, I also noticed that. However, decided to keep as it significantly improves the performance. Let's remove it after we meet some real problem with that check 😊

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: Done, looks better. Can revert the last commit if needed.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
stakx: No, I'm fine with leaving it as you've rewritten it.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -74,25 +76,23 @@ bool IsReadOnly(ParameterInfo parameter)
...
-						return true;
-					}
-				}
+				if ((parameter.Attributes & (ParameterAttributes.In | ParameterAttributes.Out)) != ParameterAttributes.In)
+					return false;
 
-				return false;
+#if FEATURE_CUSTOMMODIFIERS
+				// Check for the required modifiers first, as it's faster.
+				// This check allows to make the detection logic more robust on the platforms which support custom modifiers.
```
stakx: One more thing. Would it perhaps be possible to spell out *why* this check makes the detection logic "more robust"? For anyone stumbling upon this code without having followed this and my previous PR, that's perhaps not evident at all.

IIRC the reason was that DynamicProxy doesn't always reproduce the `IsReadOnlyAttribute` custom attribute placed on a parameter (and thus won't recognize them as `in` parameters); specifically, in places where the Roslyn compiler emitted a `internal` copy of it into the proxied type's assembly because it's not present in the targeted BCL. In such cases, DynamicProxy would have to (but currently doesn't) emit a copy of that attribute type into its dynamic assembly so it could apply it to the proxy method parameter.

Perhaps you could add a short explanation about this in the code comment?

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
```diff
@@ -74,25 +76,23 @@ bool IsReadOnly(ParameterInfo parameter)
...
-				bool IsIsReadOnlyAttribute(object attribute)
-				{
-					// The comparison by name is intentional; any assembly could define that attribute.
-					// See explanation in comment above.
-					return attribute.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute";
-				}
+				// The comparison by name is intentional; any assembly could define that attribute.
+				// See explanation in comment above.
+				if (parameter.GetCustomAttributes(false).Any(x => x.GetType().FullName == "System.Runtime.CompilerServices.IsReadOnlyAttribute"))
+					return true;
```
jonorossi: The changes look good. Only nit, could you wrap the 3 if-statement bodies with curly braces as per our coding standard.

src/Castle.Core/DynamicProxy/Generators/GeneratorUtil.cs
zvirja: Done!

zvirja:
@stakx @jonorossi Many thanks for the code review and the valuable comments you left! Appreciate a lot  🍻 🙇 

I fixed all of them, so we should be fine to proceed further. Let me know if something left.

P.S. Sorry for the "laconic" replies and "dirty" commits - am extremely busy these days 😖

zvirja:
@jonorossi @stakx Are we waiting for somebody else's approve to merge this? Unfortunately, I don't have enough permissions to run the merge...

jonorossi:
@zvirja I didn't merge it since @stakx hadn't marked his review as complete, and since he has been more involved with these changes than myself I thought best to defer to him. Thanks for the clean set of commits.

jonorossi:
@zvirja oh, just saw the "Looks good to me!" comment by @stakx.

zvirja:
@stakx Recover soon, mate :wink: Thanks for the time spent on this.

PR #411, opened 2018-10-02
Optimize `TypeUtil.Sort`

### LINKED ISSUE #386: GC problem on Proxy Creation
I get performance issue in a programm using Proxy generation. I use perfview fo diagnostic and find out that proxy creation involve many GC string allocationns wich result in too many garbage collections.

gc_castle

So reason is TypeUtil.Sort which uses "AssemblyQualifiedName" which do a string.concat. 
Why there is a need to sort interfaces? And is there a way to to this without AssemblyQualifiedName?

stakx:
This resolves #386 through a few (admittely micro-) optimizations:

* Avoids comparing by `Type.AssemblyQualifiedName`, which performs hidden string concatenation & allocation. A quick benchmark suggests that comparing `Type.FullName` and `Type.Assembly.FullName` separately is about twice as fast.

* Using an `IComparer<Type>` instead of a `Comparison<Type>` delegate avoids unnecessary double dispatch (caused by the latter being wrapped as an `IComparer<Type>` inside the framework).

* Comparing case-sensitively instead of case-insensitively seems more appropriate for .NET type names, and should be faster.

(This might seem like too much additional code for too little gain, but consider that the C# compiler generates about the same amount of hidden code for a lambda.)

### DIFF
--- a/src/Castle.Core/DynamicProxy/Internal/TypeUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/TypeUtil.cs
@@ -246,8 +246,27 @@ namespace Castle.DynamicProxy.Internal
 			var array = new Type[types.Count];
 			types.CopyTo(array, 0);
 			//NOTE: is there a better, stable way to sort Types. We will need to revise this once we allow open generics
-			Array.Sort(array, (l, r) => string.Compare(l.AssemblyQualifiedName, r.AssemblyQualifiedName, StringComparison.OrdinalIgnoreCase));
+			Array.Sort(array, TypeNameComparer.Instance);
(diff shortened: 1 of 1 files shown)


jonorossi:
@stakx looks good, I don't think it is too much code for the performance gain/GC impact. The only thing I think would be good is to add a comment about why it exists so someone doesn't try reverting the change in the future to simplify code.

stakx:
@jonorossi - Good point. Done!

jonorossi:
@stakx is it worth updating the changelog?

stakx:
@jonorossi, I don't think so, there will likely be no noticeable difference in behavior or speed.

PR #412, opened 2018-10-11
CI: Update Mono to version 5.16 & un-skip a bunch of tests

stakx:
Just saw in the [Mono 5.16.0 release notes](https://www.mono-project.com/docs/about-mono/releases/5.16.0/) that this version should fix some bugs which caused us to skip some tests.

This PR attempts to unskip some of the tests skipped on Mono by updating the Travis CI build to Mono 5.16.0 and removing specific `[ExcludeOnFramework(Framework.Mono)]` attributes.

(This is work in progress: ~~If it won't work as expected, I'll close the PR until a later time when I'll be trying the same procedure again.~~)

* [X] Unskip as many Mono tests as possible.
* [X] Skip tests that have started failing. (This wasn't exactly planned, obviously.)
* [x] Update [`docs/dynamicproxy-optional-parameter-value-limitations.md`](https://github.com/castleproject/Core/blob/4f741f578746e5a0cbdd87da3e7d2774c877bfe1/docs/dynamicproxy-optional-parameter-value-limitations.md) where necessary
* [X] Submit new bug reports to the Mono project
* [x] ~~Update [Mono guidance in `README.md`](https://github.com/castleproject/Core/blob/4f741f578746e5a0cbdd87da3e7d2774c877bfe1/README.md#mono)~~ (did most of this work in a separate PR already)

PR #415, opened 2018-10-21
Update build instructions & Mono version guidance in the README

stakx:
This updates the README's _"Building"_ section in response to https://github.com/castleproject/Core/pull/412#issuecomment-429632176:

> It is probably also time we update the guidance about Mono version in the readme.

See it rendered in its [current version](https://github.com/castleproject/Core/blob/master/README.md#building) vs. [this PR's version](https://github.com/stakx/Castle.Core/blob/build-instructions/README.md#building).

I am suggesting several changes:

* Change the sub-headings from _".NET Framework and .NET Core"_ & _"Mono"_ to _"On Windows"_ & _"On Linux"_, respectively. This is more accurate, since the shown build commands are OS-specific, not framework-specific. (There's one build script per OS, but both of them attempt to build Castle Core for all supported frameworks.)

* Add information about tools required for the build.

* Slightly update the Mono version guidance.


stakx:
@jonorossi - True, it shouldn't be a problem but I've reverted it to `...` anyway. ;-)

PR #419, opened 2018-11-11
NLogFactory - Removed explicit loading of nlog.config by default

### LINKED ISSUE #418: Contrastive behavior of NLogIntegration against NLogs
An "FileNotFound" exception was thrown when I used NLogIntegration without nlog.config:
```
var logFactory = new NLogFactory();
var log = logFactory.Create("castle.log.Program");
log.Info("Hello world!");
```
The behavior is in contrast to NLogs. The code below will not throw exception even without nlog.config:
```
var log = NLog.LogManager.GetCurrentClassLogger();
log.Info("Hello world!");
```
Even though there is a workaround by creating factory with `new NLogFactory(true)`, but I could not found it anywhere in the documentation. This behavior also propagates to LoggingFacility in Windsor.

I suggest to change the default behavior to matching NLogs as Castle's logging is just a wrapper around NLog and should not change its default behavior, or alternatively to update the documentation highlighting the different.

The version I used:
* netcoreapp2.1
* Castle.Core 4.3.1
* Castle.Core-NLog 4.3.1

snakefoot:
Since NLog will automatically load its configuration. Resolves #418 

See also: https://github.com/nlog/NLog/wiki/Configuration-file#configuration-file-locations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Deprecations:
 
 Enhancements:
  - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
- - NLogFactory will no longer load NLog.config by default, since NLog does it automatically (@snakefoot, #419)
+ - NLogFactory and ExtendedNLogFactory default constructor will no longer load NLog.config, since NLog does it automatically (@snakefoot, #419)
 
 ## 4.3.1 (2018-06-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Deprecations:
 
 Enhancements:
  - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
+ - NLogFactory will no longer load NLog.config by default, since NLog does it automatically (@snakefoot, #419)
 
 ## 4.3.1 (2018-06-21)
 
--- a/src/Castle.Services.Logging.NLogIntegration/NLogFactory.cs
+++ b/src/Castle.Services.Logging.NLogIntegration/NLogFactory.cs
@@ -32,14 +32,14 @@ namespace Castle.Services.Logging.NLogIntegration
 		///   Initializes a new instance of the <see cref="NLogFactory" /> class.
 		/// </summary>
 		public NLogFactory()
-			: this(defaultConfigFileName)
+			: this(true)
 		{
 		}
 
 		/// <summary>
 		///   Initializes a new instance of the <see cref="NLogFactory" /> class.
 		/// </summary>
-		/// <param name="configuredExternally">If <c>true</c>. Skips the initialization of log4net assuming it will happen externally. Useful if you're using another framework that wants to take over configuration of NLog.</param>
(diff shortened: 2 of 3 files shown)


snakefoot:
@stakx Have now update the ChangeLog.md

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Deprecations:
 
 Enhancements:
  - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
+ - NLogFactory will no longer load NLog.config by default, since NLog does it automatically (@snakefoot, #419)
```
stakx: Perhaps this could be a tiny bit more specific:

```suggestion
 - NLogFactory's default constructor will no longer load NLog.config by default, since NLog does it automatically (@snakefoot, #419)
```

(And in case that the same change is applied to `ExtendedNLogFactory`&mdash;see my other review comment&mdash;that other class' default ctor should be mentioned here as well.)

src/Castle.Services.Logging.NLogIntegration/NLogFactory.cs
```diff
@@ -32,14 +32,14 @@ public class NLogFactory : AbstractLoggerFactory
 		///   Initializes a new instance of the <see cref="NLogFactory" /> class.
 		/// </summary>
 		public NLogFactory()
-			: this(defaultConfigFileName)
+			: this(true)
```
stakx: I've noticed that the `ExtendedNLogFactory` has very similar constructors as `NLogFactory`. Should the same change be applied there, too, for consistency?

snakefoot: Good catch. Have now also updated `ExtendedNLogFactory`

CHANGELOG.md
snakefoot: Good catch. Have now also updated `ExtendedNLogFactory` and added to ChangeLog.md

devop228:
I look at the constructor overload `NLogFactory(bool configuredExternally)`, its existence seems redundant since it leads to almost same behavior on both path of the code. Would removing it completely be too much a breaking change?

snakefoot:
You are welcome to create your own PR

Sent from my Sony Xperia


---- DevOp228 wrote ----


I look at the constructor overload NLogFactory(bool configuredExternally), its existence seems redundant since it leads to almost same behavior on both path of the code. Would removing it completely be too much a breaking change?

jonorossi:
> I look at the constructor overload `NLogFactory(bool configuredExternally)`, its existence seems redundant since it leads to almost same behavior on both path of the code. Would removing it completely be too much a breaking change?

@devop228 that is what I proposed in #418 before this pull request was created.

Since Windsor's logging facility only has an `ConfiguredExternally` method, you cannot get Windsor to pass `false` unless you use the custom factory overload.

I'm going to close this pull request because I think just doing this makes thing inconsistent between log4net and NLog, and doesn't push us to where we want to be with the factory's being just a shim, and not setting up configuration.

PR #421, opened 2018-12-09
Resolve ambiguity with standard CollectionExtensions

tomaszstrejczek:
In some build scenarios the name CollectionExtensions clashes with the standard one.

### DIFF
--- a/src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
+++ b/src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
@@ -187,7 +187,7 @@ namespace Castle.DynamicProxy
 			{
 				return false;
 			}
-			if (!CollectionExtensions.AreEquivalent(AdditionalAttributes, proxyGenerationOptions.AdditionalAttributes))
+			if (!Castle.Core.Internal.CollectionExtensions.AreEquivalent(AdditionalAttributes, proxyGenerationOptions.AdditionalAttributes))
 			{
 				return false;
 			}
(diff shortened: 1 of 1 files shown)


stakx:
I haven't noticed any problems while compiling this code until now, so from my current point of view, this change shouldn't be necessary. Under what conditions is there any ambiguity here?

stakx:
@tomaszstrejczek -

> when compiling against .net core 2.0 and referencing System.Collections.dll

I'm not entirely clear on what you mean by that. Castle.Core does not target `netcoreapp2.0`. Were you trying to build a modified version of Castle.Core, or were you referencing Castle.Core's (unmodified) source project from a .NET Core 2.0 project that also references the System.Collections package?

tomaszstrejczek:
I tried to build castle.core with .net core 2.0 (targeting netcoreapp2.0). Unfortunately the build system I use, adds the reference to System.Collections.dll

stakx:
@tomaszstrejczek - This project comes with its own build scripts (see the ["Building" section in the README](https://github.com/castleproject/Core/blob/ed47d036193a19465286f1703a6963390dd62445/README.md#building)). If they do not work, or could be improved, please open an issue and describe what problem(s) you encountered with them.

Other ways of building are unsupported. There is no clear benefit to maintaining more than one build system, but there are real costs (e.g. keeping all of them up to date and in sync). The project is therefore likely to benefit more if our limited resources are invested elsewhere.

For this reason, I feel it would be wrong to accept a change that only benefits another build process but is otherwise unnecessary&mdash;even if the proposed change is completely benign.

If you must use a different build system, you could fork the project and make any required changes there.

PR #425, opened 2018-12-29
Add hyperlinks to issues in CHANGELOG.md

jcsonder:
Convert plain issue numbers in CHANGELOG.md to hyperlinks so the issue will open with one click.


jcsonder:
It's accomplished the same way like e.g. here: https://github.com/ninject/Ninject/blob/master/CHANGELOG.md

PR #427, opened 2019-01-05
Support locals in FromExpression

### LINKED ISSUE #426: Extend the set of expressions supported by CustomAttributeInfo.FromExpression
Prompted by https://stackoverflow.com/q/54044725/98713

Currently the expression analyzer only supports constants and arrays as the constructor arguments. This is enough in most scenarios, but not if the argument values need to be computed at runtime. Supporting any arbitrary expression would be difficult (or would involve compiling and executing expressions, which is slow), but it should at least be possible to use local variables.

thomaslevesque:
Fixes #426

### DIFF
--- a/src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
+++ b/src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
@@ -20,6 +20,7 @@ namespace Castle.DynamicProxy
 	using System.Linq.Expressions;
 	using System.Reflection;
 	using System.Reflection.Emit;
+	using System.Runtime.CompilerServices;
 
 	/// <summary>
 	/// Encapsulates the information needed to build an attribute.
@@ -169,14 +170,26 @@ namespace Castle.DynamicProxy
 
 		private static object GetAttributeArgumentValue(Expression arg, bool allowArray)
 		{
-			var constant = arg as ConstantExpression;
-			if (constant == null)
+			switch (arg.NodeType)
 			{
-				if (allowArray)
-				{
-					var newArrayExpr = arg as NewArrayExpression;
-					if (newArrayExpr != null)
+				case ExpressionType.Constant:
+					return ((ConstantExpression)arg).Value;
+				case ExpressionType.MemberAccess:
+					var memberExpr = (MemberExpression) arg;
+					if (memberExpr.Member is FieldInfo field)
 					{
+						if (memberExpr.Expression is ConstantExpression constant &&
+						    IsCompilerGenerated(constant.Type) &&
+						    constant.Value != null)
+						{
+							return field.GetValue(constant.Value);
+						}
+					}
+					break;
+				case ExpressionType.NewArrayInit:
+					if (allowArray)
+					{
+						var newArrayExpr = (NewArrayExpression) arg;
 						var array = Array.CreateInstance(newArrayExpr.Type.GetElementType(), newArrayExpr.Expressions.Count);
 						int index = 0;
 						foreach (var expr in newArrayExpr.Expressions)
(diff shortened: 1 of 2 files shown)


src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -169,28 +170,44 @@ private static Expression UnwrapBody(Expression body)
 
 		private static object GetAttributeArgumentValue(Expression arg, bool allowArray)
 		{
-			var constant = arg as ConstantExpression;
-			if (constant == null)
+			if (arg is ConstantExpression constant)
```
stakx: I noticed that you are repeatedly checking `arg`'s type. That's fine and it enables the nice `is T x` syntax, but if you're willing to micro-optimize, you might get a wee bit better performance by querying & caching `arg.NodeType` just once, then compare the various `ExpressionType` enum values against that (e.g. using a `switch` block).

thomaslevesque: I pushed a fixup commit, let me know if you like it better

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -187,10 +200,15 @@ private static object GetAttributeArgumentValue(Expression arg, bool allowArray)
 						}
 						return array;
 					}
-				}
-				throw new ArgumentException("Only constant and single-dimensional array expressions are supported");
+					break;
 			}
-			return constant.Value;
+			
+			throw new ArgumentException("Only constant and single-dimensional array expressions are supported");
```
blairconrad: This message is now a lie. We support expressions that refer to values known only at runtime.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
```diff
@@ -169,14 +170,26 @@ private static Expression UnwrapBody(Expression body)
...
-					var newArrayExpr = arg as NewArrayExpression;
-					if (newArrayExpr != null)
+				case ExpressionType.Constant:
+					return ((ConstantExpression)arg).Value;
+				case ExpressionType.MemberAccess:
+					var memberExpr = (MemberExpression) arg;
+					if (memberExpr.Member is FieldInfo field)
+					{
+						if (memberExpr.Expression is ConstantExpression constant &&
+						    IsCompilerGenerated(constant.Type) &&
```
blairconrad: What's the benefit of checking this? I understand that it will catch some cases in which the values aren't of the proper type, but would those cases not be caught when we try to create the attribute anyhow?
Or even be caught by the compiler because we shouldn't have a constructor that accepts the type?
I am especially critical because we're micro-optimizing for the `NodeType` check, as @stakx asked for (and I do like that), but the `IsDefined` call inside `IsCompilerGenerated` is very expensive.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
thomaslevesque: Yes, I was just thinking about this. I'll update it.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
thomaslevesque: > What's the benefit of checking this? I understand that it will catch some cases in which the values aren't of the proper type

No, it checks that the field is a member of a closure type, so that only locals can be accessed like this, not other fields. But maybe it's not necessary. @stakx what do you think?

blairconrad: > is a member of a closure type

Oh! Thanks for explaining.

src/Castle.Core/DynamicProxy/CustomAttributeInfo.cs
blairconrad: Thanks. I like the new message.

stakx:
@thomaslevesque - I think all is good here. But we still have "WIP" in this PR's title, so I thought I'd say this: In case you'd still like to clean up anything, now would be a good time. Otherwise I'll merge this in a day or two.

thomaslevesque:
I just needed to squash the fixup commits. It's done now, you can merge whenever you want.

PR #428, opened 2019-01-13
Fix async problem calling await before Proceed - Breaking Change

brunoblank:
This PR solved the async problem discussed in #145

Breaking Change:
`void IInvocation.Proceed()` is moved to `void IInterceptor.Intercept(IInvocation invocation, InvocationDelegate proceed);`

Before:
 ```csharp
public class Interceptor : IInterceptor
{
    {
        invocation.Proceed();
    }
}
```

Becomes:
 ```csharp
public class Interceptor : IInterceptor
{
    public void Intercept(IInvocation invocation, InvocationDelegate proceed)
    {
        proceed(invocation);
    }
}
```

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/AsyncInterceptorTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/AsyncInterceptorTestCase.cs
@@ -1,0 +1,44 @@
+﻿// Copyright 2004-2016 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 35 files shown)


stakx:
@brunoblank - In case that we *do* decide to take this further, I think it would be great if you could cherry-pick or otherwise incorporate @JSkimming's unit test from #429 into your PR.

stakx:
#439 just got merged, so I am closing this PR as it was meant to enable the same use case.

Thank you @brunoblank for having figured out how to solve this long-standing problem with `Proceed`! :bowing_man:

brunoblank:
Awesome! Glad I could help out finding a solution for the annoying async problem.

/Bruno

PR #429, opened 2019-01-16
Added tests to exercise intercepting an async method

JSkimming:
I'm re-raising this PR originally raised as #337 to present the problem.

I'm going to then merge the changes from @brunoblank in #428 and re-factor given the breaking changes to see if it fixes the issue (spoiler alert, it does on my machine 😃)

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/AsyncInterceptorTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/AsyncInterceptorTestCase.cs
@@ -1,0 +1,48 @@
+﻿// Copyright 2004-2016 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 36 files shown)


JSkimming:
As shown by the passing builds, the changes introduced by PR #428 address the problem I outlined in issue #145 in specifically [this comment](https://github.com/castleproject/Core/issues/145#issuecomment-365404690).

stakx:
@JSkimming - Apologies for the long silence from the Castle team. I've been wanting to take a careful look at your PR and the associated issue for a long time, but have been very preoccupied with work. This is just a small ping to show that this hasn't been forgotten. I'm planning to get back to work here at this repo soon... but it might take me another few weeks.

stakx:
@JSkimming - Thank you for adding a unit test to demonstrate a problem that #428 would solve. :+1:

However we now have two PRs that propose the same basic change to DynamicProxy's API; I've posted a review comment over in #428 that would also apply here. In order to not spread discussion over too many separate pages and end up with excessive cross-referencing, I propose that we discuss this change over in #428 or, if we stray too far from it, go back to the issue (#145). Is this OK with you?

JSkimming:
@stakx of course makes perfect sense.

PR #432, opened 2019-03-05
Add InterceptorBase

isdaniel:
Add an abstract class and implement Interceptor interface, other class can use it like Asp.net MVC action filter.

### DIFF
--- a/src/Castle.Core/DynamicProxy/IInterceptorBase.cs
+++ b/src/Castle.Core/DynamicProxy/IInterceptorBase.cs
@@ -1,0 +1,56 @@
+﻿
+namespace Castle.DynamicProxy
+{
+	using System;
+
+	public class ExceptionContext
+	{
+
+		public Exception Ex { get; set; }
+		public IInvocation Invocation { get; set; }
+	}
+
+	public abstract class InterceptorBase : IInterceptor
+	{
+		protected virtual void OnExcuting(IInvocation invocation)
+		{
+		}
+
+		protected virtual void OnExcuted(IInvocation invocation)
+		{
+		}
+
+		protected virtual void OnException(ExceptionContext context)
+		{
+		}
+
+		private bool NotVoidHasValue(IInvocation invocation)
+		{
+			return !(invocation.Method.ReturnType == typeof(void) ||
+			         invocation.ReturnValue == null);
+		}
+
+		public void Intercept(IInvocation invocation)
+		{
+			try
+			{
+				OnExcuting(invocation);
+
+				if (NotVoidHasValue(invocation))
+					return;
+
+				invocation.Proceed();
+
+				OnExcuting(invocation);
+			}
+			catch (Exception ex)
+			{
+				OnException(new ExceptionContext()
+				{
+					Ex = ex,
+					Invocation = invocation
+				});
+			}
+		}
+	}
+}


src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
+﻿
```
stakx: Copyright notice is missing, see other source code files.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Castle.DynamicProxy
+{
+	using System;
+
+	public class ExceptionContext
+	{
+
```
stakx: Superfluous blank line, should be removed.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+		public Exception Ex { get; set; }
+		public IInvocation Invocation { get; set; }
+	}
+
+	public abstract class InterceptorBase : IInterceptor
+	{
+		protected virtual void OnExcuting(IInvocation invocation)
```
stakx: `OnExcuting` &rarr; `OnExecuting`.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+		{
+		}
+
+		protected virtual void OnExcuted(IInvocation invocation)
```
stakx: `OnExcuted` &rarr; `OnExecuted`.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		protected virtual void OnException(ExceptionContext context)
+		{
+		}
+
+		private bool NotVoidHasValue(IInvocation invocation)
+		{
+			return !(invocation.Method.ReturnType == typeof(void) ||
+			         invocation.ReturnValue == null);
```
stakx: What if a non-`void` method *wants to* return `null`? It's a perfectly valid value, after all.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+		public void Intercept(IInvocation invocation)
+		{
+			try
+			{
+				OnExcuting(invocation);
+
+				if (NotVoidHasValue(invocation))
+					return;
+
+				invocation.Proceed();
```
stakx: I don't think that this pattern&mdash;`Proceed` if the interceptor hasn't yet provided a return value&mdash;holds true for all use cases (though admittedly for a good many). Sometimes there might be nothing to proceed *to*. You'd at least have to document your interceptor's behavior / functional semantics.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+		{
+			return !(invocation.Method.ReturnType == typeof(void) ||
+			         invocation.ReturnValue == null);
+		}
+
+		public void Intercept(IInvocation invocation)
+		{
+			try
+			{
+				OnExcuting(invocation);
```
stakx: `OnExcuting` &rarr; `OnExecuting`.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+
+				if (NotVoidHasValue(invocation))
+					return;
+
+				invocation.Proceed();
+
+				OnExcuting(invocation);
```
stakx: `OnExcuting` &rarr; `OnExecuted`.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
...
+	using System;
+
+	public class ExceptionContext
+	{
+
+		public Exception Ex { get; set; }
+		public IInvocation Invocation { get; set; }
+	}
+
+	public abstract class InterceptorBase : IInterceptor
```
stakx: I find your interceptor to actually have a fairly specific logic (which unfortunately isn't documented at all e.g. in XML doc comments), and for that this name seems to generic. And there's my main issue with this PR: The interceptor seems at the same time too specific for a low-level infrastructure library like DynamicProxy, and short enough to be easily written in your own application code.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
```diff
@@ -0,0 +1,56 @@
+﻿
+namespace Castle.DynamicProxy
+{
+	using System;
+
+	public class ExceptionContext
```
stakx: This should be `sealed` IMO (as there's no reason why someone outside the library should be able to instantiate this type), or even better, a `readonly struct` (as it's essentially a 2-tuple).

Also, documentation never hurts.

src/Castle.Core/DynamicProxy/IInterceptorBase.cs
isdaniel: Hi @stakx 
In my opinion
`invocation.ReturnValue` initial is `null`, developer set `invocation.ReturnValue`  that mean we don't want to execute this method (like do authorize).

thank you.

isdaniel:
@jonorossi @stakx  Thank for your comment and helpful suggestions.
I benefited a lot from your advice.

PR #433, opened 2019-03-12
Replaced ContainsKey and getting by key with TryGetValue

DamirAinullin:
Replaced Dictionary method ContainsKey and getting value by key with TryGetValue in order to increase perfomance.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
@@ -273,8 +273,7 @@ namespace Castle.DynamicProxy.Generators.Emitters
 
 		public Type GetGenericArgument(String genericArgumentName)
 		{
-			GenericTypeParameterBuilder genericTypeParameterBuilder;
-			if (name2GenericType.TryGetValue(genericArgumentName, out genericTypeParameterBuilder))
+			if (name2GenericType.TryGetValue(genericArgumentName, out var genericTypeParameterBuilder))
 				return genericTypeParameterBuilder.AsType();
 
 			return null;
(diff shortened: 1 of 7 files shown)


src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
```diff
@@ -273,8 +273,9 @@ public FieldReference GetField(string name)
 
 		public Type GetGenericArgument(String genericArgumentName)
 		{
-			if (name2GenericType.ContainsKey(genericArgumentName))
-				return name2GenericType[genericArgumentName].AsType();
+			GenericTypeParameterBuilder genericTypeParameterBuilder;
+			if (name2GenericType.TryGetValue(genericArgumentName, out genericTypeParameterBuilder))
```
stakx: I'd suggest using `out var` instead to keep this code shorter:

```diff
-GenericTypeParameterBuilder genericTypeParameterBuilder;
-if (name2GenericType.TryGetValue(genericArgumentName, out genericTypeParameterBuilder))
+if (name2GenericType.TryGetValue(genericArgumentName, out var genericTypeParameterBuilder))
```

src/Castle.Core/DynamicProxy/Generators/Emitters/StrongNameUtil.cs
```diff
@@ -53,12 +53,13 @@ public static bool IsAssemblySigned(this Assembly assembly)
 		{
 			lock (lockObject)
 			{
-				if (signedAssemblyCache.ContainsKey(assembly) == false)
+				bool isSigned;
+				if (signedAssemblyCache.TryGetValue(assembly, out isSigned) == false)
 				{
```
stakx: As above, you could abbreviate these two lines by using `out var`.

stakx:
Thank you for getting rid of these unnecessary double calls!

PR #436, opened 2019-03-13
Add ability to mix in delegate (types) to proxies

### LINKED ISSUE #399: Do we need a thread-safe alternative to `ModuleScope.DefineType`?

stakx:
This is a follow-up to #403, which started out with the goal of adding a set of  `ProxyGenerator.CreateDelegateProxy` methods, and ended with us thinking about achieving delegate proxies through DynamicProxy's mixin capabilities. I am closing the earlier PR in favour of this present one.

This PR adds the following capabilities to DynamicProxy:

 * **`ProxyGenerationOptions.AddDelegateMixin(Delegate @delegate)`**  
   to add a delegate mixin. This will produce an `Invoke` method on the proxy that has the same signature as the delegate. The delegate can be `Proceed`-ed to.

 * **`ProxyGenerationOptions.AddDelegateTypeMixin(Type delegateType)`**  
   adds an `Invoke` method on the proxy that has the same signature as a specified delegate type. When those `Invoke` methods are intercepted, they cannot be proceeded from as that mixin has no target.

 * **`ProxyUtil.TryCreateDelegateToMixin(object target, Type delegateType, out Delegate @delegate)`**  
   **`ProxyUtil.TryCreateDelegateToMixin<TDelegate>(object target, out TDelegate @delegate)`**  
   are helper methods to create a delegate of a given type to a suitable `Invoke` method on a proxy.

By combining the above methods, you can now create a delegate proxy:

```csharp
options.AddDelegateTypeMixin(typeof(Action));

var _ = generator.CreateClassProxy(typeof(object), options, ...);

ProxyUtil.TryCreateDelegateToMixin(_, out Action proxy);
                                      ^^^^^^^^^^^^^^^^
```

Because you're creating a class (or interface) proxy, you have full control over the "container" type for the generated interceptable `Invoke` method. You're also free to add custom attributes to that type, as you normally would.

I'd still like to...

* [X] ~~run some tests whether these new capabilities are actually useful in consuming libraries (Moq in my case)~~

   See [feedback for Moq](https://github.com/castleproject/Core/pull/436#issuecomment-472966476) and [NSubstitute](https://github.com/castleproject/Core/pull/436#issuecomment-473072088) below. Feedback is better than anticipated.

* [X] ~~I am also still unsure whether there may be serialization issues when using lambda delegates as mixin targets... and whether that is relevant in practice.~~

   ⚠️ If you use a lambda that captures any variables as a delegate mixin, then you won't be able to save the generated dynamic assembly to disk because the captured variables are put in a non-serializable "display class" by the compiler.

   This can be worked around by rewriting such lambdas & captures as explicit types. Beyond that, I'm not yet sure this is even fixable; if so, then we'd likely have to let `ProxyGenerationOptions` implement `ISerializable` and serialize it differently (specifically the `mixins` field).

   For now, I'd like to treat this as a known current limitation of the delegate mixin story, as it's probably not an immediate roadblock for most downstream libs.

Closes #399 (where we said that noone except DP itself should emit into its dynamic assembly, and the only known roadblock to that is delegate proxying; so if we get that, we no longer need `ModuleScope.DefineType`).

/cc @blairconrad, @thomaslevesque, @zvirja

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/DelegateMixinTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/DelegateMixinTestCase.cs
@@ -1,0 +1,300 @@
+// Copyright 2004-2010 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests
+{
+	using System;
+	using System.Reflection;
+
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class DelegateMixinTestCase  : BasePEVerifyTestCase
+	{
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateTypeMixin_when_given_null_throws_ArgumentNullException()
+		{
+			var options = new ProxyGenerationOptions();
+			Assert.Throws<ArgumentNullException>(() => options.AddDelegateTypeMixin(null));
+		}
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateTypeMixin_when_given_non_delegate_type_throws_ArgumentException()
+		{
+			var options = new ProxyGenerationOptions();
+			Assert.Throws<ArgumentException>(() => options.AddDelegateTypeMixin(typeof(Exception)));
+		}
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateTypeMixin_when_given_delegate_type_succeeds()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+		}
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateMixin_when_given_null_throws_ArgumentNullException()
+		{
+			var options = new ProxyGenerationOptions();
+			Assert.Throws<ArgumentNullException>(() => options.AddDelegateMixin(null));
+		}
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateMixin_when_given_delegate_succeeds()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateMixin(new Action(() => { }));
+		}
+
+
+		[Test]
+		public void ProxyGenerator_CreateClassProxy_can_create_delegate_proxy_without_target()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+			var _ = new Interceptor();
+			var proxy = generator.CreateClassProxy(typeof(object), options, _);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithoutTarget_can_create_delegate_proxy_without_target()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+			var _ = new Interceptor();
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(typeof(IComparable), options, _);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithTarget_can_create_delegate_proxy_without_target()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+			var _ = new Interceptor();
+			var proxy = generator.CreateInterfaceProxyWithTarget(typeof(IComparable), target, options, _);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateClassProxy_can_create_callable_delegate_proxy_without_target()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor();
+
+			var proxy = generator.CreateClassProxy(typeof(object), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.AreSame(typeof(Action).GetTypeInfo().GetMethod("Invoke"), interceptor.LastInvocation.Method);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithoutTarget_can_create_callable_delegate_proxy_without_target()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor();
+
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(typeof(IComparable), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.AreSame(typeof(Action).GetTypeInfo().GetMethod("Invoke"), interceptor.LastInvocation.Method);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithTarget_can_create_callable_delegate_proxy_without_target()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor();
+
+			var proxy = generator.CreateInterfaceProxyWithTarget(typeof(IComparable), target, options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.AreSame(typeof(Action).GetTypeInfo().GetMethod("Invoke"), interceptor.LastInvocation.Method);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateClassProxy_cannot_proceed_to_delegate_type_mixin()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateClassProxy(typeof(object), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			Assert.Throws<NotImplementedException>(() => action.Invoke());
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithoutTarget_cannot_proceed_to_delegate_type_mixin()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(typeof(IComparable), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			Assert.Throws<NotImplementedException>(() => action.Invoke());
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithTarget_cannot_proceed_to_delegate_type_mixin()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateInterfaceProxyWithTarget(typeof(IComparable), target, options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			Assert.Throws<NotImplementedException>(() => action.Invoke());
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateClassProxy_can_proceed_to_delegate_mixin()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateMixin(new Action(target.Method));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateClassProxy(typeof(object), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.True(target.MethodInvoked);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithoutTarget_can_proceed_to_delegate_mixin()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateMixin(new Action(target.Method));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateInterfaceProxyWithoutTarget(typeof(IComparable), options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.True(target.MethodInvoked);
+		}
+
+		[Test]
+		public void ProxyGenerator_CreateInterfaceProxyWithTarget_can_proceed_to_delegate_mixin()
+		{
+			var target = new Target();
+
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateMixin(new Action(target.Method));
+
+			var interceptor = new Interceptor(shouldProceed: true);
+
+			var proxy = generator.CreateInterfaceProxyWithTarget(typeof(IComparable), target, options, interceptor);
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+
+			action.Invoke();
+			Assert.True(target.MethodInvoked);
+		}
+
+		[Test]
+		public void Can_mixin_several_different_delegate_types_simultaneously()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+			options.AddDelegateTypeMixin(typeof(Action<int>));
+
+			var interceptor = new Interceptor();
+
+			var proxy = generator.CreateClassProxy(typeof(object), options, interceptor);
+
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+			action.Invoke();
+
+			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action<int> intAction));
+			intAction.Invoke(42);
+		}
+
+		[Test]
+		public void Cannot_mixin_several_delegate_types_with_same_signature()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Func<Exception, bool>));
+			options.AddDelegateTypeMixin(typeof(Predicate<Exception>));
+			Assert.Throws<InvalidMixinConfigurationException>(() => options.Initialize());
+
+		}
+
+		[Serializable]
+		public sealed class Target : IComparable
+		{
+			public bool CompareToInvoked { get; set; }
+
+			public bool MethodInvoked { get; set; }
+
+			public int CompareTo(object obj)
+			{
+				CompareToInvoked = true;
+				return 123;
+			}
+
+			public void Method()
+			{
+				MethodInvoked = true;
+			}
+		}
+
+		private sealed class Interceptor : IInterceptor
+		{
+			private readonly bool shouldProceed;
+
+			public Interceptor(bool shouldProceed = false)
+			{
+				this.shouldProceed = shouldProceed;
+			}
+
+			public IInvocation LastInvocation { get; set; }
+
+			public void Intercept(IInvocation invocation)
+			{
+				LastInvocation = invocation;
+				if (shouldProceed)
+				{
+					invocation.Proceed();
+				}
+			}
+		}
+	}
+}
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/DelegateMixinTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/DelegateMixinTestCase.cs
@@ -96,7 +96,8 @@ namespace Castle.DynamicProxy.Tests
 			var interceptor = new Interceptor();
 
 			var proxy = generator.CreateClassProxy(typeof(object), options, interceptor);
-			Assert.True(ProxyUtil.TryCreateDelegateToMixin(proxy, out Action action));
+			var action = ProxyUtil.CreateDelegateToMixin<Action>(proxy);
+			Assert.NotNull(action);
 
 			action.Invoke();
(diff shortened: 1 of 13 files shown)


src/Castle.Core/DynamicProxy/Contributors/DelegateTypeMembersCollector.cs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Contributors
+{
+	using System;
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Generators;
+
+	public class DelegateTypeMembersCollector : MembersCollector
```
stakx: I shouldn't make this `public`. No need to expose even more internals than we already do.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -26,11 +26,83 @@ namespace Castle.DynamicProxy
...
+		/// </summary>
+		/// <param name="proxy">The proxy instance to which the delegate should be bound.</param>
+		/// <typeparam name="TDelegate">The type of delegate that should be created.</typeparam>
+		/// <param name="delegate">The produced delegate if the method succeeds; otherwise <see langword="null"/>.</param>
+		/// <returns><see langword="true"/> if the method succeeds; otherwise <see langword="false"/>.</returns>
+		public static bool TryCreateDelegateToMixin<TDelegate>(object proxy, out TDelegate @delegate)
+		{
+			if (TryCreateDelegateToMixin(proxy, typeof(TDelegate), out var d))
+			{
+				@delegate = (TDelegate)(object)d;
```
stakx: At some point (hopefully with the next major version of Castle.Core), we would want to add a generic type constraint `where TDelegate : MulticastDelegate` to the method declaration to make this kind of casting redundant. For the moment, I've opted not to because the Roslyn compiler version shipping with Mono might be too old to support this. Even if it did, we'd probably need to add a few more of these constraints in some other places; perhaps better to do it all in one go.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -26,11 +26,83 @@ namespace Castle.DynamicProxy
...
+				throw new ArgumentException("Type is not a delegate type.", nameof(delegateType));
+			}
+
+			var invokeMethod = delegateType.GetMethod("Invoke");
+			var proxiedInvokeMethod =
+				proxy
+				.GetType()
+				.GetMember("Invoke", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+				.Cast<MethodInfo>()
+				.FirstOrDefault(m => MethodSignatureComparer.Instance.EqualParameters(m, invokeMethod));
```
stakx: In the future, we *might* conceivably want to cache the results of these queries.

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -123,19 +125,69 @@ public MixinData MixinData
...
+		}
+
+		/// <summary>
+		///   Adds a delegate to be mixed into generated proxies. The <paramref name="delegate"/>
+		///   will act as the target for calls to a `Invoke` method with a signature matching that
+		///   of the delegate.
+		/// </summary>
+		/// <param name="delegate">The delegate that should act as the target for calls to `Invoke` methods with a matching signature.</param>
+		/// <exception cref="ArgumentNullException"><paramref name="delegate"/> is <see langword="null"/>.</exception>
+		public void AddDelegateMixin(MulticastDelegate @delegate)
```
stakx: @jonorossi, what do you think, should the parameter type be `MulticastDelegate`, or `Delegate`? I'm using the one here, and the other over in `ProxyUtil`.

zvirja:
Hi @stakx!

The way you decided to sort it out looks promising and elegant!

I’ll definitely give it a more precise look. In the meanwhile, I want to clarify whether the following feature is supported (found the concern [in my old comment](https://github.com/castleproject/Core/pull/403#issuecomment-412677935)):

> I've just reviewed the NSubstitute lib and the usages we have and found that new API lucks a few features:
> - ...
> - ability to emit custom attributes to the generated method. We need this to quickly get the original delegate type from the methodInfo instance we internally store for the each recorded call. Basically, we use this attribute also to understand that methodInfo is a delegate proxy, rather than a method from a class proxy, so we render the client error messages more precisely.

Thanks! :)

stakx:
@zvirja, nice to hear from you again! 

> ability to emit custom attributes to the generated method

I can't remember all the details from the previous issue, but basically this PR follows @jonorossi's proposal made in https://github.com/castleproject/Core/pull/403#issuecomment-422787035, where he also wrote:

> It doesn't allow you to emit custom attributes on the proxy's methods (which sounded like a nice to have, not a requirement), but it does still allow custom attributes on the proxy type [...]

That pretty much sums it up. DynamicProxy didn't have the ability to add custom attributes to single methods (AFAIK) and that hasn't changed. If that's needed, perhaps we'd need to look at it in a separate PR.

stakx:
Note to self: `MixinData` equality comparison is broken when delegate type mixins are present. This will need some more work:

https://github.com/castleproject/Core/blob/a53d6b6190cde9d8b2f4e872078f769968c0ff6c/src/Castle.Core/DynamicProxy/MixinData.cs#L91-L92

https://github.com/castleproject/Core/blob/a53d6b6190cde9d8b2f4e872078f769968c0ff6c/src/Castle.Core/DynamicProxy/MixinData.cs#L112-L115

(`mixinsImpl[i]` can be null for delegate type mixins.)

zvirja:
@stakx Thanks for the replies. It started to ring a bell in my head now.

Tried to give it a go, but it fails with the following error:
```
Castle.DynamicProxy.InvalidMixinConfigurationException : There is a problem with the mixins added to this ProxyGenerationOptions: The list of mixins contains at least two delegate mixins for the same delegate signature.
Parameter name: mixinInstances
  ----> System.ArgumentException : The list of mixins contains at least two delegate mixins for the same delegate signature.
Parameter name: mixinInstances
   at Castle.DynamicProxy.ProxyGenerationOptions.Initialize()
...
```

In our code I also mix in an additional object. It might be related to the finding you already described. Please ping me when it's ready, so I can give it a try 😉

From the first glance I could say that API looks perfect from the client perspective and should cover all the needs we have 👍 

One more question (warning, might be stupid): are you sure that after all the changes you applied the container generated type is still cached?

stakx:
> it fails with the following error

I would need to see your code, for the moment I'll make a blind guess: Did you perhaps try to reuse the same `ProxyGenerationOptions` for different delegate proxy types? I suspect that you might end up with the same usage pattern as I did: You need at least one distinct `ProxyGenerationOptions` instance per mocked delegate type.

> are you sure that after all the changes you applied the container generated type is still cached?

I ran some quick tests this morning. Basically, caching should work as before if equality is correctly implemented on both `ProxyGenerationOptions` (and `MixinData`, which is used by the former), I've added some additional unit tests for that. You can test for yourself by attaching e.g. a `ConsoleLogger` to your `ProxyGenerator`. When you ask for a proxy, you should see a log message stating whether a cached type was found (and used) or not.

stakx:
@zvirja, there was a glaringly obvious bug that I didn't notice anymore at 2am last night. 😁 Fixing it also fixed all NSubstitute unit tests except one (`When_checking_across_multiple_subs_including_delegates`, which appears to fail because of differently named types).

stakx:
Speaking for Moq, I am very satisfied with how well this change would work out: https://github.com/stakx/moq4/commit/50fd9fa53e4b3af2e431e7fdbddce8f57b034c04 (14 insertions, 122 deletions). Almost all of the delegate special-casing code is gone, and gets replaced by the very short `ProxyUtil.TryCreateDelegateToMixin` bit.

zvirja:
@stakx After working a bit more on our code I managed to finally get it work. Because `MethodInfo` passed to the invocation is actually taken from the delegate type now, I don't even need to emit custom attributes  - simply use `methodInfo.DeclaringType` to get original delegate type 😊 I really enjoy the way it goes! Thanks a lot! 🤝 

The only issue I discovered is that delegate proxy activation degraded in terms of performance:

```
BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-8850H CPU 2.60GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
Frequency=2531252 Hz, Resolution=395.0614 ns, Timer=TSC
.NET Core SDK=2.1.504
  [Host] : .NET Core 2.1.8 (CoreCLR 4.6.27317.03, CoreFX 4.6.27317.03), 64bit RyuJIT
  Clr    : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3324.0
  Core   : .NET Core 2.1.8 (CoreCLR 4.6.27317.03, CoreFX 4.6.27317.03), 64bit RyuJIT

BEFORE

                           Method |  Job | Runtime |      Mean |     Error |    StdDev |
--------------------------------- |----- |-------- |----------:|----------:|----------:|
         CreateDelegateSubstitute |  Clr |     Clr | 10.837 us | 0.1587 us | 0.1407 us |
         CreateDelegateSubstitute | Core |    Core |  7.557 us | 0.0533 us | 0.0352 us |
		 
AFTER

                           Method |  Job | Runtime |      Mean |     Error |    StdDev |
--------------------------------- |----- |-------- |----------:|----------:|----------:|
         CreateDelegateSubstitute |  Clr |     Clr | 13.452 us | 0.1607 us | 0.1424 us |
         CreateDelegateSubstitute | Core |    Core |  9.630 us | 0.1893 us | 0.2714 us |
```

Obviously, I don't want you to profile our code 😅 But if you could benchmark Moq instead, it would be interesting to see whether you notice the same degradation. So it might be easier for you to check whether degradation resides somewhere in the code you are adding (as previously we also used `Castle.Proxy` to activate proxies and `Delegate.CreateDelegate()` to make delegates)...🤔

stakx:
@zvirja: I suspect that `ProxyUtil.TryCreateDelegateToMixin` might be more costly than it needs to be in some circumstances. If you do your own `proxy.GetMethod("Invoke").CreateDelegate(delegateType, proxy)` that might bring you down a few &mu;s.

(That method has to deal with the possibility of there being several `Invoke` overloads, and it has to pick the right one. However, if you *know* that you're mixing only one delegate type into a type that doesn't have other `Invoke` methods, then you can forego all of that and just pick the method called `Invoke`.)

Here are my quick benchmark results:

```
BenchmarkDotNet=v0.11.4, OS=Windows 10.0.17134.407 (1803/April2018Update/Redstone4)
Intel Core i5-8250U CPU 1.60GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3221.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3221.0
```

| Method                                          |     Mean |     Error |   StdDev  |   Median |
|------------------------------------------------ |---------:|----------:|----------:|---------:|
| 1. Moq 4.10.1 unchanged                           | 15.98 us | 0.1320 us | 0.1235 us |
| 2. Reference latest Castle.Core                   | 12.53 us | 0.3527 us | 1.023 us  | 11.89 us |
| 3. Use new delegate proxying approach             | 16.74 us | 0.1800 us | 0.1503 us |
| 4. Don't use `ProxyUtil.TryCreateDelegateToMixin` | 14.14 us | 0.0251 us | 0.0235 us |

(Not sure btw. why referencing Castle.Core as a project instead of its latest NuGet package makes such a big difference.)

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
stakx: Done. I've chosen to use `Delegate` in the public method signatures, but `MulticastDelegate` gets used for delegate type checks. See the last commit's message for details.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
jonorossi: Makes sense, good call.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -26,11 +26,84 @@ namespace Castle.DynamicProxy
...
+			{
+				throw new ArgumentException("Type is not a delegate type.", nameof(delegateType));
+			}
+
+			var invokeMethod = delegateType.GetMethod("Invoke");
+			var proxiedInvokeMethod =
+				proxy
+				.GetType()
+				.GetMember("Invoke", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
+				.Cast<MethodInfo>()
```
jonorossi: Does `GetMember` return null or throw, if returns null won't `Cast<>` throw?

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -26,11 +26,84 @@ namespace Castle.DynamicProxy
...
+		///   Attempts to create a delegate of the specified type to a suitable `Invoke` method
+		///   on the given <paramref name="proxy"/> instance.
+		/// </summary>
+		/// <param name="proxy">The proxy instance to which the delegate should be bound.</param>
+		/// <param name="delegateType">The type of delegate that should be created.</param>
+		/// <param name="delegate">The produced delegate if the method succeeds; otherwise <see langword="null"/>.</param>
+		/// <returns><see langword="true"/> if the method succeeds; otherwise <see langword="false"/>.</returns>
+		public static bool TryCreateDelegateToMixin(object proxy, Type delegateType, out Delegate @delegate)
+		{
+			if (proxy == null)
```
jonorossi: I thought our style guidelines allowed simple single line if-statements for control statements, throw, return, break, continue.

src/Castle.Core/DynamicProxy/MixinData.cs
```diff
@@ -38,28 +44,77 @@ public MixinData(IEnumerable<object> mixinInstances)
...
 				var sortedMixedInterfaceTypes = new List<Type>();
 				var interface2Mixin = new Dictionary<Type, object>();
+				delegateMixinCount = 0;
 
 				foreach (var mixin in mixinInstances)
 				{
-					var mixinInterfaces = mixin.GetType().GetInterfaces();
+					Type[] mixinInterfaces;
+					object target;
+					if (mixin is MulticastDelegate @delegate)
```
jonorossi: The `@delegate` variable isn't used.

Since all delegates are `MulticastDelegate`s and therefore also `Delegate`s either type could be used here, but it seems a tad strange to see `MulticastDelegate` used here when it isn't needed. I understand the use in `IsDelegateType`.

https://stackoverflow.com/questions/4833053/why-delegate-types-are-derived-from-multicastdelegate-class-why-not-it-directly

src/Castle.Core/DynamicProxy/ProxyUtil.cs
stakx: Indeed they do! I changed a few instances of this to single-line formatting in https://github.com/castleproject/Core/pull/436/commits/68e8fefa55886a39a625f935a9d30aa355a3e1d0.

src/Castle.Core/DynamicProxy/MixinData.cs
stakx: I remembered you mentioning your preference for `MulticastDelegate` at an earlier time and went with that... probably a little too far in this case.

I changed the check from `is MulticastDelegate @delegate` to just `is Delegate`.

jonorossi: Yer I recall that too but don't remember why, and I saw I had a link to an article on the MSDN blogs about why. That page has since been killed with the recent change to remove MSDN branding from domains and I couldn't find the article.

Today 😉 I think we should use `Delegate` wherever possible, since `MulticastDelegate` should have been merged into `Delegate` nearly 20 years ago.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
stakx: Neither: `GetMember` returns an empty array if no matching members are found.

From the [API reference documentation for `Type.GetMember(string, BindingFlags)`](https://docs.microsoft.com/en-us/dotnet/api/system.type.getmember?view=netframework-4.7.2#System_Type_GetMember_System_String_System_Reflection_BindingFlags_):

> #### Returns
> MemberInfo[]
>
> An array of MemberInfo objects representing the public members with the specified name, if found; otherwise, an empty array.

Note also that this documentation incorrectly claims that only public members are returned; if you specify `BindingFlags.NonPublic`, it will also return non-public members. But the documentation *is* correct with respect to returning an empty array in the failure case.

Therefore I'd say that LINQ query should work as intended.

src/Castle.Core/DynamicProxy/MixinData.cs
stakx: Agreed. So i suppose `MulticastDelegate` is only useful to *quickly* check whether a type is a delegate type, as `type.IsSubclassOf(typeof(MulticastDelegate))` does less work than `typeof(Delegate).IsAssignableFrom(type)`.

stakx:
Out of curiosity, @jonorossi, can you say whether this new way of creating delegate proxies will also work for Windsor? I haven't looked at its code closely, and am still not overly familiar with it.

IIRC, Windsor is currently using a lower-level component (`DelegateProxyGenerator`), which might no longer be feasible once we make more of DP's internals `internal`.

stakx:
> [...] whether this new way of creating delegate proxies will also work for Windsor?

If I'm not mistaken, the necessary changes for Windsor&mdash;excluding any possible optimizations&mdash;might look as follows (applied [here](https://github.com/castleproject/Windsor/blob/35ebd6e5a77de289536b3862f366d6cbfad19ecc/src/Castle.Windsor/Facilities/TypedFactory/Internal/DelegateProxyFactory.cs#L51-L57)):

```diff
 		private Type GetProxyType(IProxyBuilder builder, Type targetDelegateType)
 		{
-			var scope = builder.ModuleScope;
-			var logger = builder.Logger;
-			var generator = new DelegateProxyGenerator(scope, targetDelegateType)
-			{
-				Logger = logger
-			};
-			return generator.GetProxyType();
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(targetDelegateType);
+			return builder.CreateClassProxyType(typeof(object), Type.EmptyTypes, options);
 		}
```

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -26,11 +26,73 @@ namespace Castle.DynamicProxy
...
+
+		/// <summary>
+		///   Attempts to create a delegate of the specified type to a suitable `Invoke` method
+		///   on the given <paramref name="proxy"/> instance.
+		/// </summary>
+		/// <param name="proxy">The proxy instance to which the delegate should be bound.</param>
+		/// <param name="delegateType">The type of delegate that should be created.</param>
+		/// <param name="delegate">The produced delegate if the method succeeds; otherwise <see langword="null"/>.</param>
+		/// <returns><see langword="true"/> if the method succeeds; otherwise <see langword="false"/>.</returns>
+		public static bool TryCreateDelegateToMixin(object proxy, Type delegateType, out Delegate @delegate)
```
stakx: Somehow, using the `Try` pattern for this method (and its generic counterpart) feels wrong. In the typical use case of having previously called `options.AddDelegate[Type]Mixin`, there is no reason why users should be expected to deal with failure at this point. *If* this method fails, users will generally have no idea what might have gone wrong, and they likely won't be able to proceed, so it would be better for DynamicProxy to throw an exception directly.

src/Castle.Core.Tests/DynamicProxy.Tests/DelegateMixinTestCase.cs
```diff
@@ -0,0 +1,311 @@
...
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateTypeMixin_when_given_delegate_type_succeeds()
+		{
+			var options = new ProxyGenerationOptions();
+			options.AddDelegateTypeMixin(typeof(Action));
+		}
+
+		[Test]
+		public void ProxyGenerationOptions_AddDelegateMixin_when_given_delegate_type_throws_ArgumentNullException()
```
stakx: Typo, should say `_when_given_null_`.

jonorossi:
> Out of curiosity, @jonorossi, can you say whether this new way of creating delegate proxies will also work for Windsor? I haven't looked at its code closely, and am still not overly familiar with it.
> 
> IIRC, Windsor is currently using a lower-level component (`DelegateProxyGenerator`), which might no longer be feasible once we make more of DP's internals `internal`.
> [...]
> var options = new ProxyGenerationOptions();
> options.AddDelegateTypeMixin(targetDelegateType);
> return builder.CreateClassProxyType(typeof(object), Type.EmptyTypes, options);

Yep, looks about right. Maybe we should mark `DelegateProxyGenerator` obsolete now since we've got a better option.

> Somehow, using the `Try` pattern for this method (and its generic counterpart) feels wrong. In the typical use case of having previously called `options.AddDelegate[Type]Mixin`, there is no reason why users should be expected to deal with failure at this point. _If_ this method fails, users will generally have no idea what might have gone wrong, and they likely won't be able to proceed, so it would be better for DynamicProxy to throw an exception directly.

I agree, I'm glad you are changing it. I didn't comment on it but I did wonder what the use case of wanting to handle the failure case yourself was, just assumed you had a good reason 😉.

stakx:
> just assumed you had a good reason

Glad to hear you approve of that last-minute change, which didn't get reviewed. I was fairly confident that it was the right call, so I just went ahead.

I've been a fan of the `Try` pattern esp. since C# introduced the possibility to declare variables inline (`out var`, `is T x`), but there's such a thing as overuse. ;-)

That, and I recently came to suspect that the `Try` pattern will once again become somewhat less useful with C# 8's nullable reference types, where you can make it explicit in the return type that a method might *not* succeed.

Issue #238, opened 2017-03-19
Thread safe invocations


TKul6:
Hello,

I wanted to create an "async shell" for methods to run asynchronously.

I created an **AsyncInterceptor** which look like this:

```csharp
public void Intercept(IInvocation invocation)
        {


            //Methods starts with add_ or remove_ are registration and unregistration methods to the class's events.
            //Those methods should be invoked synchronously
            if (invocation.Method.Name.StartsWith("add_") || invocation.Method.Name.StartsWith("remove_"))
            {
                invocation.Proceed();
            }
            else
            {
                   Task.Run(() => invocation.Proceed());
            }


       }
```

Unfortunately, it didn't work. For some reason the interceptor was called much more than once.

By looking at [AbstractInvocation](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/AbstractInvocation.cs) I saw the Proceed method is not thread safe ( Mostly the usage of **currentInterceptorIndex** ).

I wonder if there is anyway to support this kind of behavior.

Thanks,

osoykan:
There is a workaround solution that already implemented. [It might gives an idea.](https://github.com/osoykan/Stove/blob/8de3d1a86c62e41396d5780aa725dacc916a83cd/src/Stove/Domain/Uow/UnitOfWorkInterceptor.cs#L68-L84)

TMaster:
Unfortunately it won't work for me.  
My intercepted methods return void, I just wan't to execute them on a different thread (it is a WCF callback contract and I want to free the WCF thread ASAP). My problem is the invocation.Proceed call itself being unsafe. 
I know I can  wrap it in a lock block and solve the issue, However I'll prefer if there is a built in solution for this type of behavior. like a built in thread safe invocation..

fir3pho3nixx:
Considered this: http://stackoverflow.com/questions/12144077/async-await-when-to-return-a-task-vs-void?

fir3pho3nixx:
This might also help: http://blog.stephencleary.com/2012/02/async-and-await.html

TMaster:
[...]

PR #439, opened 2019-03-24
Add `IInvocation.CaptureProceedInfo()` method

### LINKED ISSUE #145: Asynchronous Interceptors

stakx:
**TL;DR:** This resolves #145.

This is another attempt (after #438) at enabling asynchronous interceptors (#145) via an API addition that's as non-breaking as possible, in the sense of both "not breaking any abstractions" and "not introducing any breaking changes".

I abandoned #438 because the abstraction introduced there didn't seem quite right: We don't really need the ability to snapshot an `IInvocation` object's state (specifically, `currentInterceptorIndex`) and restore it later; what we *do* need is a way to make `invocation.Proceed()` calls work from an async continuation that executes when interception of `invocation` has already completed.

The new API introduced here is an `invocation.CaptureProceedInfo()` method that captures, in the form of an `IInvocationProceedInfo` descriptor object (think Reflection!), whatever `invocation.Proceed()` would do at that specific moment. That action can then be `Invoke`d at a later time via that returned descriptor object:

```diff
 void IInterceptor.Intercept(IInvocation invocation)
 {
     invocation.ReturnValue = InterceptAsync(invocation);
 }

 async Task InterceptAsync(IInvocation invocation)
 {
+    var proceed = invocation.CaptureProceedInfo();

     await ...              // early return, i.e. interception completes before the next line
-    invocation.Proceed();  // this would "proceed" back to the very first interceptor

+    proceed.Invoke();      // this proceeds to the interceptor following the current one
 }
```

This should be sufficient to let people implement some kind of `AsyncInterceptor` base class.

What this does *not* solve is #238, i.e. the fact that `AbstractInvocation.Proceed` is inherently thread-unsafe due to the non-atomic `currentInterceptorIndex` increment/decrement dance. #428 may possibly do a better job in that regard, FWIW.

/cc @JSkimming, @brunoblank (sorry for needlessly CCing you in #438; I hope this PR will fare better)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Enhancements:
 - Added trace logging level below Debug; maps to Trace in log4net/NLog, and Verbose in Serilog (@pi3k14, #404)
 - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
 - DictionaryAdapter: Exposed GetAdapter overloads with NameValueCollection parameter in .NET Standard (@rzontar, #423)
+- New `IInvocation.GetProceedInfo()` method to enable better implementations of asynchronous interceptors (@stakx, #439)
 
 Deprecations:
 - The API surrounding `Lock` has been deprecated. This consists of the members listed below. Consider using the Base Class Library's `System.Threading.ReaderWriterLockSlim` instead. (@stakx, #391)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Enhancements:
 - Added trace logging level below Debug; maps to Trace in log4net/NLog, and Verbose in Serilog (@pi3k14, #404)
 - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
 - DictionaryAdapter: Exposed GetAdapter overloads with NameValueCollection parameter in .NET Standard (@rzontar, #423)
-- New `IInvocation.GetProceedInfo()` method to enable better implementations of asynchronous interceptors (@stakx, #439)
+- New `IInvocation.CaptureProceedInfo()` method to enable better implementations of asynchronous interceptors (@stakx, #439)
 
 Deprecations:
 - The API surrounding `Lock` has been deprecated. This consists of the members listed below. Consider using the Base Class Library's `System.Threading.ReaderWriterLockSlim` instead. (@stakx, #391)
--- a/docs/dynamicproxy-async-interception.md
+++ b/docs/dynamicproxy-async-interception.md
@@ -1,0 +1,290 @@
+# Asynchronous interception
+
+This article discusses several interception scenarios related to asynchrony. We'll look at these scenarios in increasing order of implementation difficulty:
+
+ * Intercepting awaitable methods that don't produce a return value (e.g. `Task`), with and without using `async`/`await` in interceptor code
+
+ * Intercepting awaitable methods that do produce a return value (e.g. `Task<TResult>`), with using `async`/`await` in interceptor code
+
+ * Using `invocation.Proceed()` in combination with `async`/`await`
+
+This article contains C# code examples that make use of the `async`/`await` keywords. Before you proceed, please make sure that you have a good understanding of how these work. They are merely "syntactic sugar" that cause the .NET compilers to rewrite an async method to a state machine with continuations. The [Async/Await FAQ by Stephen Toub](https://devblogs.microsoft.com/pfxteam/asyncawait-faq/) explains these keywords in more detail.
+
+For brevity's sake, the examples shown in this article will focus on an interceptor's `Intercept` method. Assume that code examples are backed by the following code:
+
+
+```csharp
+var serviceProxy = proxyGenerator.CreateInterfaceProxyWithoutTarget<IService>(new AsyncInterceptor());
+
+// Examples will show how interception gets triggered:
+int result = await serviceProxy.GetAsync();
+
+public interface IService
+{
+	Task DoAsync();
+	Task<int> GetAsync(int n);
+}
+
+class AsyncInterceptor : IInterceptor
+{
+	// Examples will mostly focus on this method's implementation:
+	public void Intercept(IInvocation invocation)
+	{
+		...
+	}
+}
+```
+
+
+## Intercepting awaitable methods that don't produce a return value
+
+Intercepting awaitable methods (e.g. ones with a return type of `Task`) is fairly easy and not much different from intercepting non-awaitable methods. You'll simply have to set the intercepted invocation's return value to any valid `Task` object.
+
+
+```csharp
+// calling code:
+await serviceProxy.DoAsync();
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = httpClient.PostAsync(...);
+}
+```
+
+
+That wasn't very interesting, and in real-world scenarios you'll quickly reach a point where you'd like to use `async`/`await`. That, however, is also fairly trivial:
+
+
+```csharp
+// calling code:
+await serviceProxy.DoAsync();
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = InterceptAsync(invocation);
+}
+
+private async Task InterceptAsync(IInvocation invocation)
+{
+	// In this method, you have the comfort of using `await`:
+	var response = await httpClient.PostAsync(...);
+	...
+}
+```
+
+
+Things get more complicated once you want to return a value to the calling code. Let's look at that next!
+
+
+## Intercepting awaitable methods that produce a return value
+
+When intercepting an awaitable method that produces a return value (such as a method with a return type of `Task<TResult>`), it is important to remember that the very first `await` that gets hit by execution effectively returns to the caller right away. (The C# compiler will move the statements following the `await` into a continuation that will execute once the `await` "completes".)
+
+In other words, the very first `await` inside your interceptor completes the proxy method interception!
+
+DynamicProxy requires that interceptors (or the proxy target object) provide a return value for intercepted non-`void` methods. Naturally, the same requirement holds in an async scenario. Because the first `await` causes an early return to the caller, you must make sure to set the intercepted invocation's return value prior to any `await`.
+
+We already did that at the end of the last section; let's quickly go back and take a closer look:
+
+
+```csharp
+// calling code:
+var result = await serviceProxy.DoAsync();
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = InterceptAsync(invocation);
+	//         ^^^^^^^^^^^^^
+}
+
+private async Task InterceptAsync(IInvocation invocation)
+{
+	// At this point, interception is still going on
+	// as we haven't yet hit upon an `await`.
+
+	var response = await httpClient.PostAsync(...);
+
+	// At this point, interception has already completed!
+
+	invocation.ReturnValue = ...;
+	// ^ Any assignments to `invocation.ReturnValue` are no longer
+	//   observable by the calling code, which already received its
+	//   return value earlier (specifically, the `Task` produced by
+	//   the C# compiler representing this asynchronous method).
+	//   `invocation` is essentially a "cold", "detached", "stale",
+	//   or "dead" object (pick your favourite term).
+}
+```
+
+
+So, how would we communicate a return value back to the calling code when the intercepted method's return type is not just `Task`, but say, `Task<int>`? The following example shows how you can control the result of the task received by calling code:
+
+
+```csharp
+// calling code:
+var result = await serviceProxy.GetAsync(41);
+Assert.Equal(42, result);
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = InterceptAsync(invocation);
+}
+
+private async Task<int> InterceptAsync(IInvocation invocation)
+{
+	// We can still use `await`:
+	await ...;
+
+	// And we can simply `return` values, and the C# compiler
+	// will do the rest for us:
+	return (int)invocation.Arguments[0] + 1;
+}
+```
+
+
+Unfortunately, it usually won't be *that* easy in real-world scenarios, where you cannot assume that every method that your interceptor deals with will have the exact same return type of `Task<int>`. So, instead of being able to just comfortably `return someInt;` from a `async Task<int>` method, you'll have to resort to a non-generic `Task`, a `TaskCompletionSource`, and some reflection:
+
+
+```csharp
+// calling code--as before:
+var result = await serviceProxy.GetAsync(41);
+Assert.Equal(42, result);
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	var returnType = invocation.Method.ReturnType;
+
+	// For this example, we'll just assume that we're dealing with
+	// a `returnType` of `Task<TResult>`. In practice, you'd have
+	// to have logic for non-`Task`, `Task`, `Task<TResult>`, and
+	// any other awaitable types that you care about:
+	Debug.Assert(typeof(Task).IsAssignableFrom(returnType) && returnType.IsGenericType);
+
+	// Instantiate a `TaskCompletionSource<TResult>`, whose `Task`
+	// we will return to the calling code, so that we can control
+	// the result:
+	var tcsType = typeof(TaskCompletionSource<>)
+	              .MakeGenericType(returnType.GetGenericArguments()[0]);
+	var tcs = Activator.CreateInstance(tcsType);
+	invocation.ReturnValue = tcsType.GetProperty("Task").GetValue(tcs, null);
+
+	// Because we're not in an `async` method, we cannot use `await`
+	// and have the compiler generate a continuation for the code
+	// following it. Let's therefore set up the continuation manually:
+	InterceptAsync(invocation).ContinueWith(_ =>
+	{
+		// This sets the result of the task that we have previously
+		// returned to the calling code, based on `invocation.ReturnValue`
+		// which has been set in the (by now completed) `InterceptAsync`
+		// method (see below):
+		tcsType.GetMethod("SetResult").Invoke(tcs, new object[] { invocation.ReturnValue });
+	});
+}
+
+private async Task InterceptAsync(IInvocation invocation)
+{
+	// In this method, we now have the comfort of `await`:
+	var response = await httpClient.GetStringAsync(...);
+
+	// ... and we can still set the final return value! Note that
+	// the return type of this method is now `Task`, not `Task<TResult>`,
+	// so we can no longer `return` a value. Instead, we use the
+	// "stale" `invocation` to hold the real return value for us.
+	// It will get processed in the continuation (above) when this
+	// async method completes:
+	invocation.ReturnValue = (int)invocation.Arguments[0] + 1;
+}
+```
+
+
+Phew! And things get even more complex once we want to do an `invocation.Proceed()` to a succeeding interceptor or the proxy's target method. Let's look at that next!
+
+
+## Using `invocation.Proceed()` in combination with `async`/`await`
+
+Here's a quick recap about `invocation.Proceed()`: This method gets used to proceed to the next interceptor in line, or, if there is no other interceptor but a proxy target object, to that. Remember this image from the [introduction to DynamicProxy](dynamicproxy-introduction.md#interception-pipeline):
+
+![](images/proxy-pipeline.png)
+
+Let's get straight to the point: `Proceed()` will not do what you might expect in an async scenario! Remember that the very first `await` inside your interceptor completes interception, i.e. causes an early return to the calling code? This means that interception starts to "bubble back up" towards the calling code (i.e. along the green arrows in the above picture).
+
+Therefore, after having `await`-ed in your interceptor, interception has completed at the position of the last green arrow. Calling `invocation.Proceed` in the continuation (i.e. after the `await`) will then simply advance to the very first interceptor again... that's likely not what you want, and can cause infinite loops and other unexpected malfunctions.
+
+
+```csharp
+// calling code:
+await serviceProxy.DoAsync();
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = InterceptAsync(invocation);
+}
+
+private async Task InterceptAsync(IInvocation invocation)
+{
+	await ...;
+	invocation.Proceed(); // will not proceed, but reenter this interceptor
+	                      // or--if this isn't the first--an earlier one!
+}
+```
+
+
+In order for you to be able to work around this problem, DynamicProxy offers another method, `invocation.CaptureProceedInfo()`, which allows you to capture the invocation's current position in the interception pipeline (i.e. where along the yellow arrows it is currently located). This method returns an object to you which you can use to continue interception from this very same location onward.
+
+The solution to the problem demonstrated above now becomes very simple:
+
+
+```csharp
+// calling code--as before:
+await serviceProxy.DoAsync();
+
+// interception:
+public void Intercept(IInvocation invocation)
+{
+	invocation.ReturnValue = InterceptAsync(invocation);
+}
+
+private async Task InterceptAsync(IInvocation invocation)
+{
+	// If we want to `Proceed` at any point in this method,
+	// we need to capture how far in the invocation pipeline
+	// we're currently located *before* we `await`:
+	var proceed = invocation.CaptureProceedInfo();
+
+	await ...;
+
+	// At this point, interception is completed and we have
+	// a "stale" invocation that has been reset to the very
+	// beginning of the interception pipeline. However,
+	// that doesn't mean that we cannot send it onward to the
+	// remaining interceptors or to the proxy target object:
+	proceed.Invoke();
+
+	// At this point, a later interceptor might have over-
+	// written `invocation.ReturnValue`. As explained earlier,
+	// while the calling code will no longer observe this
+	// value, we could inspect it to set our own task's
+	// result (if we returned a `Task<TResult>`).
+}
+```
+
+
+## Closing remarks
+
+ * As you have seen, async interception is only trivial in very simple scenarios, but can get quite complex very quickly. If you don't feel comfortable writing such code yourself, look for third-party libraries that help you with async interception, for example:
+
+   * [Castle.Core.AsyncInterceptor](https://www.nuget.org/packages/Castle.Core.AsyncInterceptor) (third-party, despite the Castle Project's package namespace)
+
+   If you are the author of a generally useful async interception helper library, and would like to add your library to the above list, feel free to submit a PR.
+
+ * The above examples have shown cases where a "stale" invocation object has its `ReturnValue` repeatedly set, even though only the first value might be observed by calling code. It is however possible that invocation objects are recorded somewhere, and could be inspected later on. Other parties might then observe a `ReturnValue` that does not reflect what the original caller got.
+
+   Because of that, it might be good practice for your async interceptors to restore, at the end of interception, the `ReturnValue` to the value that actually made it back to the calling code.
+
+ * The same recommendation&mdash;restoration to the value(s) observed by the calling code&mdash;applies to by-ref arguments that have been repeatedly overwritten using `invocation.SetArgumentValue`.
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
@@ -35,7 +35,7 @@ namespace Castle.DynamicProxy.Tests
 				new SetReturnValueInterceptor(0),
 				new WithCallbackInterceptor(invocation =>
 				{
-					var proceed = invocation.GetProceedInfo();
+					var proceed = invocation.CaptureProceedInfo();
 				}),
 			};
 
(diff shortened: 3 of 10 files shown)


JSkimming:
@stakx I'm gonna clone/build your PR locally and look to get a PR raised with https://github.com/JSkimming/Castle.Core.AsyncInterceptor to see how things fare. I'll feed back ASAP.

JSkimming:
@stakx, just some early feedback (I'm still hacking)

__`IInvocationProceedInfo.Invocation` property__

A property on `IInvocationProceedInfo` to return the underlying `IInvocation` would be handy. It will make the implementation of [AsyncInterceptorBase](https://github.com/JSkimming/Castle.Core.AsyncInterceptor/blob/v1.7.0/src/Castle.Core.AsyncInterceptor/AsyncInterceptorBase.cs#L66-L83) simpler as I'm thinking of changing the two abstract methods to take just the `IInvocationProceedInfo`. This will allow implementations of `AsyncInterceptorBase` to still access the `IInvocation` if needed.

__Heap allocation due to calling `GetProceedInfo()`__

Your implementation instantiates the heap object for every interception that uses `GetProceedInfo()` which was something you wanted to avoid (IIFK from a comment on a different thread). I agree, and I'm wondering if a way to avoid this is possible.

Instead of GetProceedInfo() returning an interface, can it return a struct instead? The implementation is merely a reference and an `int` which isn't much more to copy than a reference.

stakx:
@JSkimming - thanks a lot for trying this out right away! Hopefully that way, we can get this async riddle solved soon!

> A property on IInvocationProceedInfo to return the underlying IInvocation would be handy.

Sounds reasonable, and while I'd like to keep that interface as tight as possible for now, if it's really helpful for users (i.e. currently just you :wink:), I'll add it.

But: Given that you retrieve a `IInvocationProceedInfo` from an `IInvocation`, you know the latter at some point in time. Couldn't you just create a `(IInvocationProceedInfo, IInvocation)` tuple and store that instead of just the former?

> Your implementation instantiates the heap object for every interception that uses GetProceedInfo() which was something you wanted to avoid [...].
>
> Instead of GetProceedInfo() returning an interface, can it return a struct instead? The implementation is merely a reference and an int which isn't much more to copy than a reference.

I've considered making it a `readonly struct`, but decided against it because it is problematic from an API design point of view: Any `IInvocation` implementation should be free to decide how to model `Proceed` state. By making `InvocationProceedInfo` a value type with an `int interceptorIndex`, we'd force any and all implementations to be able to map their own state to that way of modelling `Proceed` state. That would work for present-day `AbstractInvocation`, but not in the general case.

This could be mitigated by simply having `InvocationProceedInfo` have a `Action proceed` field instead of a combination of `IInvocation` + `int`. But then someone would have to allocate that delegate, which also creates an object on the heap. Nothing gained!

I think I can live with this very small performance dip, as long as these objects only get built on demand ("you only pay for what you use"), vs. building a whole batch of them in advance.

Finally, this facility will likely only be useful when dealing with async interception peculiarities, and there's a lot more overhead involved in juggling with async / `Task` objects than allocating one or two objects on the heap, so in the big scheme of things `IInvocationProceedInfo` being short lived heap objects shouldn't matter all *that* much.

brunoblank:
Good discussion and thoughts.

I like the fact that it makes a fairly straight forward workaround possible for the cases where await before proceed problem occur. This adds the extra complexity to _how to implement the Invoke method_.

This can be abstracted away in as @JSkimming is pointing out in `AsyncInterceptorBase`. 

I also agree with @stakx regarding the cost of allocating the extra object on the heap shouldn't matter that much.

I am happy an fix for the long lived _await_ problem is near :)

I do think we should take the opportunity to do the breaking change in [AsyncInterceptor](https://github.com/JSkimming/Castle.Core.AsyncInterceptor)
- removing the `Proceed` method

I also think after looking at the changes AsyncInterceptor - that the splitting up the "return" value to different methods should be looked at again. Just make the "Proceed" method return a `Task` and have the return-value fetched from IInvokation (same as for "synchronous" DP) 

_Side-note:_
If `InvokeMethodOnTarget` was added to `IInvocation` then AsyncInterceptor could solve this problem entirely internally.

JSkimming:
@stakx, thanks so much for these fixes. It's a top effort to navigate the challenges of maintaining backwards compatibility of a very popular library, but also to help drive it forward, so first and foremost, hats off to you my friend.

respect_hats_off

I've picked up an old branch I was hacking on in a failed attempt to fix this in [Castle.Core.AsyncInterceptor](https://github.com/JSkimming/Castle.Core.AsyncInterceptor "Castle.Core.AsyncInterceptor"), and applied the changes against a local copy of Castle.Core. The changes are in this commit JSkimming/Castle.Core.AsyncInterceptor@96d9a0f.

⭐️⭐️⭐️⭐️ I can confirm it has resolved the issues. ⭐️⭐️⭐️⭐️

This is great. There is a number of outstanding issues against the library which were dependent on this fix.

The fix I applied above was with the `IInvocation Invocation { get; }` proprty. I added a second commit with the changes without it JSkimming/Castle.Core.AsyncInterceptor@d7b90f8.

So concerning the `IInvocation` property, I don't have strong feelings, if you prefer the cleaner `IInvocationProceedInfo` I think that's perfectly reasonable.

I'm also considering a bit of a refactor of [AsyncInterceptorBase](https://github.com/JSkimming/Castle.Core.AsyncInterceptor/blob/v1.7.0/src/Castle.Core.AsyncInterceptor/AsyncInterceptorBase.cs) as I feel it's imposing too many assumptions. It's trying to make life simple for implementations by providing just two asyncronious methods to implement, but in doing so, it forces asynchrony on synchronous method invocations (with [some hackery](https://github.com/JSkimming/Castle.Core.AsyncInterceptor/blob/v1.7.0/src/Castle.Core.AsyncInterceptor/AsyncInterceptorBase.cs#L98) to get around it), and now it forces `IInvocationProceedInfo` on implementations even if they don't need it.

So given this will be a breaking change, I may choose a bit of a larger refactor at the same time.

Anyway, thanks again for this. It's very much appreciated.

brunoblank:
@JSkimming glad to hear the fix is resolving the issues!

I think it is perfectly fine to force asynchrony on AsyncInterceptions :) 

When thinking on the use case for AsyncInterceptor is when you in an "async call flow". So either the method should be async (and everything is fine) or sync then the method "should be cpu bound" (this is fine also) and not I/O (this would lead to the sync over async threadpool thread starvation). 

The point being - if that could was written without using `Core` it would face the same problem. 

Given this assumption, the Invoke method could run the non `Task` based methods synchronously on the same thread and return _Task.Completed_

stakx:
@JSkimming - That's some awesome news! Very happy to hear that you got things working! 🍾🎆

Thank you also for the kind words. :heart: I must confess that I've stayed away from those async issues longer than I should have, simply because they seemed so daunting and complex. Now I'm more than a little surprised at how quickly we've been able to make progress; I didn't expect it to go so well. Thanks to you for giving feedback so quickly.

I'm probably not going to merge this just yet. Ideally, we'd get @jonorossi to look at this new API and give it his blessings. Also, do you think we should wait until after you've refactored your AsyncInterceptor library, or do you think we'll be good to go either way?

brunoblank:
> Also, do you think we should wait until after you've refactored your AsyncInterceptor library, or do you think we'll be good to go either way?

I vote for: good to go directly

JSkimming:
:shipit:

JSkimming:
It could be worth releasing a beta version to nuget, we could then iterate on that and solicit feedback with a beta AsyncInterceptor.

stakx:
@jonorossi - Could we do a pre-release that includes this PR anytime soon to unblock the AsyncInterceptor project?

src/Castle.Core/DynamicProxy/AbstractInvocation.cs
```diff
@@ -188,5 +193,31 @@ private MethodInfo EnsureClosedMethod(MethodInfo method)
...
+				this.interceptorIndex = invocation.currentInterceptorIndex;
+			}
+
+			public void Invoke()
+			{
+				var previousInterceptorIndex = invocation.currentInterceptorIndex;
+				try
+				{
+					invocation.currentInterceptorIndex = interceptorIndex;
+					invocation.Proceed();
```
jonorossi: At this point the invocation really is complete because an interceptor chose to set a `ReturnValue` and decided not to proceed anymore returning control and completing the method invocation. Obviously in the async scenario we have to deal with the slightly leaky abstraction to handle the async machinery.

What do you think of adding a `completed` (or similar) boolean field to `AbstractInvocation` to enforce a readonly `ReturnValue` and prevent `SetArgumentValue` calls (unfortunately can't do much with the arguments array) after the invocation has completed. This should help catch broken interceptors because it would be a fairly easy mistake to make.

src/Castle.Core/DynamicProxy/IInvocationProceedInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+	///   Describes the <see cref="IInvocation.Proceed"/> operation for an <see cref="IInvocation"/>
+	///   at a specific point during interception.
+	/// </summary>
+	public interface IInvocationProceedInfo
+	{
+		/// <summary>
+		///   Executes the <see cref="IInvocation.Proceed"/> operation described by this instance.
+		/// </summary>
+		/// <exception cref="NotImplementedException">There is no interceptor, nor a proxy target object, to proceed to.</exception>
+		void Invoke();
```
jonorossi: Is there a reason you chose to call this `Invoke` rather than `Proceed`? Is it just because you had `GetProceedInfo` returning a delegate?

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
```diff
@@ -0,0 +1,97 @@
...
+	using System;
+
+	using Castle.DynamicProxy.Tests.Interceptors;
+	using Castle.DynamicProxy.Tests.InterClasses;
+	using Castle.DynamicProxy.Tests.Interfaces;
+
+	using NUnit.Framework;
+
+	[TestFixture]
+	public class InvocationProceedInfoTestCase
```
jonorossi: Could we get a unit test with 3 interceptors (i.e. 2 interceptors after the `WithCallbackInterceptor`), that's the one I'd be worried about, even thought the code appears to be fine for that since the index is restored.

src/Castle.Core/DynamicProxy/IInvocation.cs
```diff
@@ -114,6 +114,12 @@ public interface IInvocation
 		/// </remarks>
 		void Proceed();
 
+		/// <summary>
+		///   Returns an object describing the <see cref="Proceed"/> operation for this <see cref="IInvocation"/>
+		///   at this specific point during interception.
+		/// </summary>
+		IInvocationProceedInfo GetProceedInfo();
```
jonorossi: Something is bugging me that this looks too normal, like a part of the API people should actually use commonly. Like how no one thinks twice about using `Type.GetTypeInfo` even though many of the methods have been added back to `Type` in .NET Standard.

I wondered why we couldn't just have a `CreateProceedDelegate` (or `GetProceedDelegate`) and remembered you said this about allocations:

> This could be mitigated by simply having InvocationProceedInfo have a Action proceed field instead of a combination of IInvocation + int. But then someone would have to allocate that delegate, which also creates an object on the heap. Nothing gained!

This is currently allocating the `ProcessInfo` object, what if we just allocated a **delegate instead**? Could we have a private `Proceed(int index)` (or maybe `ProceedCompleted(int)` which throws if the invocation isn't flagged completed yet) that `CreateProceedDelegate` would call? The `Proceed` `MethodInfo` is already cached, we can easily do this one too.

jonorossi:
> Could we do a pre-release that includes this PR anytime soon to unblock the AsyncInterceptor project?

Right now you can get `0.0.516` from AppVeyor:
https://ci.appveyor.com/nuget/core-0mhe40ifodk8

We do have "Do not publish NuGet package artifacts to account/project feeds on Pull Requests" set but it is publishing them anyway. I can't find anyway to filter the feed so you'll just have to look at the last build and pick that exact version.

src/Castle.Core/DynamicProxy/AbstractInvocation.cs
stakx: > What do you think of adding a `completed` (or similar) boolean field to `AbstractInvocation` to enforce a readonly `ReturnValue` and prevent `SetArgumentValue` calls [...] after the invocation has completed.

For simplicity's sake, let's ignore setting by-ref arguments and focus on setting the return value, which is the more important and common scenario (by-ref parameters don't play well with async).

In an async interception scenario, it's true that setting `ReturnValue` only has an effect on the caller before the first `await`. But it's still useful that subsequent interceptors or the proxy target object can set `ReturnValue` again, because right now, they have no other way of communicating their return value to earlier interceptors.

Consider a case where we intercept a method `Task<int> ComputeAsync()`. We have one interceptor, which (for argument's sake) needs to `await` something ("X"), then wants to `Proceed` to the proxy target object, then use the return value produced by that to compute its own modified return value. This might be implemented as follows:

1. The interceptor sets `ReturnValue` to a non-completed `Task<int>` and `await`s X. This causes the task to be returned to the caller.

2. Now the interceptor `Proceed`s to the proxy target object.

3. The method on the proxy target object is just a regular method that *has to* `return` a `Task<int>` (except if it threw). The invocation object generated by DP will turn that return value into an assignment to `ReturnValue`. (This new return value will never be seen by the original caller.)

4. Now we're back in our interceptor, which can inspect this new `ReturnValue`, `await` it, extract its result, derive its own return value from that, and complete its own `Task<int>` to the final return value.

Now, if we forbade repeated setting of `ReturnValue` by means of a `completed` flag, how would the interceptor be able to receive the return value produced by the proxy target object?

I think we should allow setting `ReturnValue` after interception has completed. But we should perhaps provide documentation that explains how `IInvocation` objects can essentially go "stale" in async scenarios, and that it's an async interceptor's responsibility to restore an `IInvocation` to a clean state (i.e. it should finally restore `ReturnValue` to the value that the caller received).

Shall I try to cook up some documentation?

src/Castle.Core/DynamicProxy/IInvocationProceedInfo.cs
stakx: There's two aspects to this:

* Linguistics. Given that `IInvocationProceedInfo` represents a specific `Proceed` call, `Invoke` seemed like a more appropriate name: You can "invoke" the `Proceed` operation; you don't "proceed to" the `Proceed` operation.

* Extensibility. We could just as well have a method `Action IInvocation.GetProceeder()` that returns a simple delegate. I opted for something more extensible. Originally I had additional methods on `IInvocationProceedInfo` that allowed inspection of what exactly you'd be proceeding to: `bool IsInterceptor(out IInterceptor)` and `bool IsTarget(out object)`. It turned out that those aren't needed to implement an async interceptor so I removed them, but I liked the idea of being able to "reflect over" the interception pipeline, so I kept `IInvocationProceedInfo` instead of a simple delegate.

That being said, if you prefer the method to be named `Proceed`, I don't really mind renaming it.

src/Castle.Core/DynamicProxy/IInvocation.cs
stakx: There's no way to *just* allocate a delegate. You'll need at least one additional object that holds the copied interceptor index. Say if you returned a lambda that captures the `currentInterceptorIndex`, the C# compiler will transparently copy that over into a display class which will act as the delegate's `Target` object. So there'll be two objects, vs. just the one in the case of `IInvocationProceedInfo`.

Adding a `private void Proceed(int index)` won't help because `CreateProceedDelegate` should return a parameterless `Action` (lest we expose `AbstractInvocation`'s internals!), but that delegate will still have to know what argument to pass to `Proceed(int)`. So we're back to needing an additional state object.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
stakx: Sure. Can you say more what exactly worries you about a 3-interceptors scenario?

src/Castle.Core/DynamicProxy/IInvocation.cs
stakx: > Something is bugging me that this looks too normal, like a part of the API people should actually use commonly.

I think I know what you mean. I think there's a trade-off to be made here: #428 (i.e. giving `Intercept` an additional parameter `Action proceed`) is elegant but a big breaking change. This PR is less elegant but a much smaller breaking change, and more efficient. `Action CreateProceedDelegate` is somewhere in-between, also from an efficiency perspective.

ndrwrbgs:
Could you add a test that calls proceed multiple times? E.g. an async retry interceptor.

stakx:
@ndrwrbgs - Can do. Alternatively, feel free to suggest a concrete test case, I'd be happy to cherry-pick it from you. You might be more efficient at writing that particular test, or even already have one ready for adaptation somewhere, given your previous involvement with async interception. If not I'll have a go at it.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
```diff
@@ -0,0 +1,97 @@
...
+	{
+		private readonly ProxyGenerator generator = new ProxyGenerator();
+
+		[Test]
+		public void Proxy_without_target_and_last_interceptor_ProceedInfo_succeeds()
+		{
+			var interceptors = new IInterceptor[]
+			{
+				new SetReturnValueInterceptor(0),
```
jonorossi: Does this and the next unit test do what you expect, `SetReturnValueInterceptor` doesn't proceed so the `WithCallbackInterceptor` wouldn't get called. Maybe I'm just misreading it.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
jonorossi: I had to think about the logic to make sure that when inside the continuation the index is restored correctly so you could call `GetProceedInfo` or `Proceed` a second time to move into the 3rd interceptor. I'm realising now that you already test this since it is proceeding to the target so that handles the scenario, however having a test setting up another continuation inside the 2nd interceptor would probably be a good test to make sure GetProceedInfo doesn't get broken in the future if we do replace the index.

src/Castle.Core/DynamicProxy/AbstractInvocation.cs
jonorossi: > because right now, they have no other way of communicating their return value to earlier interceptors.

Ah yer, silly me. Yep, reusing `ReturnValue` (and arguments) for the real invocation and continuations is a good compromise, noting this in documentation maybe even with a rough diagram would greatly help users when they have to come here.

> I think we should allow setting `ReturnValue` after interception has completed. But we should perhaps provide documentation that explains how `IInvocation` objects can essentially go "stale" in async scenarios, and that it's an async interceptor's responsibility to restore an `IInvocation` to a clean state (i.e. it should finally restore `ReturnValue` to the value that the caller received).
> 
> Shall I try to cook up some documentation?

Cooking up some documentation definitely sounds like the next step. Talking about the lifetime of an `IInvocation` and how you can use the instance after the fact via continuations is the key.

src/Castle.Core/DynamicProxy/IInvocation.cs
jonorossi: > There's no way to _just_ allocate a delegate. You'll need at least one additional object that holds the copied interceptor index

Of course the compiler's capture class, I'm such an idiot. Sorry for my sloppy review, I shouldn't do code reviews at 1am 😞.

I'm now coming around to this idea, maybe just some small naming changes, you've definitely thought about it a lot more than me.

I guess my next question is do you see `IInvocationProceedInfo` a read-only info class (like `TypeInfo`) or something more stateful like `System.Threading.ExecutionContext`? My guess is the former since you've got `GetProceedInfo` rather than "capture" or another verb. I just wonder if the API reads clear enough that the proceed info you get is tied directly to the invocation it was created/captured from, not just any invocation of that method, and that "invoking" the proceed info actually modifies the invocation.

Would you ever consider enhancing `ProceedInfo.Invoke` to restore not just the interceptor index but also the return value as it exits in the finally block?

Maybe the naming is clear enough, the framework has `ExceptionDispatchInfo`, with static `Capture(ex)` and instance `Throw`. Had you considered hiding this functionality inside the info class without an interface, e.g. `InvocationProceedInfo.Capture(invocation)` and then `proceedInfo.Invoke`. Capture would just throw if it wasn't given an `AbstractInvocation`, not that DP works with other invocations, and users won't even know about this type unless they need it. It would move this more advanced functionality and a little bit of a wart to the side while being a supported API. Maybe I worrying way too much, anyone using DP has to have a decent idea of what they are doing.

I think at minimum we should make the XML documentation describe that this isn't a general purpose method, we don't want people doing `invocation.GetProceedInfo().Invoke()` not realising they should be using `invocation.Proceed()`.

My comments are a bit rambling, but I thought I'd keep them even though they don't provide actionable feedback. Happy to talk on anything that you think is useful, but there is no need to respond to each thing I've said in this comment.

src/Castle.Core/DynamicProxy/IInvocationProceedInfo.cs
jonorossi: Makes sense. This object obviously describes where you are proceeding to. I was thinking more like this object represented the current context so you'd "proceed" from the current position/index in the chain since that is what it stores. Semantics I guess, you've been consistent with naming so `Invoke` fits.

src/Castle.Core/DynamicProxy/IInvocation.cs
stakx: > Sorry for my sloppy review

Not at all, I'm glad you're approaching this from different angles than I did!

> do you see `IInvocationProceedInfo` a read-only info class

Not just that, but immutable. That's important to guarantee the fact that it captures the `Proceed` action *at a specific moment in time*.

Btw. I like the notion of "capturing" / getting a "snapshot". It would underline the immutable nature of `IInvocationProceedInfo` better! `CaptureProceedInfo`, perhaps?

I'm somewhat unsure about the best way to remove `GetProceedInfo` out of the direct sight of regular users (also not sure whether it's even necessary, users should read the docs instead of blindly using APIs IMO). If `EditorBrowsableState.Advanced` and XML docs (my preferred approach) are not sufficient, then `InvocationProceedInfo.Capture(...).Invoke()` would be a nice solution too, despite its limitations, i.e.: turning `IInvocation` into a leaky abstraction, and not being as extensible with reflection-like methods (since the info type would then be more of an opaque token like `ExceptionDispatchInfo`, rather than a reflection `...Info`).

> Would you ever consider enhancing `ProceedInfo.Invoke` to restore not just the interceptor index but also the return value

Probably not, because the return value isn't really part of the "proceed state"... and you still need the new `ReturnValue` after the `Invole` call, if you need it at all. I think we should leave cleaning up `ReturnValue` the responsibility of downstream users.

If the capture/restore became the focus then "invocation mementos" (like e.g. the one I tried out in #438) might be a more appropriate abstraction overall.

jonorossi: > Not at all, I'm glad you're approaching this from different angles than I did!

👍

> Btw. I like the notion of "capturing" / getting a "snapshot". It would underline the immutable nature of `IInvocationProceedInfo` better! `CaptureProceedInfo`, perhaps?

Great, I think renaming `GetProceedInfo` to `CaptureProceedInfo` makes it clear you are taking a snapshot at that point in time, if you call capture again on the same `IInvocation` especially in another interceptor you'll obviously get a new `IInvocationProceedInfo` instance.

> I'm somewhat unsure about the best way to remove `GetProceedInfo` out of the direct sight of regular users (also not sure whether it's even necessary, users should read the docs instead of blindly using APIs IMO).

Yer, my idea of a static method is overkill, I think naming it `CaptureProceedInfo` would be enough to push someone to look at the XML docs or full docs to know what this method is for, its name doesn't sound like a normal getter that always returns the same object.

stakx: Done.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
stakx: I added a few additional multi-interceptor tests in https://github.com/castleproject/Core/pull/439/commits/62b3698c44766a6e8eccbd50f96bcd8e04146490. Is this the kinds of tests you were thinking of?

src/Castle.Core/DynamicProxy/AbstractInvocation.cs
stakx: I've made an attempt to documentation in https://github.com/castleproject/Core/pull/439/commits/b378594c0c94cd2befb90eaa329eb5941b04e973. See it rendered at https://github.com/stakx/Castle.Core/blob/invocationproceedinfo/docs/dynamicproxy-async-interception.md.

No nice illustrations for now, as it's getting a little late; but perhaps it's "good enough" for a start? Let me know if you have any requests or suggestions to improve it.

src/Castle.Core.Tests/DynamicProxy.Tests/InvocationProceedInfoTestCase.cs
jonorossi: Yep, looks good.

docs/dynamicproxy-async-interception.md
```diff
@@ -0,0 +1,290 @@
...
+
+This article discusses several interception scenarios related to asynchrony. We'll look at these scenarios in increasing order of implementation difficulty:
+
+ * Intercepting awaitable methods that don't produce a return value (e.g. `Task`), with and without using `async`/`await` in interceptor code
+
+ * Intercepting awaitable methods that do produce a return value (e.g. `Task<TResult>`), with using `async`/`await` in interceptor code
+
+ * Using `invocation.Proceed()` in combination with `async`/`await`
+
+This article contins C# code examples that make use of the `async`/`await` keywords. Before you proceed, please make sure that you have a good understanding of how `async`/`await` works. See e.g. the article ["Asynchronous Programming with `async` and `await`" on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/).
```
jonorossi: Spelling: contains

stakx: Well spotted, thanks. Fixed.

docs/dynamicproxy-async-interception.md
```diff
@@ -0,0 +1,290 @@
...
+
+This article discusses several interception scenarios related to asynchrony. We'll look at these scenarios in increasing order of implementation difficulty:
+
+ * Intercepting awaitable methods that don't produce a return value (e.g. `Task`), with and without using `async`/`await` in interceptor code
+
+ * Intercepting awaitable methods that do produce a return value (e.g. `Task<TResult>`), with using `async`/`await` in interceptor code
+
+ * Using `invocation.Proceed()` in combination with `async`/`await`
+
+This article contains C# code examples that make use of the `async`/`await` keywords. Before you proceed, please make sure that you have a good understanding of how `async`/`await` works. See e.g. the article ["Asynchronous Programming with `async` and `await`" on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/).
```
jonorossi: That MSDN page is pretty basic, there is a big gap between it and this article. I think adding a link to something that discusses how the compiler actually does things would be a good idea.

Maybe link to something like https://devblogs.microsoft.com/pfxteam/asyncawait-faq/, and make some brief comment that this DP article assumes that the developer is aware of the async/await machinery and that async/await is syntactic sugar over normal CLR methods.

stakx: Oh, you're right, Stephen Toub's FAQ is a much better resource. I've updated this paragraph.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Enhancements:
 - Added trace logging level below Debug; maps to Trace in log4net/NLog, and Verbose in Serilog (@pi3k14, #404)
 - Recognize read-only parameters by the `In` modreq (@zvirja, #406)
 - DictionaryAdapter: Exposed GetAdapter overloads with NameValueCollection parameter in .NET Standard (@rzontar, #423)
+- New `IInvocation.GetProceedInfo()` method to enable better implementations of asynchronous interceptors (@stakx, #439)
```
jonorossi: @stakx the method name in the changelog is out of date

stakx: Another good catch... fixed. Say thanks to your hawk's eyes!

jonorossi:
Seems like a good time to merge to master. Go ahead if you are ready.

stakx:
@jonorossi: Thanks for reviewing this so carefully! :+1:

stakx:
@ndrwrbgs: While I did add two tests that proceed multiple times, I opted not to add too many test about async scenarios in order to keep this PR focused.

Feel free to submit a PR that e.g. adds more tests to `AsyncInterceptorTestCase.cs`. Like I said, you're perhaps quicker than me at writing the exact test you had in mind, anyway.

ndrwrbgs:
@stakx  I don't have the attention span right now to fully review what changes were involved, but I blindly took the change and the feature branch in https://github.com/JSkimming/Castle.Core.AsyncInterceptor and it seems that whatever you did + @JSkimming's changes in his project at https://github.com/JSkimming/Castle.Core.AsyncInterceptor/tree/hacking-for-async-fix have made it so an asynchronous retry interceptor is now possible, cool!

(not to mention other solutions, like asynchronous logging interceptor)

Issue #389, opened 2018-06-20
Reduce public API surface


stakx:
As noted in https://github.com/castleproject/Core/issues/193#issuecomment-398749977:

> Castle library in general expose way too much publicly. Back in the old days of Castle especially with MonoRail everything was public by default and you'd be able to poke or extend things wherever you chose, pros and cons obviously.
>
> I do however think we should start closing these things that are public [...].

The goal of this issue is to let us keep track of which types and type members should be removed from Core's publicly visible API in the next major version (and be marked as `[Obsolete]` even before that).

I suggest that we use posts for discussion and keep an up-to-date list of all decisions here at the top. If you have a better idea or format for this issue, please share!

### Should be `[Obsolete]` in vNextMinor / removed from vNextMajor's public API:

(A check mark means that a member is already marked `[Obsolete]` in the `master` branch.)

* [X] `Castle.Core.Internal.Lock` (class)
* [X] `Castle.Core.Internal.ILockHolder` (interface)
* [X] `Castle.Core.Internal.IUpgradeableLockHolder` (interface)
* [X] `Castle.DynamicProxy.Generators.BaseProxyGenerator.AddToCache` (method)
* [X] `Castle.DynamicProxy.Generators.BaseProxyGenerator.GetFromCache` (method)
* [X] `Castle.DynamicProxy.Generators.CacheKey` (class)
* [x] ~~`Castle.DynamicProxy.Generators.DelegateProxyGenerationHook` (type)~~
* [x] ~~`Castle.DynamicProxy.Generators.DelegateProxyGenerator` (type)~~
* [X] `Castle.DynamicProxy.Generators.Emitters.ArgumentsUtil.IsAnyByRef(ParameterInfo[])`
* [X] `Castle.DynamicProxy.IChangeProxyTarget.ChangeProxyTarget(object)`
* [x] ~~`Castle.DynamicProxy.Internal.InternalsUtil`~~
* [X] `Castle.DynamicProxy.Internal.InternalsUtil.IsAccessible(this MethodBase)`
* [X] `Castle.DynamicProxy.Internal.InternalsUtil.IsInternal(this MethodBase)`
* [X] `Castle.DynamicProxy.Internal.InternalsUtil.IsInternalToDynamicProxy(this Assembly)`
* [x] `Castle.DynamicProxy.ModuleScope.DefineType` (method)
* [X] `Castle.DynamicProxy.ModuleScope.GetFromCache` (method)
[...]

PR #440, opened 2019-03-28
Export public API to C# text files and fail unit test if not matching

jonorossi:
For quite some time I've wanted to be able to see (and verify) public API changes as we make them, and easy see what APIs we need to mark obsolete and then remove.

This is related to #389.

I wanted something like the `ref` directory in .NET Standard and .NET Core that hold their reference assembly's source code (their built tools aren't (EDIT) meant to be used outside their build process), but I found [the FakeItEasy guys have been doing this for quite some time](https://github.com/FakeItEasy/FakeItEasy/tree/master/tests/FakeItEasy.Tests.Approval), and they are using [PublicApiGenerator](https://github.com/JakeGinnivan/ApiApprover) to do it (the ApiApprover part is marked obsolete).

I wanted these files to be more obvious as something we care about (a bit like a public contract) and followed the `ref` directory convention, however realised they aren't really reference assemblies since we never build them, however not sure that matters much, but I'd prefer them out in front rather than hidden in a unit test project directory, so happy to hear suggestions on a different naming convention.

What are people's thoughts on this whole thing in general?

### DIFF
--- a/ref/Castle.Core-net35.cs
+++ b/ref/Castle.Core-net35.cs
@@ -2451,6 +2451,7 @@ namespace Castle.DynamicProxy
         public object Proxy { get; }
         public object ReturnValue { get; set; }
         public abstract System.Type TargetType { get; }
+        public Castle.DynamicProxy.IInvocationProceedInfo CaptureProceedInfo() { }
         public object GetArgumentValue(int index) { }
         public System.Reflection.MethodInfo GetConcreteMethod() { }
         public System.Reflection.MethodInfo GetConcreteMethodInvocationTarget() { }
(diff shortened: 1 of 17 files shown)


jonorossi:
I should have mentioned the workflow:
- As a developer you run the unit tests and it'll overwrite all the .cs files in the ref directory for you
- The build server (`CI=true` env var) will compare the built assemblies to the files on disk, the assertion will obviously fail if someone changes the public API without also including the change to the ref/*.cs files

stakx:
> and followed the `ref` directory convention, however realised they aren't really reference assemblies since we never build them [...].

That probably doesn't matter much. The difference to, say, dotnet/corefx, doesn't seem all that great: As far as I know, their reference assembly sources are generated from the reference assemblies, not vice versa. They just use a different tool ([GenAPI](https://github.com/dotnet/buildtools/tree/master/src/GenAPI)) and a different source assembly from which the public contract is derived.

If it's important to you that the `ref/` directory actually reflects a real reference assembly, we could easily build such a one using an up-to-date Roslyn compiler. I believe they added a compiler switch for that sometime earlier during the C# 7.x timeline.

Or, we could also put the public contracts in an `api/` directory.

I think we should simply go with `ref/` as it's an established and well-understood convention. And because we don't actually distribute reference assemblies, I don't see how any harm can come of it.

jonorossi:
I've got no desire to actually build a reference assembly, so happy to continue with what we've got. The approval process (mentally reminding you of public API changes) is what I'm mostly after, which then also makes it easy for us to trim down the public API which contains so much cruft that is holding us back.

There is a .NET Core Global Tool beta just released, not sure it fits exactly what we are doing here, but we may be able to move to that later: https://github.com/JakeGinnivan/ApiApprover/issues/55

I've rebased on master.

PR #444, opened 2019-03-29
Accuracy of docs & exception regarding `Proceed`

### LINKED ISSUE #442: Documentation falsely claims that each interceptor can only `Proceed` once
The [introductory article for DynamicProxy](https://github.com/castleproject/Core/blob/c6354b954fc5f19c345ef19e253eac5f29611772/docs/dynamicproxy-introduction.md#interception-pipeline) currently claims that:

> An interceptor can call `Proceed` at most once, otherwise an exception is thrown.

However, that no longer appears to be true today. [The current implementation of `Proceed`](https://github.com/castleproject/Core/blob/c6354b954fc5f19c345ef19e253eac5f29611772/src/Castle.Core/DynamicProxy/AbstractInvocation.cs#L100-L143) which goes back to https://github.com/castleproject/Core/commit/ea8fa099325d85fb03396be21430cce366aefdf5 allows an interceptor to `Proceed` more than once (in order to be able to implement retry logic).

We should update that part of the documentation.

(I am also uncertain under what circumstances the _"`invocation.Proceed()` has been called more times than expected"_ exception still gets thrown. This could possibly be unreachable code today, but I have yet to check with the unit test suite.)

stakx:
This closes #442.

ea8fa099 made it permissible to call `invocation.Proceed()` more than once from the same interceptor. The introductory article in the docs does not yet reflect that.

That previous commit also removed an error case for which `AbstractInvocation` goes to great lengths to build an error message. Today it is no longer needed as the error will only get triggered by seriously faulty client code. Replace it with a simpler constant message.

(I opted to keep the new message fairly unspecific, as there are a variety of possible usage scenarios today where a more specific error message could get readily misinterpreted.)

Finally, this slightly increases the accuracy of the description (in the introductory article about DP) of how the interception pipeline and `Proceed` work.


docs/dynamicproxy-introduction.md
```diff
@@ -47,8 +47,9 @@ Schematic view of how DynamicProxy works.
...
+* Each interceptor gets an `IInvocation` object (which is another important interface from DynamicProxy) that holds all the information about the current request, such as the `MethodInfo` of the intercepted method, along with its parameters and preliminary return value; references to the proxy and the proxied object; and a few other bits. Each invoked interceptor gets a chance to inspect and change those values before the actual method on the target object is called. For example, an interceptor may log debug information about the arguments passed to the method, or validate them.
+* Each interceptor can call `invocation.Proceed()` to pass control further down the pipeline. Interceptors usually call this method just once, but multiple calls are allowed, e.g. when implementing a "retry". (It is also permissible to cut short the interception pipeline and omit the call to `Proceed` altogether.)
+* When the last interceptor calls `Proceed`, the actual method on the proxied object is invoked, and then the call travels back, up the pipeline (green arrow) giving each interceptor another chance to inspect and act on the returned value or thrown exceptions.
```
stakx: Here's a small side-by-side summary of the relevant changes (typo, grammar, and punctuation fixes are not marked):

image

PR #450, opened 2019-07-13
WIP: Forward non-intercepted methods on class proxies to target

### LINKED ISSUE #67: Not proxied methods on class proxy with target should forward to the target [DYNPROXY-183]
Relogging YouTrack issue logged by Krzysztof 2 years ago.
http://issues.castleproject.org/youtrack/issue/DYNPROXY-183

---

Currently when we have a proxyable method on a class proxy with target, and chose not to proxy it, the default body of the method on the proxy object will be invoked (same behavior as inheritance-base class proxy).

The correct behavior should be to forward the call to the target object instead, like all the other "with target" proxy types are doing.

stakx:
This should resolve #67.

(At least as far as is possible: non-interceptable methods such as non-virtual, sealed, or private methods of course still cannot be forwarded to the proxy.)

### DIFF
--- a/src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
@@ -37,17 +37,12 @@ namespace Castle.DynamicProxy.Contributors
 
 		protected override MetaMethod GetMethodToGenerate(MethodInfo method, IProxyGenerationHook hook, bool isStandalone)
 		{
-			if (ProxyUtil.IsAccessibleMethod(method) == false)
+			if (method.IsPublic == false || ProxyUtil.IsAccessibleMethod(method) == false)
 			{
 				return null;
 			}
 
 			var accepted = AcceptMethod(method, true, hook);
-			if (!accepted && !method.IsAbstract)
-			{
-				//we don't need to do anything...
-				return null;
-			}
 
 			return new MetaMethod(method, method, isStandalone, accepted, hasTarget: true);
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Forward non-intercepted methods on class proxies to target (@stakx, #450)
+
 ## 4.4.0 (2019-04-05)
 
 Enhancements:
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/Classes/HasVirtualStringAutoProperty.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/Classes/HasVirtualStringAutoProperty.cs
@@ -1,0 +1,21 @@
+// Copyright 2004-2019 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Tests.Classes
+{
+	public class HasVirtualStringAutoProperty
+	{
+		public virtual string Property { get; set; }
+	}
+}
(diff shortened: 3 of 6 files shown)


src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
```diff
@@ -43,7 +43,7 @@ protected override MetaMethod GetMethodToGenerate(MethodInfo method, IProxyGener
 			}
 
 			var accepted = AcceptMethod(method, true, hook);
-			if (!accepted && !method.IsAbstract)
+			if (!accepted && Array.IndexOf(AllMethodsHook.skippedTypes, method.DeclaringType) >= 0)
```
stakx: I suspect that the `!method.IsAbstract` check is no longer necessary, since we're going to forward to a target object (which wouldn't be instantiable if it had abstract methods).

src/Castle.Core/DynamicProxy/AllMethodsHook.cs
```diff
@@ -32,6 +32,8 @@ public class AllMethodsHook : IProxyGenerationHook
 #endif
 		};
 
+		protected static readonly ICollection<Type> SkippedTypes = skippedTypes;
```
stakx: It would have been simpler to just change the access modifier of `SkippedTypes` from `protected` to `protected internal`, however that would (somewhat surprisingly) have led to changes in the `ref/` contract files, which I'd like to avoid whenever possible.

src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
stakx: I added that additional filter because without it, DP will create forwarding implementations for each and every member available, including all `object` members. Letting `MinimalisticMethodGenerator` create implementations for two of those, `object.Finalize` and `object.MemberwiseCopy`, will cause PE verification to fail.

To be honest, I didn't analyse the cause for this in detail, instead I saw that `AllMethodsHook` (contrary to its name) has `SkippedTypes`, an exclusion list for members of certain types. I figured the PE verification might be why `SkippedTypes` exists in the first place.

If we remove that `Array.IndexOf` check...

* I'd need to figure out why PE verification fails, so it can be avoided.
* I supect we'd still need some exclusion list / criteria regarding certain members. In particular, I'm not sure how we'd deal with `object.Finalize`. If the target doesn't override that method, then ideally we would not generate a forwarding implementation, either, since this will negatively affect GC.
* There might be other special cases we'd need to account for: Do we need special logic for `object.Clone`, `object.Equals`, or is it always safe to forward these to the target?

stakx:
Hmm, I'm seeing your all of your review comments twice. I put my replies with the first batch.

src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
stakx: > `object.Finalize` and `object.MemberwiseCopy` will cause PE verification to fail

Ah, it's actually rather simple: those are `protected` and therefore not callable from the proxy. `ForwardingMethodGenerator` might have been written with interfaces in mind, it doesn't perform any accessibility checks.

So I think what I'll do is to remove that weird `SkippedTypes` check, and do an accessibility check instead. (`!method.IsPublic` should cover most cases, plus we might need some logic to recognize explicitly implemented interface methods and call them through the interface.)

src/Castle.Core/DynamicProxy/AllMethodsHook.cs
stakx: This is no longer necessary as we really don't need to access `SkippedTypes` at all (see your other review comment), so we don't need to change this code at all.

stakx:
Once again, this is going to take more work than anticipated. What is needed here is a utility method that checks accessibility of a member `M` (the method on the target type) to some referent type `T` (the proxy type). Once properly implemented, the current `method.IsPublic` check can be replaced with something like `method.IsAccessibleTo(proxyType)`. There are some subtleties I'm possibly not thinking of right now, I'm pretty sure I'll discover them while writing tests.

(Also, we might be able to reimplement `ProxyUtil.IsAccessible*` in terms of the new accessibility check method, as the latter covers a superset of the former's functionality.)

I'll ping you once I'm done.

src/Castle.Core/DynamicProxy/Contributors/WrappedClassMembersCollector.cs
jonorossi: > `AllMethodsHook` (contrary to its name) has `SkippedTypes`

I think it was named that because it was the previous behaviour for all methods, most people don't need to intercept object methods.

> > `object.Finalize` and `object.MemberwiseCopy` will cause PE verification to fail
> 
> Ah, it's actually rather simple: those are `protected` and therefore not callable from the proxy

Yep, I don't recall ever having problems with those 2.

jonorossi:
> Once again, this is going to take more work than anticipated.

Haha, as always with DP.

> I'll ping you once I'm done.

👍

jonorossi:
@stakx how about this one?

stakx:
Same here. I'll eventually get back to it.

PR #454, opened 2019-07-24
Fix Matrix quote

PeterRockstars:
Updated the text to contain the actual quote from The Matrix :)


jonorossi:
Thanks, merged.

stakx:
Working on DynamicProxy can easily feel as easy as bending spoons with ones mind (that is, not so easy after all...), so it's great that we're now at least finally getting our quotes right. ;-)

What a refreshing PR! Thanks for making me smile, @PeterRockstars! :+1:

PeterRockstars:
> Working on DynamicProxy can easily feel as easy as bending spoons with ones mind (that is, not so easy after all...), so it's great that we're now at least finally getting our quotes right. ;-)
> 
> What a refreshing PR! Thanks for making me smile, @PeterRockstars! 👍

You're more than welcome! And thank you for making _me_ smile by using analogies from one of my favourite movies to write light-hearted documentation about such complex topics, making them much easier to undersand.

PR #455, opened 2019-07-26
Make NUnit tests runnable in VS Test Explorer

stakx:
While the test projects can simply be started to run the tests (thanks to NUnitLite), tests cannot be run in VS Test Explorer, which could be handy especially when one wants to run only a subset of all tests.
Currently, doing that requires one to set a `--test=...` command-line argument in the test project settings.)

This commit adds the packages required to run our tests in VS Test Explorer.

### DIFF
--- a/src/Castle.Core.Tests/Castle.Core.Tests.csproj
+++ b/src/Castle.Core.Tests/Castle.Core.Tests.csproj
@@ -17,6 +17,7 @@
 		<AssemblyOriginatorKeyFile>..\..\buildscripts\CastleKey.snk</AssemblyOriginatorKeyFile>
 		<PublicSign Condition="'$(OS)'=='Unix'">true</PublicSign>
 		<LangVersion>7.2</LangVersion>
+		<StartupObject>Program</StartupObject>
 	</PropertyGroup>
 
 	<ItemGroup Condition="'$(TargetFramework)'=='netcoreapp1.1'">
@@ -38,8 +39,10 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
 		<PackageReference Include="NUnit" Version="3.6.1" />
 		<PackageReference Include="NUnit.Console" Version="3.6.1" />
+		<PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
 		<PackageReference Include="NUnitLite" Version="3.6.1" />
 		<PackageReference Include="log4net" Version="2.0.8" />
 		<PackageReference Include="NLog" Version="4.5.0" />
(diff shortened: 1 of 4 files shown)


stakx:
Agh! I had everything working fine, then updated Visual Studio to 16.2.0, and now tests once again no longer run. I'll investigate.

stakx:
Found a workaround and [reported the problem in Microsoft Developer Community](https://developercommunity.visualstudio.com/content/problem/663145/test-explorer-in-vs-1620-can-no-longer-run-certain.html).

I know we're uglifying a unit test here (hopefully only for the time being), however I think it's worth the price if it allows us to pick and run tests straight from Visual Studio.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -38,8 +39,10 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
```
jonorossi: Interesting to see `Microsoft.NET.Test.Sdk` is versioned 16.2.0. Is this a package just for Visual Studio with a name making it sound like it is for all .NET CLI unit testing.

jonorossi:
> I know we're uglifying a unit test here (hopefully only for the time being), however I think it's worth the price if it allows us to pick and run tests straight from Visual Studio.

Agreed. I was expecting something much uglier 😉.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
stakx: AFAIK, the package version isn't directly tied to a particular version of VS. But I have never been too clear what exactly requires this package... maybe VS, maybe VSTest. I might run some experiments to find out. Wil report back my findings if & when I do so.

stakx: @jonorossi, while I don't understand the whole theory behind that package, it indeed enables `dotnet test`! In fact, we can use all of `dotnet restore|build|test` now, if we make one further addition in `common.props` that looks roughly like the following (as per https://github.com/Microsoft/msbuild/issues/1333#issuecomment-296346352):

```xml
<PropertyGroup>
  <FrameworkPathOverride Condition="'$(TargetFramework)|$(OS)' == 'net35|Windows'">$(MSBuildProgramFiles32)\Reference Assemblies\Microsoft\Framework\.NETFramework\v3.5\Profile\Client</FrameworkPathOverride>
  <FrameworkPathOverride Condition="'$(TargetFramework)|$(OS)' == 'net35|Unix'">$(MONO_DIR)/../lib/mono/2.0-api</FrameworkPathOverride>
</PropertyGroup>
```

(We already do something similar for .NET 3.5 compilation using Mono. I've only verified `dotnet` tooling for Windows, the Mono/Linux override will need some more work. But essentially, the `MONO_DIR` variable can be set on the shell using `export MONO_DIR=$(dirname $(which mono))`, then passed to the build.)

PR #456, opened 2019-07-26
Fix culture-dependent test failures

stakx:
While playing around with migrating the Unix/Mono build over to AppVeyor, I stumbled on some failing tests: We have two DictionaryAdapter tests that only work when `CultureInfo.CurrentCulture` happens to be set to anything other than the invariant culture.

This PR fixes these two tests by actively setting the culture instead of relying on the environment's choice.

(Funnily enough, fixing the tests is easiest when deciding on the invariant culture, since that one is guaranteed to be available, and its precise behavior as far as the tests are concerned is documented.)


stakx:
I'm going to merge this straight away without waiting for a review, as it's a relatively minor defect in the way the tests were written... and, given that they concern DictionaryAdapter, we probably don't want to spend too much time on it anyway. :)

jonorossi:
Looks good 👍

PR #458, opened 2019-07-27
WIP: Fix tests that do (or might) fail on Mono

stakx:
While playing around with migrating the Unix/Mono build over to AppVeyor, I encountered several tests that failed when running the test suite on Mono 6 (when run inside a `mono:6` Docker container).

This PR should take care of all those (potentially) failing tests.

Note: It's something of a riddle to me why some of these COM tests don't already fail today on Travis CI.

PR #459, opened 2019-07-27
Migrate away from Travis CI by running the Linux build on AppVeyor

### LINKED ISSUE #441: CI build server platform consolidation
We are using a mix of AppVeyor (for Windows) and Travis CI (for Linux), and recently both platforms have added the other operating system they didn't have previously. I think it is about time we look at just moving things to one service.

With [what is happening with Travis CI last month after being acquired](https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/) and the fact we lost all our history with the move to .com, I'd be a little wary of moving to Travis.

I think since we run all .NET Framework and .NET Core builds on AppVeyor, do releases and use the NuGet feeds; and only the Mono and .NET Core ones on Travis, we should just bring the Linux ones to AppVeyor to simplify things.

Note that we do releases from AppVeyor, so we need to make sure the Linux build doesn't also try doing releases when we tag a build. Not sure if we should add another job for making the release or we do something else, just something to be aware of.

Thoughts? Anyone interested in tackling this?

stakx:
This closes #441 and supersedes my previous PR #458 (which I'm merging into the present PR).

We start by defining an AppVeyor build matrix for a Windows and a Ubuntu Linux build. At this stage, some tests fail on the Linux build, so a few commits follow that fix and/or document these failures.

* [X] ~~At least one thing remains to be done here: It appears that tests guarded by `FEATURE_TEST_COM` do not run on the Linux build, even though they should:~~

(See update below.)

<sub>~~In theory, COM interop ought to work on Mono, as long as the COM subsystem (`ole32.dll`) isn't actually invoked. Whether or not COM works is more a question of Windows vs. other OSes than about CLR vs. Mono. That's why I'm introducing `[ExcludeOnOSPlatformsOtherThanWindows]` for one specific test. (That attribute could be more generic, like `[ExcludeOnFramework]`, but it's probably not worth the additional work at this point.)~~</sub>

<sub>~~`FEATURE_TEST_COM` on the other hand is a build-time guard mostly to exclude tests for .NET Core builds (TODO: need to fact-check this), but why does it (I guess) lead to tests being excluded on the Linux build & Mono test run?~~</sub>


stakx:
I am having some trouble figuring out why some tests are not discovered in the Ubuntu / Mono (`net461`) test run. I'm looking specifically at the tests in the DynamicProxy `ComInterfacesTests` fixture. I've established that while this fixture and its tests is present in the test assembly, NUnit doesn't discover it when running on AppVeyor.

stakx:
> It appears that tests guarded by `FEATURE_TEST_COM` do not run on the Linux build, even though they should:

The culprit is the .NET 3.5-era `ADODB.dll` which we use for some COM-based unit tests. It apparently fails to load on AppVeyor for Linux during test discovery, causing the latter to fail for referencing tests or test fixtures at a very early stage... so early, in fact, that NUnit simply doesn't look at these tests again; any attributes such as `[Ignore]`, `[Category("WindowsOnly")], `[ExcludeOnOSPlatformsOtherThanWindows]` etc. won't get processed. 

That's why I've had to come up with a hack to prevent assembly load failure during test discovery (see the last commit). Not a particularly nice solution, but it works.

The alternative would be to look more deeply at *why exactly* `ADODB.dll` fails to load in the first place. My best guess is that Mono's assembly binding policy will fail to match the referenced `mscorlib` version 1.0.3300.0 (.NET 3.5!) to the running framework's `mscorlib`, which likely has a version of `4.0.0.0` (.NET 4+). (I cannot say why it would've worked on Travis, though.)

If so, we could generate a primary interop assembly of our own from IDL definitions using .NET 4+ tooling to get more-up-to-date assembly references. (See [this example](https://github.com/moq/moq4/tree/master/tests/Moq.Tests.ComTypes); we'd have to do IDL definitions for some co-class that actually exists on a typical Windows installation.)

The current hack is a lot less work, of course.

stakx:
@jonorossi, I think this is ready for a first review. I tried not to stray too far from the main mission here (which is parting from Travis), but would like to follow up this PR with some others to modernize/simplify the build process as a whole.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
```diff
@@ -35,26 +35,47 @@ public void Marshal_Release_throws_when_called_with_IntPtr_Zero()
...
 		}
 
+		// NOTE: You may wonder why the following tests delegate their implementation to static methods
+		// in a private static class. This is because the method bodies contain references to ADODB.dll,
+		// which fails to load under Mono/Linux, causing test discovery to abort early. By moving these
+		// method bodies somewhere else, test discovery will succeed and we get a chance at specifying
+		// how the tests should run, e.g. using the `[Platform]` attributes.
+
 		[Test]
+		[Platform(Include = "Win", Reason = "Depends on ADODB.dll, which can fail to load under Mono/Linux.")]
```
stakx: To be honest, I am being lazy here: NUnit's `[Platform]` is not available on .NET Standard / Core, so this compiles only because these targets are currently excluded by `FEATURE_TEST_COM`. It would be more correct to e.g. declare a test `[Category("Windows")]`, and then exclude that category from the test run in `build.sh`. Or build a `[IncludeOnOS]` attribute similar to our `[ExcludeOnFramework]`.

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
```diff
@@ -36,6 +36,7 @@
 		<PackageReference Include="System.Collections.Specialized" Version="4.3.0" />
 		<PackageReference Include="System.ComponentModel" Version="4.3.0" />
 		<PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
+		<PackageReference Include="System.Runtime.CompilerServices.VisualC" Version="4.3.0" />
```
stakx: Needed to bring in the `IsLong` type (used by some modreq/modopt-related tests) in the `netstandard1.x` targets.

README.md
```diff
@@ -83,6 +83,7 @@ Symbol                              | NET35              | NET40              |
 `FEATURE_SYSTEM_CONFIGURATION`      | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
 `FEATURE_TARGETEXCEPTION`           | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
 `FEATURE_TEST_COM`                  | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
+`FEATURE_TEST_DATASET`              | :white_check_mark: | :white_check_mark: | :white_check_mark: | :no_entry_sign:
```
stakx: I called this `FEATURE_TEST_DATASET` (instead of just `FEATURE_DATASET`) to give a hint that `DataSet` is only used by tests, and not by the main library.

appveyor.yml
```diff
@@ -1,41 +1,57 @@
...
-  name: core
+for:
+-
+  matrix:
+    only:
+    - image: Ubuntu
+
+  build_script:
+  - uname -a
+  - mono --version
```
jonorossi: In our Travis script we were installing Mono 5.16.0, is there a way we can pin the Mono version? Otherwise the chance that a new version will break tests is increased if Mono just gets upgraded on every build.

appveyor.yml
```diff
@@ -1,41 +1,57 @@
...
+  - ./build.sh
+
+  deploy: off
+
+-
+  matrix:
+    only:
+    - image: Visual Studio 2017
+
+  init:
```
jonorossi: I'm having a hard time reading the YAML here to work out which sections belong to Windows and which apply to both. Can you add some block comments or something to make this clearer, I wanted to make sure only one job publishes to nuget.org, but both should upload test results and artifacts.

Can we do anything to make sure that we only publish to nuget.org if both builds succeed? Is that one of the future things you want to do?

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
```diff
@@ -29,29 +29,53 @@ public class ComInterfacesTests:BasePEVerifyTestCase
...
 		public void Can_proxy_existing_com_object()
 		{
-			var command = new Command();
-			var parameter = ExtractActualComParameterObject(command.CreateParameter("foo"));
-			generator.CreateInterfaceProxyWithTargetInterface(parameter);
+			Implementation.Can_proxy_existing_com_object(generator);
 		}
 
-		private Parameter ExtractActualComParameterObject(Parameter parameter)
+		private static class Implementation
```
jonorossi: Just so I've got this right, this is relying on the JIT not loading the private class during test discovery?

Another option would be to lazily reference/load adodb with reflection, and obviously call `CreateInterfaceProxyWithoutTarget` with a type variable and use reflection to new the `Command` class.

I'd prefer it to be correct code rather than relying on a JIT behaviour.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
jonorossi: Happy for it to be this way for now. We've actually got a bunch of this sort of thing where it only works because of the conditional compilation of something else, we'll consolidate when we remove .NET 3.5 and .NET Standard 1.x support in the future.

appveyor.yml
stakx: 1. I suppose AppVeyor don't update their VM images all that often, and they usually provide the previous image for some time in case breakages happen. Breaking changes would be a hassle, but it might be good for the Mono ecosystem as a whole if people noticed & reported those. I probably wouldn't mind dealing with that every now and then.

If we do want to avoid such breaks, I can see two fairly easy options:

2. We probably have admin rights on the Linux VM that AppVeyor spins up, so we could just uninstall Mono and reinstall a different version. That might be a fairly slow process, though.

3. We could prepare a Docker image with specific versions of Mono and the .NET Core SDK installed, then run our build inside a Docker container using that image. This would have to be updated by us every now and then.

I guess we could go with (1) and see how often Mono version updates actually break our build, and if that actually happens too often, still decide on (2) or (3) later on. What is your preference?

jonorossi: Number 1 sounds good, you've convinced me that getting new Mono versions is probably good for us, we get new .NET Framework versions without any control anyway.

Mono 6 was out in June so it'll probably be pretty soon on AppVeyor.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
stakx: > I'd prefer it to be correct code rather than relying on a JIT behaviour.

I was just thinking that we could completely eliminate JIT / assembly loader / version binding policy failure as an error source if we simply manually embedded the needed COM interop type definitions in our test project. We can open up `ADODB.dll` in ILDASM or ILSpy, then bring the required type definitions over into our own project. Then we could remove `ADODB.dll` from the solution.

jonorossi: > manually embedded the needed COM interop type definitions in our test project

👍. That one isn't a PIA so there shouldn't be any reliance on being signed a certain way.

appveyor.yml
stakx: > Can you add some block comments or something to make this clearer

Will do.

> Can we do anything to make sure that we only publish to nuget.org if both builds succeed? 

I *think* that's what will already happen, but it's a little difficult to verify.

Right now, because we're on the Free plan, we only get 1 concurrent job, so one build's jobs run sequentially. If we put the Linux job first and additionally enable fail-fast (`matrix.fast_finish: true`) then the Windows build will never run and deployment cannot happen.

The above strikes me as inelegant, we still want results from the Windows build even if the Linux build job failed. So I'd rather have a guarantee from AppVeyor that any `on_success` script or `deploy` section only runs if all build jobs have been successful, even when the `on_success` script is specific to one job & another job is still running in parallel.

I'm still checking with the AppVeyor docs, I'll report back my findings.

> both should upload test results and artifacts

Yes, that would be a good addition. I'll add the upload for the Linux build.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
stakx: OK, will do this then. It'll be nice to say bye to that DLL. :smirk:

appveyor.yml
stakx: > Can we do anything to make sure that we only publish to nuget.org if both builds succeed?

Not easily. From the [AppVeyor Discussions forum](https://help.appveyor.com/discussions/questions/1164-how-do-i-run-a-script-after-all-matrix-possibilities-succeed):

> All jobs run independently and `on_success` is part of the job, not a build

What we have now&mdash;two jobs that execute sequentially, letting the last one handle deployment&mdash;combined with the fail-fast strategy mentioned above would be the easiest option.

We could, of course, look at other CI offers such as Microsoft's Azure Pipelines to see if they offer the flexibility we're looking for...

jonorossi: Let's just let them continue to run in whatever order and only publish in the Windows build, it will be no different to what we have now with AppVeyor and Travis being completely separate.

It shouldn't be a problem anyway because pull requests build, then master builds, then finally you tag and get another build so should have picked up by then if the Linux build will fail.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
stakx: Done. I've had to make a somewhat arbitrary choice about which ADO types to bring in fully, and which to reproduce only partially (and where to make the cut in the latter cases). I've brought in all required enums, but truncated co-class types as well as the interface hierarchies. (This shouldn't matter since the vtables of derived COM interfaces are a superset of the base interfaces' vtables, and because we're only calling methods on interfaces, not on any co-classes directly.) I've also opted to omit custom marshalers, and COM event interfaces. I've however chosen to reproduce method signatures as accurately as possible.

appveyor.yml
stakx: Done. I added some comments to `appveyor.yml` to make it easier to read, and I restored the Windows build to run first. Regarding the test counts stated in AppVeyor's UI, these don't add up from the test count reported in the log, I dare say that's more of a problem with the AppVeyor UI than with the way how we upload test results from `appveyor.yml`.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
```diff
@@ -42,40 +42,18 @@ public void Marshal_Release_throws_when_called_with_IntPtr_Zero()
...
-			}
-
-			public static void Can_proxy_existing_com_object(ProxyGenerator generator)
-			{
-				var command = new Command();
-				var parameter = ExtractActualComParameterObject(command.CreateParameter("foo"));
-				generator.CreateInterfaceProxyWithTargetInterface(parameter);
-			}
-
-			private static Parameter ExtractActualComParameterObject(Parameter parameter)
```
stakx: This is no longer needed, because there will no longer be any custom marshaler converting the actual `Parameter` COM object to an `InternalParameter` wrapper.

src/Castle.Core.Tests/DynamicProxy.Tests/ComInterfacesTests.cs
```diff
@@ -29,30 +29,32 @@ public class ComInterfacesTests:BasePEVerifyTestCase
...
+			//           ^^^^^^^^^^^^^^^^^
+			// We're not going to be more precise because Mono and the CLR don't throw the same exception type.
+			// See https://github.com/mono/mono/issues/15853.
 		}
 
+		// NOTE: You may wonder why the following tests delegate their implementation to static methods
+		// in a private static class. This is because the method bodies contain references to ADODB.dll,
+		// which fails to load under Mono/Linux, causing test discovery to abort early. By moving these
+		// method bodies somewhere else, test discovery will succeed and we get a chance at specifying
+		// how the tests should run, e.g. using the `[Platform]` attributes.
```
jonorossi: ADODB.dll is now gone, is this still accurate?

stakx: Oh, you're absolutely right. I'll remove this before merging.

PR #464, opened 2019-08-04
XML docs for `ProxyGenerationOptions` & its mutating members

### LINKED ISSUE #387: `CacheKey` is mutable which can lead to collisions in type cache
The following was mentioned by @BrunoJuchli in https://github.com/castleproject/Core/issues/193#issuecomment-398640084 &mdash; it might be good to look into this, possibly in conjunction with #386:

> `CacheKey` is not immutable. Specifically `ProxyGenerationOptions` is mutable. Probably this should not be a problem in this test scenario (since we're using `ProxyGenerationOptions.Default` all the way, aren't we?), but this will, at least, cause other problems:
>
> It's possible to have two `CacheKey`s which are different but later, when they're added to a dictionary, already, make them the same. Imagine:
>
> ```csharp
> var hook1 = new AllMethodsHook();
> var hook2 = new DelegateProxyGenerationHook();
> var options1 = new ProxyGenerationOptions(hook1);
> var options2 = new ProxyGenerationOptions(hook2);
>
> // ---- create a proxy of <IFoo> with options1
> // ---- create a proxy of <IFoo> with options2
>
> options2.Hook = hook1;
> // ---> et voilà, we've got a key collision in the dictionary.
> ```
>
> Now I don't know whether this is just a performance problem (same proxy cached twice) or if it's a more severe problem. Ideally, after building the `ProxyGenerationOptions`, they should be immutable. Also, `CacheKey.GetHashCode` could be computed @ctor (performance optimization, but would also make it clear that it's immutable).

(Minor reformattings by me.)

stakx:
This adds XML documentation comments to `ProxyGenerationOptions` and those of its members that can lead to mutations. The documentation warns against mutating `ProxyGenerationOptions` once it's been used for proxy creation.<sub>~~, which may be sound advice regardless of what I said in https://github.com/castleproject/Core/issues/387#issuecomment-517937112. (That is, despite the fact that modifying  `ProxyGenerationOptions` should still work in practice even after using it for proxy creation. This might change in future DP versions, so let's give advice towards the more prudent path.)~~</sub>

Closes #387.


src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -103,12 +111,45 @@ public void GetObjectData(SerializationInfo info, StreamingContext context)
...
+		///   <para>
+		///     You should not modify this property once this <see cref="ProxyGenerationOptions"/> instance
+		///     has been used to create a proxy.
+		///   </para>
+		/// </summary>
 		public IProxyGenerationHook Hook { get; set; }
 
+		/// <summary>
+		///   Gets or sets the <see cref="IInterceptorSelector"/> that should be used by created proxies
+		///   to determine which interceptors to use for an interception. If set no <see langword="null"/>
```
stakx: Typo: `no` &rarr; `to`

PR #465, opened 2019-08-04
Match generic type parameters by position, not by name

### LINKED ISSUE #106: Generic method with differently named generic arguments to parent throws KeyNotFoundException
Issue: Throws KeyNotFoundException in https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs at method public Type[] GetGenericArgumentsFor(MethodInfo genericMethod)

Conditions: When Type name of generic type doesn't match the name of the Type in an interface.

Workaround: Change any generic Type names to match the name on the source interface.

This happens in Mono and .Net. If a generic method's generic Type name mismatches with its implemented name then this throws in Moq.

Original issues incorrectly posted on Moq's GitHub: [here](https://github.com/Moq/moq4/issues/193)

stakx:
Fixes #106.

I didn't (and still don't quite) trust this bug fix, mainly because it seems far too easy for DP standards (+-1 LoC!? 😝). However, it makes good sense to match generic type parameters by position instead of by name; in IL, for instance, generic type parameters are referred to by position (e.g. ` ``1 `). The execution environment doesn't really care about the names.

To give us some added safety, I ran a few additional checks:

* [X] I ran the Moq unit test suite (approx. 1,500 additional tests) against the updated version of DP to see whether this might cause regressions further downstream. (Doesn't look like it.)
* [X] I verified that the [original Moq issue which led to #106](https://github.com/Moq/moq4/issues/193) is fixed by this. (It is.)

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
@@ -279,6 +279,7 @@ namespace Castle.DynamicProxy.Generators.Emitters
 			return null;
 		}
 
+		[Obsolete("Intended for internal use only.")] // TODO: Remove this method.
 		public Type[] GetGenericArgumentsFor(Type genericType)
 		{
 			var types = new List<Type>();
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
@@ -17,6 +17,7 @@ namespace Castle.DynamicProxy.Generators.Emitters
 	using System;
 	using System.Collections.Generic;
 	using System.Diagnostics;
+	using System.Linq;
 	using System.Reflection;
 	using System.Reflection.Emit;
 
@@ -303,7 +304,8 @@ namespace Castle.DynamicProxy.Generators.Emitters
 			var types = new List<Type>();
 			foreach (var genType in genericMethod.GetGenericArguments())
 			{
-				types.Add(name2GenericType[genType.Name].AsType());
+				var matchingGenType = name2GenericType.Values.Single(t => t.GenericParameterPosition == genType.GenericParameterPosition);
+				types.Add(matchingGenType.AsType());
 			}
 
 			return types.ToArray();
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
@@ -17,7 +17,6 @@ namespace Castle.DynamicProxy.Generators.Emitters
 	using System;
 	using System.Collections.Generic;
 	using System.Diagnostics;
-	using System.Linq;
 	using System.Reflection;
 	using System.Reflection.Emit;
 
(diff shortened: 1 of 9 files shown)


src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs
```diff
@@ -303,7 +303,7 @@ public Type[] GetGenericArgumentsFor(MethodInfo genericMethod)
 			var types = new List<Type>();
 			foreach (var genType in genericMethod.GetGenericArguments())
 			{
-				types.Add(name2GenericType[genType.Name].AsType());
+				types.Add(genericTypeParams[genType.GenericParameterPosition].AsType());
```
jonorossi: This makes sense. I wonder if it would be a good idea to check if `genericTypeParams` isn't null and the array contains the element before accessing it since it is provided externally by `SetGenericTypeParameters`, then provide a useful exception, so we don't get something useless like the previous `KeyNotFoundException` one.

Have you looked at the previous `GetGenericArgumentsFor` overload to see if that should also be changed?

stakx: > I wonder if it would be a good idea to check if `genericTypeParams` isn't null and the array contains the element before accessing it since it is provided externally by `SetGenericTypeParameters`

Having an assertion to that effect might be good, but it shouldn't be any more necessary than checking whether `name2GenericType` contains the key `genType.Name`. Note how the only place where `SetGenericTypeParameters` is called from initializes both `genericTypeParams` and populates `name2GenericType`:

https://github.com/castleproject/Core/blob/808ef994eca34399bf46f934fd4b675ae8e4caf1/src/Castle.Core/DynamicProxy/Generators/Emitters/AbstractTypeEmitter.cs#L121

I'll give it some more thought.

stakx: > Have you looked at the previous `GetGenericArgumentsFor` overload to see if that should also be changed?

Good point. It indeed looks like this might also profit from the same change. I'll look into it.

stakx: > Have you looked at the previous `GetGenericArgumentsFor` overload to see if that should also be changed?

@jonorossi, it turns out that other overload is most probably unreachable code. See the new commit for details. Without any test cases, it's hard to whether or not it is faulty. But given that it's public and there may in theory be external consumers, I think the safest option is to leave it as is, but mark it `[Obsolete]`.

We have yet another method, `Type GetGenericArgument(string genericArgumentName)`, which precedes the two that we've discussed so far. The code paths surrounding generic type arguments are rather convoluted, and I don't understand them well enough (yet) to know whether or not there needs to be a change here, too.

stakx:
I think I'll need to take some time to study the (rather convoluted) code paths that are involved here in much more depth. I'd like to feel more confident in the change I'm proposing here... if possible.

jonorossi:
> I think I'll need to take some time to study the (rather convoluted) code paths that are involved here in much more depth. I'd like to feel more confident in the change I'm proposing here... if possible.

No problem.

jonorossi:
> I think I'll need to take some time to study the (rather convoluted) code paths that are involved here in much more depth. I'd like to feel more confident in the change I'm proposing here... if possible.

@stakx any change in your position? Should we merge or close this pull request?

stakx:
I haven't got around to really study this through as the code surrounding this is rather convoluted.

Let's close this for now. When I eventually get this fully done, I'll reopen.

PR #484, opened 2020-02-24
Shorten lock time

tangdf:


tangdf:
Review the  code  please.

tangdf:
```C#
    public class UserAppService
    {
        public virtual int Add(UserDto userDto)
        {
            //add to database;
            return 0;
        }
    }

    public class RoleAppService
    {
        public virtual int Add(RoleDto roleDto)
        {
            //add to database;
            return 0;
        }
    }

    public static class ServiceFactory
    {
        private static readonly IProxyBuilder s_proxyBuilder = new DefaultProxyBuilder(new ModuleScope(false, true));

        private static readonly ProxyGenerationOptions s_Options = new ProxyGenerationOptions(new PipelineServiceProxyHook());

        public static TService CreateService<TService>()
        {
            IInterceptor interceptor = new PipelineInterceptor();

            ProxyGenerator generator = new ProxyGenerator(s_proxyBuilder);

            return (TService) generator.CreateClassProxy(typeof(TService), s_Options, interceptor);
        }
    }
```

```C#
            //demo code

            //thread 1
            UserAppService userAppService = ServiceFactory.CreateService<UserAppService>();

            userAppService.Add(new UserDto());

            //thread 2
            RoleAppService roleAppService = ServiceFactory.CreateService<RoleAppService>();

            roleAppService.Add(new RoleDto());
````

When the proxy class of type UserAppService is generated, but when generatint the  proxy class of type RoleAppService , it prevents the creation of an instance of UserAppService.

jonorossi:
> When the proxy class of type UserAppService is generated, but when generating the proxy class of type RoleAppService , it prevents the creation of an instance of UserAppService.

Ah yes, thanks. I'll update the changelog.

stakx:
First off, apologies for having kept silent about this PR. The reason being that when I first saw this proposed change, it didn't feel like one that was entirely safe to make. It might be a good idea to look beyond `SynchronizedDictionary` and verify that locking in DynamicProxy will still work as expected. At surface level however, this LGTM. :+1:

jonorossi:
@stakx no problem, I've been MIA lately anyway.

I wondered that too when I first looked at it. Since this doesn't change the external semantics of `SynchronizedDictionary`, I then looked at `BaseProxyGenerator.ObtainProxyType` and it accesses the naming scope and calls `GenerateType` delegates all inside the same lock, now an upgradeable read lock rather than the write lock but still synchronised, I then confirmed that only generators use naming scopes since they aren't thread safe (i.e. can't read while writing), yes they are the only thing using it. I don't foresee any problems.

PR #491, opened 2020-04-29
Fix proxy generation for private protected constructors

mriehm:
Previously, it would not be possible to generate a proxy of a class with just a private protected constructor. This is because BaseProxyGenerator.IsConstructorVisible() did not check for the case of MethodBase.IsFamilyAndAssembly.

Now, we remove BaseProxyGenerator.IsConstructorVisible() and replace the call to it with a call to ProxyUtil.IsAccessibleMethod(), which has identical behavior as BaseProxyGenerator.IsConstructorVisible() except that it does already correctly handle MethodBase.IsFamilyAndAssembly.

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
@@ -1,7 +1,3 @@
-//----------------------------------------------------
-// Copyright 2020 Epic Systems Corporation
-//----------------------------------------------------
-
 // Copyright 2004-2016 Castle Project - http://www.castleproject.org/
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
@@ -1,3 +1,7 @@
+//----------------------------------------------------
+// Copyright 2020 Epic Systems Corporation
+//----------------------------------------------------
+
 // Copyright 2004-2016 Castle Project - http://www.castleproject.org/
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
```diff
@@ -380,6 +380,21 @@ public void ClassProxyShouldCallProtectedDefaultConstructor()
 			Assert.AreEqual("Something", ((ClassWithProtectedDefaultConstructor)proxy2).SomeString);
 		}
 
+		[Test]
+		public void ClassProxyShouldHaveDefaultConstructorWhenBaseClassHasPrivateProtected()
+		{
+			object proxy = generator.CreateClassProxy<ClassWithPrivateProtectedConstructor>();
+			Assert.IsNotNull(Activator.CreateInstance(proxy.GetType()));
```
stakx, line 387: I initially had a bit of trouble understanding this test. All I initially  saw was that you're creating two instances of the proxy type, which didn't make much sense until I realized the second instantiation is only a means to an end. I think this test would be easier to understand if it explicitly looked for the expected constructor using reflection, instead of assuming that `Activator` will do it for us.

src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
```diff
@@ -380,6 +380,21 @@ public void ClassProxyShouldCallProtectedDefaultConstructor()
...
+		}
+
+		[Test]
+		public void ClassProxyShouldCallPrivateProtectedDefaultConstructor()
+		{
+			object proxy = generator.CreateClassProxy<ClassWithPrivateProtectedConstructor>();
+			object proxy2 = Activator.CreateInstance(proxy.GetType());
+			Assert.AreEqual("Something", ((ClassWithPrivateProtectedConstructor)proxy2).SomeString);
```
stakx, line 395: Here I'm even less sure I understand why the assertion has to happen on an object created by `Activator`. Why create the first one at all? If you just need the proxy type, you could create that via `generator.ProxyBuilder`... but why not assert on `proxy` directly?

src/Castle.Core.Tests/DynamicProxy.Tests/BasicClassProxyTestCase.cs
mriehm: Honestly, I just copied both of these tests from the three other sets of nearly identical existing pairs of tests immediately above for different constructor accessibilities.

From the perspective of this specific fix, all that is strictly needed is one test which has one line of code which calls generator.CreateClassProxy<ClassWithPrivateProtectedConstructor>(), to verify that it doesn't throw. However, I thought that I should go with consistency with the existing tests, which is why you see these extra lines of code in the tests.
Considering that, would you still like me to make the change that you suggested to this test? If so, would you like me to update the similar tests immediately above in this file as well?

stakx: Consistency with the existing tests sounds reasonable (I didn't notice that's how you did it, sorry), and good enough for me.

stakx:
You appear to have made commits with different Git user names. Do you want to do something about this before we merge the PR?

mriehm:
Sorry, I'm still pretty unfamiliar with Github and Git in general (use different source control at work), so I'm not exactly sure how I would go about fixing the inconsistency and it doesn't seem like something worth spending too much time investigating unless you disagree.

So, you can just go ahead and commit this as it is now. Thanks.

jonorossi:
@mriehm FYI the commits are mixed between these 2 email addresses:
- mriehm@epic.com (no link to your account because this email address is not associated with your GitHub account)
- 3916550+mriehm@users.noreply.github.com (this one got used because you edited files on the GitHub web site and you've got "Keep my email addresses private" set in GitHub's email settings)

@stakx we should just squash and merge the 6 commits anyway to keep the history clean in this case. I'll leave it to you to merge and update the changelog.

stakx:
> to merge and update the changelog

Both done.

Thank you for contributing, @mriehm!

PR #494, opened 2020-05-01
Add hyperlink to my own `AsyncInterceptor` package to documentation page

stakx:
> If you are the author of a generally useful async interception helper library, and would like to add your library to the above list, feel free to submit a PR.

And voilà: such a PR has just arrived.

The proverbial paint on this library hasn't dried yet so obviously it cannot possibly be battle-tested at this stage. However, I believe the basic implementation idea behind to be sound, and I'd love to see people pick it up so that it has a chance to become reliable & hardened through actual use. Therefore my attempt here to advertise it. :smirk:

@castleproject/committers, I'm obviously biased here, so for fairness' sake, I perhaps shouldn't handle this PR myself. 😄 No problem at all if you think it's unwise / too early to add such a new library to the list.


jonorossi:
Your library looks "generally useful" 😉, so it qualifies 🏅

PR #495, opened 2020-05-01
Remove framework targets `net40`, `netstandard1.3`, `netstandard1.5`

stakx:
This is a follow-up to https://github.com/castleproject/Core/issues/407#issuecomment-620601558.

* [X] ~~`net35`~~ &mdash; **NOTE:** already removed in #496
* [X] `net40`
* [X] `netstandard1.3`
* [x] `netstandard1.5`

~~My suggestion is to get #485 done first~~ (**Update:** ✔️), then I'll rebase this PR and complete the removal of all obsolete targets.

PR #485, opened 2020-03-10
add targets netstandard2.0 and netstandard2.1

lg2de:
The discussion on #407 is not yet completed.
I hope this PR can revive the discussion.

BTW: The white space changes were automatically done by Visual Studio.

PR #496, opened 2020-05-01
Fix broken Linux CI build on Appveyor

stakx:
Our CI test run on Linux is currently broken due to a regression in the version of Mono that comes preinstalled in AppVeyor's `Ubuntu` worker images.

After much trial and error, I think I've found a reasonable compromise to get CI working again. I'm calling it a compromise because if we want to keep our CI build script reasonably simple, we need to give up our `net35` target now, rather than later (in a better-fitting PR like #495). Having to give up `net35` is a consequence of us now running `mono` via Docker during the Linux CI test run (so that we can choose its version without ending up in an uninstall/install orgy). It would be a little nonsensical to run Mono via Docker on the one hand, but OTOH still rely on a locally installed Mono for the .NET Framework reference assemblies. The latter can now be got from an official NuGet package (which is a much neater solution than the `FrameworkOverride` hack used so far); *that* however doesn't include any reference assemblies for .NET 3.5.

All of this isn't really a problem, since we were going to stop supporting .NET 3.5 anyway&mdash;we just have to do that now.

This PR also includes a few minor changes to our CI build scripts which aren't strictly necessary, but make it a wee bit simpler or more up-to-date.


build.sh
```diff
@@ -14,6 +14,35 @@
...
 
+set -ex
+
+# Install Mono:
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get -y install gnupg ca-certificates
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb https://download.mono-project.com/repo/ubuntu stable-bionic/snapshots/6.0.0.319 main" | tee /etc/apt/sources.list.d/mono-official-stable.list
+apt update
+DEBIAN_FRONTEND=noninteractive apt -y install mono-devel=6.0.0.*
```
stakx, line 25: It just occurred to me that we may be able to forego this whole installation orgy by simply `alias`-ing `mono` to something like `docker run --rm mono:<VERSION>`. Not having to install anything would speed up the CI build a lot. I'll try to refactor...

build.sh
```diff
@@ -14,6 +14,35 @@
...
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get -y install apt-transport-https
+DEBIAN_FRONTEND=noninteractive apt-get -y install dotnet-sdk-3.1
+
+# Install .NET Core 1.1 runtime:
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get -y install libunwind-dev
+cd /usr/share/dotnet
+wget https://download.visualstudio.microsoft.com/download/pr/d3bbce9c-a15e-4224-a271-064546459e53/694a1c880d7140c80215acb5dceab3db/dotnet-ubuntu.18.04-x64.1.1.13.tar.gz -O dotnet-ubuntu.18.04-x64.1.1.13.tar.gz
+tar xvzf dotnet-ubuntu.18.04-x64.1.1.13.tar.gz
```
stakx, line 41: It might likewise be possible to build and run using `dotnet` Docker images. While not necessary right now, it might be a relatively easy way to decouple ourselves from whatever Appveyor decides to preinstall, while keeping it relatively simple to choose a different SDK / runtime version.

stakx: This turns out to be a little more complicated because we'd run a different container for building with the .NET Core SDK 3.1 (`mcr.microsoft.com/dotnet/core/sdk:3.1`) than when running with the .NET Core 1.1 runtime (`mcr.microsoft.com/dotnet/core/runtime:1.1`), and things can go wrong if packages restored by the former aren't seen by the latter; we'd have to at least map a NuGet package restore directory.

stakx:
@jonorossi, this should fix our CI and get rid of `net35`. I'd suggest we do this first, then add `netstandard2.0` and `netstandard2.1` in #485, and finally remove `net40`, `netstandard1.3`, and `netstandard1.5` in #495.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+Deprecations:
+ - Removed support for the .NET Framework 3.5.
```
jonorossi, line 6: Do we want to release a 4.4.1 with the pending fixes before we get into a 5.0? Create a branch for 5.0?

docs/dynamicproxy-introduction.md
```diff
@@ -9,7 +9,7 @@ DynamicProxy differs from the proxy implementation built into the CLR which requ
 To use Castle DynamicProxy you need the following environment:
 
 * one of the following runtimes installed
-  * .NET Framework 3.5 SP1+
+  * .NET Framework 4.0+
   * .NET Core 1.0+ (Castle Core 4.0+)
   * Silverlight 4.0+ (Silverlight support removed in Castle Core 4.0)
```
jonorossi, line 14: Might as well remove Silverlight while you are there.

CHANGELOG.md
stakx: Sounds good. Could you publish a 4.4.1 publish now (without waiting for any further PRs&mdash;I think they can all go into a 5.0.0)? Then I'll rebase this PR right afterwards.

docs/dynamicproxy-introduction.md
stakx: Oh, right. I overlooked that one. Will remove.

CHANGELOG.md
jonorossi: Will do, #498.

docs/dynamicproxy-introduction.md
stakx: Done in 9e72dec6cd80c164dda1d634401cf4f06d38a79b.

stakx:
Rebased, and removed the Silverlight thingy, @jonorossi, I'll merge this tonight (i.e. in a few hours' time) if nothing else comes up.

jonorossi:
Sounds good to me. 5.0 here we come!

PR #500, opened 2020-05-15
Switch to NuGet `<PackageLicenseExpression>`

stakx:
Switching from `<PackageLicenseUrl>` to `<PackageLicenseExpression>` will take care of the build warnings we're currently getting for each built project:

> **warning NU5125:**
> The `licenseUrl` element will be deprecated. Consider using the `license` element instead.
> [`C:\projects\core\src\Castle.Core\Castle.Core.csproj`]

Using an SPDX license expression makes it easier for tools to recognise the license type of our project.

This will also affect how tools will display package metadata. For example, Visual Studio's NuGet Package Manager:

##### Before (with `<PackageLicenseUrl>`):
image

##### After (with `<PackageLicenseExpression>`):
image

Also, let's keep packaging the `LICENSE` file (there's no immediate reason why we shouldn't).

### DIFF
--- a/buildscripts/common.props
+++ b/buildscripts/common.props
@@ -18,7 +18,7 @@
 		<FileVersion>$(BuildVersionNoSuffix)</FileVersion>
 		<VersionPrefix>$(BuildVersion)</VersionPrefix>
 		<AssemblyVersion>$(BuildVersionMajor).0.0</AssemblyVersion>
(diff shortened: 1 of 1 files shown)


stakx:
Looks that way. I've tried with both VS2017 and VS2019. I think it is reasonable to assume that people are working with up-to-date tool versions. (If they don't, they'll run into a ton of other problems, anyway.)

PR #502, opened 2020-05-16
Deprecate Code Access Security (CAS)

### LINKED ISSUE #477: Can we remove all Code Access Security (CAS) transparency attributes from this project?
### TL;DR:

May I submit a PR to remove all applications of CAS security transparency-related attributes, i.e. `[SecurityCritical]` and `[SecuritySafeCritical]` and `[assembly: SecurityRules(SecurityRuleSet.Level2)]`?

### Reasons:

1. The attributes in question have zero effect without `[assembly: AllowPartiallyTrustedCallers]`, which got removed back in #241 (see also https://github.com/castleproject/Windsor/issues/248 and https://github.com/castleproject/Windsor/issues/300).

2. Microsoft has obsoleted large parts of CAS, and .NET Core no longer supports it at all. There's simply no more future for CAS.

3. The attributes are applied only sporadically, and as far as I can tell, in a haphazard, non-systematic manner.

4. Some of these attributes are applied in apparently pointless places, making me suspect we never really cared enough about CAS to think it through & get it right:
    - It is often `private` methods that are marked `[SecurityCritical]`, but who could call these methods except other methods in the very same class? Does it make sense that some sibling methods may call them, and others may not? Security auditing at such a fine-grained level doesn't make much sense, IMHO.
   - In one instance, a type (`static`) constructor is marked with one of these attributes. According to MS documentation, this won't have any effect at all, as type constructors are called by the runtime, which is always fully-trusted and therefore not constrained by these attributes.

5. By removing these, we could get rid of a lot of conditional code guards `#if FEATURE_SECURITY_PERMISSIONS && DOTNET40`. I'd like to especially see usage of `DOTNET40` minimized, for the reason given below.

### Some further background re: (5), `DOTNET40`

I'd love to see the next minor version of the Castle.Core NuGet package acquire two framework targets: `netcoreapp2.0` and `netstandard2.1`. <sup>**(1)**</sup> This could greatly simplify the transitive dependency graphs of .NET projects relying (directly or indirectly) on Castle.Core. Our existing `netstandard1.x` targets will bring in a lot of dependencies even when downstream libraries such as Moq, FakeItEasy, NSubstitute etc. already target `netstandard2.x`.

To support these new targets, we'd need to adjust some of our conditional compilation symbols (`#if FEATURE_*`) as the APIs of these target frameworks are once again different from all of the target frameworks we currently support (.NET Framework 3.5, 4.x, and .NET Standard 1.x).

I've been trying to determine the set of conditional compilation symbols that would work for these two new targets. Many of the existing symbols can be used without any changes; some will require changes. One symbol that's especially problematic is `DOTNET40`, it is a kitchen sink for several different things. It would be good to minimize usages of `DOTNET40` and replace it with more specific symbols wherever possible, so that these can be cleanly mapped to the various target frameworks.

One of the things that falls under the `DOTNET40` umbrella is Code Access Security transparency level 2; look for code guarded by `#if FEATURE_SECURITY_PERMISSIONS && DOTNET40`). Since the code thus guarded doesn't really make sense anymore (see reasons given above), I believe it would be easiest to just remove it rather than introducing a new `FEATURE_SECURITY_TRANSPARENCY_LEVEL2` symbol.

----

<sup>**(1)**</sup> I'd also like to see some existing target frameworks dropped&mdash;`net35`, `netstandard1.x`, `netstandard1.5`, for reasons I'll happily lay out if requested&mdash; but that would be a breaking change and cannot be done in a minor version.

stakx:
Closes #477.

This is quite a mixed bag, so as usual lately, I've done this one aspect at a time. (Perhaps we should squash this one when merging?)

We could further simplify this by removing `IEmailSender` and `DefaultSmtpSender` altogether (in a separate, preceding PR).


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Enhancements:
 
 Deprecations:
  - Removed support for the .NET Framework < 4.5 and .NET Standard 1.x. (@stakx, #495, #496)
+ - Removed support for Code Access Security (CAS). (@stakx, #502)
+ - The following public members have been removed:
+    - `Castle.Core.Internal.PermissionUtil.IsGranted` (method)
```
stakx, line 12: Note that Windsor is currently using this method, so we will have some work to do when upgrading it to Castle.Core v5.

src/Castle.Core/DynamicProxy/Generators/Emitters/ClassEmitter.cs
```diff
@@ -30,13 +30,13 @@ public class ClassEmitter : AbstractTypeEmitter
 		private readonly ModuleScope moduleScope;
 
 		public ClassEmitter(ModuleScope modulescope, String name, Type baseType, IEnumerable<Type> interfaces)
-			: this(modulescope, name, baseType, interfaces, DefaultAttributes, ShouldForceUnsigned())
+			: this(modulescope, name, baseType, interfaces, DefaultAttributes, forceUnsigned: false)
```
jonorossi, line 33: Why not drop the `forceUnsigned` parameter from the private `CreateTypeBuilder` method?

stakx: There is one code location where `true` may be passed for `forceUnsigned`:

https://github.com/castleproject/Core/blob/77e76b3c92fa7e244dbd2fb966b75f77859adf8b/src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs#L266

(It's part of the logic responsible for placing invocation types in the same module (strong/non-strong named) as the proxy type.)

At least at the surface level, it doesn't look like there is a trivial way of getting rid of that logic.

jonorossi: Ah yer, thanks. Merge away!

CHANGELOG.md
Jevonius: I'm hitting this now - not sure of the best way forward. Could bring the method into Windsor instead, or remove CAS from Windsor too, although I've not used it enough to be entirely sure of some of the consequences.

stakx: @Jevonius, I would suggest that you open an issue over at the Windsor repo (if you haven't already) so there is a dedicated place to discuss removal of CAS support. Until a final decision has been reached there, you could perhaps start by trying to keep the code changes minimal, i.e. simply reintroduce the required method over in Windsor (but keep it `private` / `internal` so it doesn't end up as an additional public API).

Jevonius: Good suggestion, thanks. There's various other CAS-related attributes dotted around, so warrants a wider discussion. Will replicate these classes for now, but internal/private as suggested.

PR #503, opened 2020-05-19
Upgrade PublicApiGenerator to version 10.1.0

stakx:
As part of our move towards version 5, we will likely want to make many of DynamicProxy's internals truly `internal` or, in some places, `private protected`. This last modifier isn't properly supported by PublicApiGenerator 9.3.0, so we need to upgrade it first before we start changing accessibilities everywhere.

This causes some benign changes to our `ref/` contract files which can be roughly classified as follows:

* slightly different ordering of type members
* `Attribute` suffix and empty parentheses omitted in custom attribute applications
* syntax for nullable value types changes from `Nullable<T>` &rarr; `T?`
* correctness: e.g. `class static` &rarr; `static class`, accessibility and modifiers are more accurate

### DIFF
--- a/ref/Castle.Core-net45.cs
+++ b/ref/Castle.Core-net45.cs
@@ -1,7 +1,7 @@
-[assembly: System.CLSCompliantAttribute(true)]
-[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"Castle.Core.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010077f5e87030dadccce6902c6adab7a987bd69cb5819991531f560785eacfc89b6fcddf6bb2a00743a7194e454c0273447fc6eec36474ba8e5a3823147d214298e4f9a631b1afee1a51ffeae4672d498f14b000e3d321453cdd8ac064de7e1cf4d222b7e81f54d4fd46725370d702a05b48738cc29d09228f1aa722ae1a9ca02fb")]
-[assembly: System.Runtime.InteropServices.ComVisibleAttribute(false)]
-[assembly: System.Runtime.Versioning.TargetFrameworkAttribute(".NETFramework,Version=v4.5", FrameworkDisplayName=".NET Framework 4.5")]
+[assembly: System.CLSCompliant(true)]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo(@"Castle.Core.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010077f5e87030dadccce6902c6adab7a987bd69cb5819991531f560785eacfc89b6fcddf6bb2a00743a7194e454c0273447fc6eec36474ba8e5a3823147d214298e4f9a631b1afee1a51ffeae4672d498f14b000e3d321453cdd8ac064de7e1cf4d222b7e81f54d4fd46725370d702a05b48738cc29d09228f1aa722ae1a9ca02fb")]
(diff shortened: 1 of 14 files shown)


ref/Castle.Core-net45.cs
```diff
@@ -1582,7 +1582,7 @@ public abstract class XmlNodeBase : Castle.Components.DictionaryAdapter.IVirtual
         public Castle.Components.DictionaryAdapter.Xml.IXmlNamespaceSource Namespaces { get; }
         public Castle.Components.DictionaryAdapter.Xml.IXmlNode Parent { get; }
         public virtual Castle.Components.DictionaryAdapter.Xml.CompiledXPath Path { get; }
-        public event System.EventHandler Realized;
+        public  virtual  event System.EventHandler Realized;
```
jonorossi, line 1585: This looks like an unexpected change.

ref/Castle.Core-net45.cs
```diff
@@ -3756,24 +3756,24 @@ public abstract class InheritanceInvocation : Castle.DynamicProxy.AbstractInvoca
         public override object InvocationTarget { get; }
         public override System.Reflection.MethodInfo MethodInvocationTarget { get; }
         public override System.Type TargetType { get; }
-        protected virtual void InvokeMethodOnTarget() { }
+        protected abstract override void InvokeMethodOnTarget() { }
```
jonorossi, line 3759: Another unexpected change?

ref/Castle.Core-net45.cs
```diff
@@ -327,7 +327,7 @@ public interface IBindingList<T> : Castle.Components.DictionaryAdapter.IBindingL
         bool SupportsChangeNotification { get; }
         bool SupportsSearching { get; }
         bool SupportsSorting { get; }
-        public event System.ComponentModel.ListChangedEventHandler ListChanged;
+        event System.ComponentModel.ListChangedEventHandler ListChanged;
```
jonorossi, line 330: Unexpected?

stakx: I'd say "correctness": this is a declaration for an interface member, so `public` is implied and shouldn't have been there to begin with.

ref/Castle.Core-net45.cs
stakx: ... but correct. `Realized` is in fact declared `virtual`.

(Sorry, my summary of the changes to our `ref/` files was indeed incomplete, I should've made that list more complete.)

ref/Castle.Core-net45.cs
stakx: Yes, but again, the change is towards the more correct declaration, [see the corresponding source code](https://github.com/castleproject/Core/blob/fae755cf2cf642068117f09711603a411ae10023/src/Castle.Core/DynamicProxy/Internal/InheritanceInvocation.cs#L50) which specifies `abstract override`, also.

ref/Castle.Core-net45.cs
jonorossi: Haha, yer it is. I've never written code like that before.

ref/Castle.Core-net45.cs
jonorossi: Makes sense since `InheritanceInvocation` is abstract.

ref/Castle.Core-net45.cs
jonorossi: Yes, I stupidly misread it as a member of `GroupAttribute` in the diff.

PR #504, opened 2020-05-20
Remove `[Obsolete]` members

stakx:
This is another preparatory step on our way to reducing the public API surface in version 5.


src/Castle.Core/DynamicProxy/IChangeProxyTarget.cs
```diff
@@ -37,23 +37,5 @@ public interface IChangeProxyTarget
...
-		///   Also while it's technically legal to pass null reference (Nothing in Visual Basic) as <paramref name = "target" />, for obvious reasons Dynamic Proxy will not be able to call the intercepted method on such target.
-		///   In this case last interceptor in the pipeline mustn't call <see cref = "IInvocation.Proceed" /> or a <see
-		///    cref = "NotImplementedException" /> will be throws.
-		///   Also while it's technically legal to pass proxy itself as <paramref name = "target" />, this would create stack overflow.
-		///   In this case last interceptor in the pipeline mustn't call <see cref = "IInvocation.Proceed" /> or a <see
-		///    cref = "InvalidOperationException" /> will be throws.
-		/// </remarks>
-		/// <exception cref = "InvalidCastException">Thrown when <paramref name = "target" /> is not assignable to the proxied type.</exception>
-		[Obsolete("Use ((IProxyTargetAccessor)invocation.Proxy).DynProxySetTarget(target) instead.")]
-		void ChangeProxyTarget(object target);
```
stakx, line 57: Since this was the name-giving method for this interface, I am wondering whether we should rename this interface to `IInvocationTargetAccessor` (in analogy to `IProxyTargetAccessor`).

Such a name change would be a little bit annoying for user code, but if we went along with that idea, we might just as well rename the `IProxyTargetAccessor.DynProxySetTarget` method to `ChangeProxyTarget`.

I'm not too familiar with these methods, so take these suggestions with a grain of salt.

buildscripts/common.props
```diff
@@ -2,7 +2,6 @@
 
 	<PropertyGroup>
 		<NoWarn>$(NoWarn);CS1591;CS3014;CS3003;CS3001;CS3021</NoWarn>
-		<NoWarn>$(NoWarn);CS0612;CS0618</NoWarn> <!-- TODO: Remove this line once `[Obsolete]` members have been dealt with. -->
```
jonorossi, line 5: Shouldn't we keep this so we can easily obsolete things in the future without having to re-add it back?

src/Castle.Core/DynamicProxy/IChangeProxyTarget.cs
jonorossi: Could you drop this commit out of the pull request (and merge the pull request).

I think we should write up some minimal docs on the purpose and usage of these interfaces, which might give us a better idea of how they should look going forward.

stakx: Sure, will do.

buildscripts/common.props
stakx: Good thought. Let's leave this in for now, then.

Issue #501, opened 2020-05-16
Remove obsolete or redundant conditional compilation symbols


stakx:
PR #495 removed approx. half of all conditional compilation symbols, but a few are left. This issue is a follow-up to @jonorossi [comment in said PR](https://github.com/castleproject/Core/pull/495#discussion_r426131150):

> After this is merged we can talk through these, more definitely need to go.

As to how more of those symbols can be removed, I would need to take a closer look. Right now, I'm assuming the following:

| Conditional compilation symbol | Remarks
| ---- | ----
| `FEATURE_SERIALIZATION` | Some serialization support was reintroduced in .NET Standard 2.0, e.g. all targeted platforms now support the `[Serializable]` attribute... but the binary serializer can't deal with e.g. `Type` nor delegates, so we can't easily enable this on .NET Core). But: This symbol is closely related to `FEATURE_APPDOMAIN` & `FEATURE_ASSEMBLYBUILDER_SAVE`, perhaps we could merge those into one.
| `FEATURE_REMOTING` | We can remove this iff we decide to give up support for remoting / `IsTransparentProxy` checks / `MarshalByRefObject`. (Is this something that's still used these days? For which types? I don't know.)
| `FEATURE_EVENTLOG` | We can remove this iff we deprecate `Castle.Core.Logging.DiagnosticsLogger`, which logs to Windows' event log.
| `FEATURE_SYSTEM_CONFIGURATION` | Tied to the facility in the `Castle.Core.Resource` namespace, which we could deprecate/remove.
| `FEATURE_TEST_PEVERIFY` | Needed for now, but I'm wondering whether we could offload the location of a PEVerify binary into the build scripts / MSBuild files. It might be neat to set an environment variable `PEVERIFY` to the PEVerify binary's location. (Setting an environment variable in an MSBuild project file might be non-trivial, but perhaps it [can be done](https://stackoverflow.com/a/20014410).)
| `FEATURE_TEST_COM` | .NET Core has some nominal COM support. It may not support functionality tied to `CoCreateInstance`, but AFAIK it *does* support COM interop and the COM attributes. So it may be possible to *reduce* usage of this attribute.

jonorossi:
- `FEATURE_SERIALIZATION`: not sure, I'd have to look into it.
- `FEATURE_REMOTING`: I can't think of a good reason anyone would want do this today, there could be old code out there using `MarshalByRefObject`, but this is definitely one of those things that isn't moving forward with .NET 5.
[...]

PR #507, opened 2020-06-08
Remove support for remoting

stakx:
This is a follow-up to https://github.com/castleproject/Core/issues/501#issuecomment-640471295, where it was mentioned that there's no future for .NET remoting.

We can keep some minimal support for .NET remoting where it doesn't cost us anything (excluding `MarshalByRefObject` methods during proxy type creation), but there will be functional changes with our `net45` target. See the commit message for details.

@jonorossi, I've never worked much with app domains and remoting, so I figured we might just gear v5's feature set towards "modern" .NET; regarding backwards compatibility (i.e. `net45`), I'll defer to your judgement whether these changes are acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Enhancements:
 Deprecations:
  - Removed support for the .NET Framework < 4.5 and .NET Standard 1.x. (@stakx, #495, #496)
  - Removed support for Code Access Security (CAS). (@stakx, #502)
+ - Removed support for Remoting. This library no longer defines any types deriving from `MarshalByRefObject`, and `ProxyUtil.IsProxy` (which used to recognize remoting/"transparent" proxies) now tests only for DynamicProxy proxies. (@stakx, #507)
  - The following public members have been removed:
     - `Castle.Core.Internal.Lock` (class) along with all related types and methods
     - `Castle.Core.Internal.PermissionUtil.IsGranted` (method)
--- a/src/Castle.Core/DynamicProxy/ProxyUtil.cs
+++ b/src/Castle.Core/DynamicProxy/ProxyUtil.cs
@@ -21,10 +21,6 @@ namespace Castle.DynamicProxy
 	using System.Runtime.CompilerServices;
 	using System.Threading;
 
-#if FEATURE_REMOTING
-	using System.Runtime.Remoting;
-#endif
-
 	using Castle.Core.Internal;
 	using Castle.DynamicProxy.Generators;
 	using Castle.DynamicProxy.Internal;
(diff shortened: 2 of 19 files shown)


src/Castle.Core/DynamicProxy/StandardInterceptor.cs
```diff
@@ -19,11 +19,7 @@ namespace Castle.DynamicProxy
 #if FEATURE_SERIALIZATION
 	[Serializable]
 #endif
```
jonorossi, line 21: **Off-topic for this review:** Hmmm. How is this class serialisable, I guess this was marked serialisable so an invocation could be serialised? An interceptor can't really be serialised because it isn't part of the contract so implementers of the interface wouldn't know to do that, and it made little sense to be marked serialisable and inherit from MBRO.

Definitely need to review all serialisation even though .NET Core 2.0 brings binary serialisation back.

src/Castle.Core/DynamicProxy/ProxyUtil.cs
```diff
@@ -128,12 +110,6 @@ public static Type GetUnproxiedType(object instance)
 
 		public static bool IsProxy(object instance)
 		{
-#if FEATURE_REMOTING
-			if (RemotingServices.IsTransparentProxy(instance))
-			{
-				return true;
-			}
-#endif
```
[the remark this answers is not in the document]
jonorossi: Happy for `IsProxy` to return false for the transparent proxy, it really isn't one of our proxies anyway. If there is someone doing the remoting dance still with the next version of DP they can pretty easily handle that check themselves.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Enhancements:
 Deprecations:
  - Removed support for the .NET Framework < 4.5 and .NET Standard 1.x. (@stakx, #495, #496)
  - Removed support for Code Access Security (CAS). (@stakx, #502)
+ - Minimized support for Remoting. This library no longer defines any types deriving from `MarshalByRefObject`, and `ProxyUtil.IsProxy` (which used to recognize remoting/"transparent" proxies) now tests only for DynamicProxy proxies. (@stakx, #507)
```
jonorossi, line 11: "Minimized support for Remoting" sounds a bit strange, we've really dropped support for .NET remoting, without MBRO you can't make those RPCs across AppDomains.

src/Castle.Core/DynamicProxy/StandardInterceptor.cs
stakx: AFAIK, invocations needn't be serializable. My guess is `CacheMappingsAtteibute` &rarr; `CacheKey` &rarr; `ProxyGenerationOptions` &rarr; `IInterceptorSelector` & `IInterceptor` & `StandardInterceptor`. There interface types in that chain obviously don't need to be marked `[Serializable]` since that attribute is not inherited, but implementing types should be. Also, it's possible that a interceptor selector stores a field of interceptors, so I suppose `StandardInterceptor` might have been declared serializable so it would be compatible with that scenario.

CHANGELOG.md
stakx: True, I suppose. I'll replace _"Minimized"_ with _"Removed"_. I'll probably keep the 2nd sentence, seems like a short explanation what exactly "removal" means *might* be mildly helpful.

PR #509, opened 2020-06-09
Remove all `FEATURE_TEST_COM` guards

stakx:
:warning: This PR depends on #510, which needs to be merged first! 

----

`FEATURE_TEST_COM` currently guards code that would compile just fine for all of our targets, so it isn't needed from a purely compile-time perspective.

Tests that use the COM subsystem (there are 2 of those) may still fail at runtime on platforms where no COM subsystem exists. ~~Unfortunately, we cannot use NUnit's `[Platform]` attribute to conditionally skip them, because that attribute... 🥁... surprisingly isn't cross-platform itself!~~ (I was wrong about that, [see below](https://github.com/castleproject/Core/pull/509#issuecomment-642234389). Because of that, the following approach is now arguably obsolete, too:)

~~Instead, I am aborting tests at runtime using `Assert.Ignore` when a COM-related failure gets detected. Regrettably, the 2 tests end up having more error detection than actual test code, but I suppose that's the price to pay for error tolerance & .NET Standard being a leaky abstraction (static contracts vs. `PlatformNotSupportedException`).~~

On the positive side, using that approach taught me that COM tests pass for CoreCLR on Windows. I wouldn't have thought so. Had I been able to use `[Platform(Exclude = ...)]` I would've excluded CoreCLR wholesale.

* [X] ~~Draft status because I still need to figure out why `PlatformNotSupportedException` doesn't get caught as it should.~~ (See https://github.com/castleproject/Core/pull/509#issuecomment-642234389 below.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,6 @@ Symbol                              | .NET 4.5           | .NET Standard 2.x
 `FEATURE_EVENTLOG`                  | :white_check_mark: | :no_entry_sign:
 `FEATURE_SERIALIZATION`             | :white_check_mark: | :no_entry_sign:
 `FEATURE_SYSTEM_CONFIGURATION`      | :white_check_mark: | :no_entry_sign:
-`FEATURE_TEST_COM`                  | :white_check_mark: | :no_entry_sign:
 `FEATURE_TEST_PEVERIFY`             | :white_check_mark: | :no_entry_sign:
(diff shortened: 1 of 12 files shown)


jonorossi:
I expected COM to work on .NET Core on Windows, it would be needed for handling VS, Office, etc.

PR #514, opened 2020-06-13
Remove `FEATURE_TEST_PEVERIFY` guards

stakx:
Turns this conditional compilation symbol into runtime checks. Whether or not PEVerify gets used should be a function of its availability during a test run, not of what target framework we compiled for.

(As an aside, this should enable a few more tests on non-Windows platforms.)

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/BasePEVerifyTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/BasePEVerifyTestCase.cs
@@ -15,14 +15,11 @@
 namespace Castle.DynamicProxy.Tests
 {
 	using System;
-#if FEATURE_ASSEMBLYBUILDER_SAVE
 	using System.Diagnostics;
-#endif
 	using System.IO;
 
 	using NUnit.Framework;
 
-#if FEATURE_TEST_PEVERIFY
 	public class FindPeVerify
 	{
 		private static readonly string[] PeVerifyProbingPaths =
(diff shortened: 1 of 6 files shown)


src/Castle.Core.Tests/DynamicProxy.Tests/BasePEVerifyTestCase.cs
```diff
@@ -55,18 +52,13 @@ private static string FindPeVerifyPath()
 					return file;
 				}
 			}
-			throw new FileNotFoundException(
-				"Please check the PeVerifyProbingPaths configuration setting and set it to the folder where peverify.exe is located");
+			return null;
```
stakx, line 55: This is the one part in this PR that I slightly worry about&mdash;it means that if you build Castle.Core on a system where PEVerify could be available, but isn't found, you won't even get a warning that tests will be less thorough. I tried replacing the `throw` with an `Assert.Warn` but that unfortunately didn't do anything.

src/Castle.Core.Tests/DynamicProxy.Tests/CustomModifiersTestCase.cs
```diff
@@ -11,7 +11,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-#if FEATURE_ASSEMBLYBUILDER_SAVE
```
stakx, line 14: There doesn't appear to be a good reason (or I cannot remember it) why the whole test fixture file was guarded instead of just the bit where we try to save & verify the dynamic assembly.

src/Castle.Core.Tests/DynamicProxy.Tests/BasePEVerifyTestCase.cs
jonorossi: Your comment in another issue about providing the path to peverify.exe via an environment variable would be the best way to do this, that means we can fail the .NET Framework CI build inside build scripts if it can't find it, and then the unit test code doesn't have the massive list of Windows SDK paths. Then if we do add ILVerify we can do it clean too.

Happy for that to be done in another pull request, maybe even after we move to GitHub Actions where we can set up a matrix build.

stakx: Glad to hear you're OK with finalizing this the proper way in a separate PR. To give some background, I didn't go straight there because I foresaw problems with IDE integration&mdash;right now, PEVerify is invoked even when you run tests from inside Visual Studio, and I wasn't sure whether this behaviour could be preserved... I expect this will take a little longer due to some MSBuild fiddling.

I'll definitely revisit this during the switch to GitHub Actions (which perhaps isn't very far off now).

PR #512, opened 2020-06-12
Remove serialization support in DynamicProxy 

stakx:
I am making this a draft to prevent accidental merging (although it's mostly ready); after all, we haven't decided yet whether to keep or retire serialization support in DynamicProxy. Until that time, consider this an "exploration" to help us better understand serialization in DynamicProxy.

This PR is perhaps best reviewed commit by commit, for the following reason:

https://github.com/castleproject/Core/issues/501#issuecomment-642314713 gave a summary of the different usages of `FEATURE_SERIALIZATION` in DynamicProxy, and appears to have been largely correct. I used it as a guideline to "undo" serialization from DynamicProxy, in the following order:

1. ~~Make all exception types serializable across all targets.~~ (Update: Moved into a separate PR #515.)

2. Change DynamicProxy so generated proxy and invocation types are no longer serializable. (~~Those two kinds of generated types cannot be easily done piecemeal because they share a common base class, `BaseProxyGenerator`.~~ Incorrect: Invocation type generators ultimately subtype `IGenerator<T>`, not `BaseProxyGenerator`.)

3. Finally, abandon `LoadAssemblyIntoCache` and get rid of the requirement that "supporting types" (hooks, interceptors, interceptor selectors, mixins) should be made serializable.

The net result is that DynamicProxy code no longer makes use of `FEATURE_SERIALIZATION` &mdash;the remaining usages are inside DictionaryAdapter and the logging facility.

Closes #367.

PR #515, opened 2020-06-15
Make all exception types serializable

stakx:
Moved out of #512 into a separate PR, as it didn't fit in too well with the other work there.

### DIFF
--- a/ref/Castle.Core-net45.cs
+++ b/ref/Castle.Core-net45.cs
@@ -2615,6 +2615,7 @@ namespace Castle.DynamicProxy
     {
         public System.Type ClassToProxy { get; }
         public System.Type ProxyType { get; }
+        public override void GetObjectData(System.Runtime.Serialization.SerializationInfo info, System.Runtime.Serialization.StreamingContext context) { }
     }
     public class MixinData
     {
--- a/src/Castle.Core/DynamicProxy/InvalidProxyConstructorArgumentsException.cs
+++ b/src/Castle.Core/DynamicProxy/InvalidProxyConstructorArgumentsException.cs
@@ -15,19 +15,33 @@
 namespace Castle.DynamicProxy
 {
 	using System;
+	using System.Runtime.Serialization;
 
-#if FEATURE_SERIALIZATION
 	[Serializable]
-#endif
 	public class InvalidProxyConstructorArgumentsException : ArgumentException
 	{
 		internal InvalidProxyConstructorArgumentsException(string message, Type proxyType, Type classToProxy) : base(message)
 		{
-			ProxyType = proxyType;
 			ClassToProxy = classToProxy;
+			ProxyType = proxyType;
+		}
+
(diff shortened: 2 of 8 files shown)


src/Castle.Core/DynamicProxy/InvalidProxyConstructorArgumentsException.cs
```diff
@@ -15,19 +15,33 @@
...
 		}
 
 		public Type ClassToProxy { get; private set; }
 		public Type ProxyType { get; private set; }
+
+		public override void GetObjectData(SerializationInfo info, StreamingContext context)
+		{
+			base.GetObjectData(info, context);
+			info.AddValue("classToProxy", ClassToProxy.AssemblyQualifiedName);
+			info.AddValue("proxyType", ProxyType.AssemblyQualifiedName);
```
jonorossi, line 44: Would it be better to use `nameof` with the properties?

ref/Castle.Core-net45.cs
```diff
@@ -2615,6 +2615,7 @@ public class InvalidProxyConstructorArgumentsException : System.ArgumentExceptio
     {
```
jonorossi, line 2615: Any idea why `SerializableAttribute` doesn't show up on these classes? PublicApiGenerator defect?

stakx: Looks like it. PublicApiGenerator is using Mono to reflect over types, and since `SerializableAttribute` is a pseudo-attribute (it gets turned into the `TypeAttributes.Serializable` bit IIRC) it won't show up like the others. I'll likely submit a PR to them.

Shall we go ahead with this (no new regression here) or wait for a fixed version of PublicApiGenerator (might take some time)?

src/Castle.Core/DynamicProxy/InvalidProxyConstructorArgumentsException.cs
stakx: My reasoning against using `nameof` was that serialized data follows its own contract that should remain as stable over time as possible (since one might persist data); if you change type or member names, that shouldn't necessarily change the contract for serialized data.

Granted, its unlikely that someone should want to persist our exception types then deserialize after many months :laughing: so I'm happy to change this to `nameof`.

ref/Castle.Core-net45.cs
jonorossi: Ah yer, forgot about it being a metadata flag.

src/Castle.Core/DynamicProxy/InvalidProxyConstructorArgumentsException.cs
jonorossi: Fair enough point, happy to leave it.

ref/Castle.Core-net45.cs
danielmarbach: PublicApiGenerator 10.1.1 is up on Nuget

jonorossi: Cheers @danielmarbach, thanks for the quick turnaround.

PR #524, opened 2020-06-25
Enable SourceLink

JTOne123:
CSProj files have been updated to enable SourceLink in your nuget
---

*[This pull request was created with an automated workflow]*

I noticed that your repository and Nuget package are important for our .NET community, but you still haven't enabled SourceLink.

**We have to take 2 steps:**
1) Please approve this pull request and make .NET a better place for .NET developers and their debugging.
2) **Then just upload the .snupkg file** to https://www.nuget.org/ (now you can find the snupkg file along with the .nuget file)

You can find more information about SourceLine at the following links  
https://github.com/dotnet/sourcelink
https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx

If you are interesting about this automated workflow and how it works  
https://github.com/JTOne123/GitHubMassUpdater

*If you notice any flaws, please comment and I will try to make fixes manually*

### DIFF
--- a/src/Castle.Core/Castle.Core.csproj
+++ b/src/Castle.Core/Castle.Core.csproj
@@ -4,7 +4,16 @@
 
 	<PropertyGroup>
 		<TargetFrameworks>net45;netstandard2.0;netstandard2.1</TargetFrameworks>
-	</PropertyGroup>
+	
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+  </ItemGroup>
+
 
 	<PropertyGroup>
 		<PackageId>Castle.Core</PackageId>


stakx:
Hi again @JTOne123, I think it would be good to enable SourceLink. I'd like to leave the final decision to @jonorossi however.

If we decide to go ahead with SourceLink here, I think the following changes will be additionally needed:

* We already build symbols packages (albeit not those of the `.snupkg` kind), see this [`<IncludeSymbols>` in `common.props`](https://github.com/castleproject/Core/blob/584a0c89cfe6a83798b81d5505e20416f53daf7a/buildscripts/common.props#L27). We should perhaps make sure that these MSBuild properties aren't spread across too many MSBuild project files. It appears to me that they belong exclusively in `Castle.Core.csproj`; we don't need debugging symbols for *every* project (such as unit test projects), after all.

* The README currently states (in the [Releases section](https://github.com/castleproject/Core/blob/master/README.md#releases)) where debugging symbols can be found. This paragraph will become obsolete and should be replaced with a hint about SourceLink (and ideally a short hint or doc link about how to enable it).

stakx:
A small P.S. question @JTOne123: [this page on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) suggests enabling deterministic builds in conjunction with SourceLink. Do you know more about this? Should we do this here, too?

jonorossi:
@stakx I have no objection to adding SourceLink support, we did talk about it a while back but the situation with symbols was a bit all over the place.

I had a very quick look at this yesterday but didn't comment. In addition to your comments, I think the properties probably belong in the next `PropertyGroup` and the automated deployment to NuGet.org needs to be changed to publish these otherwise they won't get there.

JTOne123:
https://www.meziantou.net/how-to-debug-nuget-packages-using-sourcelink.htm

src/Castle.Core/Castle.Core.csproj
```diff
@@ -4,7 +4,16 @@
 
 	<PropertyGroup>
 		<TargetFrameworks>net45;netstandard2.0;netstandard2.1</TargetFrameworks>
-	</PropertyGroup>
+	
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
stakx, line 10: Please move these properties to the following location (and make sure to use tabs for indentation, not spaces):

https://github.com/castleproject/Core/blob/a7454e5d43e29fee1451a0190b333d23b47497f1/buildscripts/common.props#L27

src/Castle.Core/Castle.Core.csproj
```diff
@@ -4,7 +4,16 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
```
stakx, line 14: Please move this `<PackageReference>` into this item group (and again, please use tabs for indentation):

https://github.com/castleproject/Core/blob/a7454e5d43e29fee1451a0190b333d23b47497f1/buildscripts/common.props#L41-L43

Issue #518, opened 2020-06-15
Can we reduce the amount of custom DynamicProxy exception types?


stakx:
DynamicProxy currently defines four exception types:

* `GeneratorException`
* `InvalidMixinConfigurationException`
* `InvalidProxyConstructorArgumentsException`
* `ProxyGenerationException`

The uses of these exceptions can be roughly summarized as follows:

* User code invokes DynamicProxy incorrectly. In many instances, a `ArgumentException` or `InvalidOperationException` could be thrown instead.
* Something truly unexpected happens, e.g. because of a possible bug. In these few circumstances, it would seem proper to throw a custom exception type.

In none of these circumstances would user code profit from catching the exceptions thrown by DynamicProxy, so there's really no point in having distinct exception types.

I may have glossed over some details in my summary above, nevertheless I would suggest that we reduce the amount of custom exception types in DP by (1) using standard FCL exception types where appropriate, and (2) using a single `ProxyGenerationException` (or perhaps better, a new, more generally named `DynamicProxyException`) for truly exceptional situations.

(If we do that, we may get rid of `GeneratorException` and thus the `Castle.DynamicProxy.Generators` sub-namespace in the public API. See #517 for the full argument.)

jonorossi:
I definitely like the idea of using FCL exceptions (e.g. `ArgumentException` and `InvalidOperationException`) instead of a custom exception because you provide your own message anyway. Why we've got `GeneratorException` and `ProxyGenerationException` it strange.

I recall DP also throws `TargetException` and `TargetInvocationException` during execution of proxies if things blow up (but they might be captured as inner exceptions), so agree there is no need for multiple custom exceptions.

If you put together a PR probably using a single `DynamicProxyException` and FCL exceptions where it makes sense I'm happy to review it.

jonorossi:
The pull requests were all great, I'm happy with all the changes. Thrown exceptions definitely follow a clean pattern now.

stakx:
@jonorossi, I'm glad you were happy with the changes. I couldn't quite shake the feeling that those PRs were a bit messy, but that was perhaps unavoidable, given the many case-by-case decisions necessary. It feels good to have those exceptions out of the way now!

jonorossi:
> I couldn't quite shake the feeling that those PRs were a bit messy, but that was perhaps unavoidable, given the many case-by-case decisions necessary.

I didn't feel that way. The commit messages made it easy to follow your thought process.

Issue #517, opened 2020-06-15
Can we move `AttributesToAvoidReplicating` into DynamicProxy's main namespace?


stakx:
[`AttributesToAvoidReplicating`](https://github.com/castleproject/Core/blob/c60b977889054452c20028c06a0aa49065a9ef81/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs) currently sits in the sub-namespace `Castle.DynamicProxy.Generators`. It's one of only two public types left in that namespace&mdash;the namespace contains mostly internals that became truly `internal` in #505.

My aim here is to concentrate DynamicProxy's public API in just one single namespace: `Castle.DynamicProxy`.

I therefore suggest that we do either of two things:

1. move `AttributesToAvoidReplicating` into the main namespace `Castle.DynamicProxy`; or
2. turn it into a collection property in `ProxyGenerationOptions`

I haven't thought about (2) too deeply, but (1) would be simple enough to do. Also, it's likely not a part of the public API that one uses a lot, so perhaps it's not worth investing too much time.

Any thoughts or opinions?

jonorossi:
I think this class got used a bit more years back when DP didn't have a framework attribute built-in and the `SecurityAttribute` hierarchy wasn't handled (just the top level), and then all the mocking frameworks added exclusions themselves rather than logging a defect.

Copying from a 2018 comment of mine:
> AttributesToAvoidReplicating was designed as a static class so for example a user of Moq could add another attribute without needing access to the ProxyGenerator instance used by Moq. I'm pretty sure it was added as a get out of jail card when you are in a pickle with some attribute causing you grief.
> (https://github.com/castleproject/Core/issues/334#issuecomment-360177220)

There will still be some use out there for things like `ServiceContractAttribute` and whatever else people have come up with, however I think if we are going to make a breaking change I'd prefer to move the 4 hardcoded attributes into `AttributeUtil` together with the opposite `ParamArrayAttribute`, and add a collection to `ProxyGenerationOptions`. I agree that it belongs on PGO not on the `ProxyGenerator` like some have suggested in GitHub issues over time.

Using PGO means that it will then be up to mocking libraries to expose their own API to exclude attributes which is a cleaner API and the way we've gone with a few other APIs over the last few years.

stakx:
> I think this class got used a bit more years back [...]

[...]

PR #527, opened 2020-06-26
Retire `InvalidMixinConfigurationException`

stakx:
After #526, this is the third of approx. four PRs (one per exception type) towards #518.


src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs
```diff
@@ -104,11 +109,24 @@ public Type CreateClassProxyType(Type classToProxy, Type[] additionalInterfacesT
...
 
 			var generator = new InterfaceProxyWithoutTargetGenerator(scope, interfaceToProxy) { Logger = logger };
 			return generator.GenerateCode(typeof(object), additionalInterfacesToProxy, options);
 		}
 
+		private void AssertValidMixins(ProxyGenerationOptions options, string paramName)
+		{
+			try
+			{
+				options.Initialize();
```
stakx, line 122: (Up until now, `options` was initialized inside the various generator classes' `GenerateCode` methods. If we're now initializing one level further up in the call hierarchy, those calls (albeit cheap) will become redundant. I'm planning to clean up redundant calls to `options.Initialize` in a future PR for #517.)

src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -92,8 +92,8 @@ public void Initialize()
 				}
 				catch (ArgumentException ex)
 				{
-					throw new InvalidMixinConfigurationException(
-						"There is a problem with the mixins added to this ProxyGenerationOptions: " + ex.Message, ex);
+					throw new InvalidOperationException(
+						"There is a problem with the mixins added to this ProxyGenerationOptions. See the inner exception for details.", ex);
```
stakx, line 96: (In case you're wondering why I'm changing the message: `ArgumentException` messages include the parameter name. When converting this `InvalidOperationException` back to an `ArgumentException` in `DefaultProxyBuilder`, we don't want the original parameter name of `MixinData`'s ctor in the message. If we kept the `+ ex.Message` bit, we'd end up with two parameter names in the final `ArgumentException`'s message.)

PR #528, opened 2020-06-26
Retire `ProxyGenerationException` & introduce `DynamicProxyException` for internal faults

### LINKED ISSUE #518: Can we reduce the amount of custom DynamicProxy exception types?

stakx:
After #527, this is the last of four PRs (one per exception type) towards #518.

Due to the many different situations where `ProxyGeneration` was used, this one turned out a little messier than the other PRs.

Closes #518 if merged together with the preceding three PRs.


src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs
```diff
@@ -144,7 +144,7 @@ public static IEnumerable<CustomAttributeInfo> GetNonInheritableAttributes(this
 							member.DeclaringType.FullName,
 							(member is TypeInfo) ? "" : ("." + member.Name),
 							typeof(AttributesToAvoidReplicating).FullName);
-					throw new ProxyGenerationException(message, e);
+					throw new DynamicProxyException(message, e);
```
stakx, line 147: This is likely still wrong, since `DynamicProxyException` is supposedly thrown for internal errors that cannot be fixed from user code. `InvalidOperationException` might be more appropriate.

stakx: I ended up adding a commit to make this a `NotSupportedException`.

PR #530, opened 2020-07-17
Added SPDX license identifiers

mburtscher:
This change adds the SPDX license identifier to the generated NuGet packages by adding the appropriate [PackageLicenseExpression](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget) tag.

**Why?** Companies use the license information provided by package managers to ensure compliance across large projects. Providing an SPDX license identifier makes it simpler for people to work with automated tools like [LicenseFinder](https://github.com/pivotal/LicenseFinder) or GitLab's license scanning feature.

### DIFF
--- a/src/Castle.Services.Logging.NLogIntegration/Castle.Services.Logging.NLogIntegration.csproj
+++ b/src/Castle.Services.Logging.NLogIntegration/Castle.Services.Logging.NLogIntegration.csproj
@@ -18,6 +18,7 @@
 		<SignAssembly>True</SignAssembly>
 		<PublicSign Condition="'$(OS)'=='Unix'">true</PublicSign>
 		<PackageTags>castle logging NLog</PackageTags>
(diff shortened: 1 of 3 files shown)


jonorossi:
We've already got `PackageLicenseExpression` in our `common.props`, is that not working for some packages?

https://github.com/castleproject/Core/blob/c60b977889054452c20028c06a0aa49065a9ef81/buildscripts/common.props#L21

mburtscher:
Hmmm, you're right. Actually it should work, however NuGet gallery is not showing the license info in the sidebar like it does for other projects. Also GitLab's license scanner (which uses LicenseFinder) did not show the correct license.

Castle.Core:
image

Newtonsoft.Json:
image

But since you're already including the tag in your `common.props` there must be a different cause.

jonorossi:
Ah, my bad. We've got it in master but haven't released it yet. See #500.

PR #540, opened 2021-01-01
Update year in copyright notices

stakx:
I never know which years to put in the copyright notice when creating a new code fie, because we have so many variations. What better day to clean this up than the 1<sup>st</sup> of January!?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2004-2016 Castle Project - http://www.castleproject.org/
+Copyright 2004-2021 Castle Project - http://www.castleproject.org/
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Debugging symbols are available in symbol packages in the AppVeyor build artifac
 
 ## License
 
-Castle Core is &copy; 2004-2020 Castle Project. It is free software, and may be redistributed under the terms of the [Apache 2.0](http://opensource.org/licenses/Apache-2.0) license.
+Castle Core is &copy; 2004-2021 Castle Project. It is free software, and may be redistributed under the terms of the [Apache 2.0](http://opensource.org/licenses/Apache-2.0) license.
 
 ## Contributing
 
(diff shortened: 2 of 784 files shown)


jonorossi (commented):
I wouldn't have approved/merged this pull request as it doesn't really make sense. If you don't know what year to put, how does bumping them all to 2021 fix this, we'll have an even worse problem in a year because we'll have to do this again.

Previous to this pull request years got updated when significant changes were made to the file, Serilog is still doing this too by the look of it. A file cannot be copyright 2021 if it hasn't been changed in 6 years.

For a few years I've considered doing what the Nancy guys did and just use the first year (https://github.com/NancyFx/Nancy/pull/2793). I too want the problem to go away so also wanted to research the legal ramifications of just referring people to the LICENSE file like the ASP.NET Core repo does which is also Apache 2.0 licensed and doesn't have a copyright year or the partial license.

I also wanted to fix the copyright owner and consider adding an `SPDX-License-Identifier` instead.

stakx:
@jonorossi My apologies. You're making a good argument, I also thought that having to change the years every year is a little annoying, but since they were already there as a range... Happy to roll this change back once you're decided on a better path.

jonorossi:
> You're making a good argument

Haha, yer I've thought about this one a bit for many years because it is a PITA, but hadn't found a nice solution until recently saw ASP.NET Core with a 2 line header.

> Happy to roll this change back once you're decided on a better path.

When you refer to rollback, do you mean `git rebase` or `git revert`? I usually wouldn't think about rebasing once in master, but if you didn't want the 2 commits in every file's history.

stakx:
>When you refer to rollback

I meant that in a non-technical way, whatever is necessary to undo the change. Technically I'd probably go with a manual change (1 step), or `git revert` (1 step + maybe another if we want to change copyrights)... rewriting the main branch doesn't seem like a good idea.

(Btw. I've also seen projects that remove the year from copyright notices entirely, claiming it isn't required. Probably depends on national laws. Keeping just the initial year would seem safer.)

jonorossi:
> rewriting the main branch doesn't seem like a good idea.

Yer true, it's been too long now, and we'll change every file again anyway.

> Btw. I've also seen projects that remove the year from copyright notices entirely, claiming it isn't required. Probably depends on national laws. Keeping just the initial year would seem safer.

I'd imagine Microsoft legal would have looked at what they are doing, but I do intend to do some reading before just blindly copying what they are doing.

Hoping we can also use the file header template thing in editorconfig so new files automatically get the header.

PR #542, opened 2021-01-01
Coding style: Prefer using C# type keywords

stakx:
This is just a small detail, but it's a little irritating to come across lines of code that e.g. contain both `String` and `string` at the same time. This switches to lower-case C# keywords (almost) everywhere.


jonorossi (commented):
@stakx glad to see them gone.

PR #544, opened 2021-01-02
Mark DynamicProxy's internal classes as `sealed`

stakx:
DynamicProxy's internals have been stable over the past few years, and now that they are no longer public (see #505), we can view their type hierarchy as essentially frozen&mdash;at least for the time being. Marking internal classes that have no subclasses of their own as `sealed` formalizes this.

Being able to recognize final classes at a glance can be helpful when we want to refactor them.

(A few refactorings are included here because the C# compiler will warn about / disallow new `protected` or `virtual` members in `sealed` types.)

**Update:** Converting to draft, this PR should perhaps also devirtualize non-overridden members, and make them `private` where possible.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/Emitters/ArgumentsUtil.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/Emitters/ArgumentsUtil.cs
@@ -20,7 +20,7 @@ namespace Castle.DynamicProxy.Generators.Emitters
 
 	using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
 
-	internal abstract class ArgumentsUtil
+	internal static class ArgumentsUtil
 	{
 		public static Expression[] ConvertArgumentReferenceToExpression(ArgumentReference[] args)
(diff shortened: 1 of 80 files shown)


stakx:
I'll close this for the moment, this turns out to be a nice refactoring (IMO) but there probably isn't as much benefit to this as I thought.

src/Castle.Core/DynamicProxy/Generators/Emitters/ArgumentsUtil.cs
```diff
@@ -20,7 +20,7 @@ namespace Castle.DynamicProxy.Generators.Emitters
 
 	using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
 
-	internal abstract class ArgumentsUtil
+	internal static class ArgumentsUtil
```
jonorossi, line 23: Would be good to at least get these `static` changes in. Shows how long ago these classes were first written to use `abstract` to prevent instantiation of the class.

PR #548, opened 2021-01-03
Remove delegate-based roundtrip from `ObtainProxyType` to `GenerateType`

stakx:
One bit in DynamicProxy that I've always found surprisingly complicated: The proxy builder calls the proxy type builder's `GenerateCode` method (somewhat inaccurate name btw.), which then calls `ObtainProxyType` in the base class, providing it with a delegate to its `GenerateType` method.

https://github.com/castleproject/Core/blob/c2428ea8b7c879155d38c06b8faae8be2561e7f7/src/Castle.Core/DynamicProxy/Generators/ClassProxyGenerator.cs#L37-L46

This PR's commits can speak for themselves, but here's a short prose summary of this refactoring anyway:

* We can simplify the issue described above by declaring `GenerateType` as an abstract method in the base class so that `ObtainProxyType` can call it directly (without the delegate scenic drive).

* This entails moving excess parameters from `GenerateType` into the ctor. Once done, we end up with duplicated code in all proxy type generator ctors that can be pushed down into the base class ctor.

* `GenerateCode` then ends up doing almost nothing besides computing a cache key, so the method is turned into `GetCacheKey`. `ObtainProxyType` becomes a public method `GetProxyType`, which is what the proxy builder now calls.

* Finally, I'm standardizing a few field and parameter names so they become easiler to distinguish and track across methods.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
@@ -66,6 +66,33 @@ namespace Castle.DynamicProxy.Generators
 			get { return scope; }
 		}
 
+		public Type GetProxyType()
+		{
+			bool notFoundInTypeCache = false;
+
+			var proxyType = Scope.TypeCache.GetOrAdd(GetCacheKey(), cacheKey =>
+			{
+				notFoundInTypeCache = true;
+				Logger.DebugFormat("No cached proxy type was found for target type {0}.", targetType.FullName);
+
+				EnsureOptionsOverrideEqualsAndGetHashCode();
+
+				var name = Scope.NamingScope.GetUniqueName("Castle.Proxies." + targetType.Name + "Proxy");
+				return GenerateType(name, Scope.NamingScope.SafeSubScope());
+			});
+
+			if (!notFoundInTypeCache)
+			{
+				Logger.DebugFormat("Found cached proxy type {0} for target type {1}.", proxyType.FullName, targetType.FullName);
+			}
+
+			return proxyType;
+		}
+
+		protected abstract CacheKey GetCacheKey();
+
+		protected abstract Type GenerateType(string name, INamingScope namingScope);
+
 		protected void AddMapping(Type @interface, ITypeContributor implementer, IDictionary<Type, ITypeContributor> mapping)
 		{
 			Debug.Assert(implementer != null, "implementer != null");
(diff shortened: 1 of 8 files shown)


src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
```diff
@@ -38,14 +38,19 @@ namespace Castle.DynamicProxy.Generators
 	internal abstract class BaseProxyGenerator
 	{
 		protected readonly Type targetType;
+		protected readonly Type[] interfaces;
 		private readonly ModuleScope scope;
 		private ILogger logger = NullLogger.Instance;
 		private ProxyGenerationOptions proxyGenerationOptions;
 
-		protected BaseProxyGenerator(ModuleScope scope, Type targetType, ProxyGenerationOptions proxyGenerationOptions)
+		protected BaseProxyGenerator(ModuleScope scope, Type targetType, Type[] interfaces, ProxyGenerationOptions proxyGenerationOptions)
```
stakx, line 46: Additional interfaces are allowed for all proxy types, so it makes sense that the base class asks for them.

src/Castle.Core/DynamicProxy/Generators/BaseProxyGenerator.cs
```diff
@@ -38,14 +38,19 @@ namespace Castle.DynamicProxy.Generators
...
 		{
+			CheckNotGenericTypeDefinition(targetType, "targetType");
+			CheckNotGenericTypeDefinitions(interfaces, "interfaces");
+
 			this.scope = scope;
 			this.targetType = targetType;
+			this.interfaces = TypeUtil.GetAllInterfaces(interfaces);
```
stakx, line 53: Perhaps worth noting that one proxy type generator previously made this call *before*, not *after*, `CheckNotGenericTypeDefinitions`. I standardized the call order as seen here. Calling `TypeUtil.GetAllInterfaces` after the check should be more efficient since there are fewer interfaces to check; it should also be perfectly safe since a closed generic interface type cannot inherit an open generic interface type (AFAIK).

src/Castle.Core/DynamicProxy/Generators/ClassProxyGenerator.cs
```diff
@@ -113,7 +108,6 @@ protected virtual Type GenerateType(string name, Type[] interfaces, INamingScope
 			// target is not an interface so we do nothing
 
 			var targetInterfaces = targetType.GetAllInterfaces();
-			var additionalInterfaces = TypeUtil.GetAllInterfaces(interfaces);
```
stakx, line 116: This call is superfluous since the same transform already happens in the `BaseProxyGenerator` ctor.

src/Castle.Core/DynamicProxy/Generators/InterfaceProxyWithTargetGenerator.cs
```diff
@@ -164,9 +162,7 @@ protected virtual Type GenerateType(string typeName, Type proxyTargetType, Type[
 			// Order of interface precedence:
 			// 1. first target
 			var targetInterfaces = proxyTargetType.GetAllInterfaces();
-			var additionalInterfaces = TypeUtil.GetAllInterfaces(interfaces);
```
stakx, line 167: This call is also superfluous (see other review comment).

stakx (commented):
Just a few additional comments.

src/Castle.Core/DynamicProxy/Generators/ClassProxyGenerator.cs
```diff
@@ -27,25 +27,21 @@ namespace Castle.DynamicProxy.Generators
...
 		{
-			interfaces = TypeUtil.GetAllInterfaces(interfaces);
-			CheckNotGenericTypeDefinitions(interfaces, "interfaces");
-			var cacheKey = new CacheKey(targetType, interfaces, ProxyGenerationOptions);
-			return ObtainProxyType(cacheKey, (n, s) => GenerateType(n, interfaces, s));
+			return new CacheKey(targetType, interfaces, ProxyGenerationOptions);
 		}
 
-		protected virtual Type GenerateType(string name, Type[] interfaces, INamingScope namingScope)
+		protected override Type GenerateType(string name, INamingScope namingScope)
```
jonorossi, line 41: This would have made `ClassProxyGenerator` immutable if `ProxyGenerationOptions` was immutable.

jonorossi (commented):
Very nice. Never understood why `GenerateCode` had the delegate, just assumed it was for a good reason.

PR #549, opened 2021-01-04
Extract abstract base class `BaseClassProxyGenerator`

stakx:
This is a refactoring that extracts a new common base class for both `ClassProxyGenerator` and `ClassProxyWithTargetGenerator`, which are mostly identical.

The first few commits represent single edits that work towards minimizing the diff between the two types' code files. Good for reviewing, but they can be squashed when merging.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
@@ -1,0 +1,204 @@
+// Copyright 2004-2021 Castle Project - http://www.castleproject.org/
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Generators
+{
+	using System;
+	using System.Collections.Generic;
+	using System.Linq;
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Contributors;
+	using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
+	using Castle.DynamicProxy.Internal;
+
+	internal abstract class BaseClassProxyGenerator : BaseProxyGenerator
+	{
+		protected BaseClassProxyGenerator(ModuleScope scope, Type targetType, Type[] interfaces, ProxyGenerationOptions options)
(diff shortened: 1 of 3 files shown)


src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+
+			CreateFields(emitter);
+			CreateTypeAttributes(emitter);
+
+			// Constructor
+			var cctor = GenerateStaticConstructor(emitter);
+
+			var constructorArguments = new List<FieldReference>();
+
+			if (TargetField is { } targetField)
```
jonorossi, line 63: Is this basically checking if `TargetField != null`?

src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
```diff
@@ -0,0 +1,204 @@
...
+	using System;
+	using System.Collections.Generic;
+	using System.Linq;
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Contributors;
+	using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
+	using Castle.DynamicProxy.Internal;
+
+	internal abstract class BaseClassProxyGenerator : BaseProxyGenerator
```
jonorossi, line 26: I'd have probably suggested `ClassProxyGeneratorBase` as this class sounds a bit like something else at the moment, also means it'll sort next to class proxy generators.

jonorossi (commented):
Nice work! A couple of comments.

src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
stakx: Having a single naming theme for abstract base classes instead of many (`AbstractX`, `BaseX`) is another refactor we should probably do, and I also tend towards `XBase`. Here, I just didn't want to introduce a third name form.

src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
stakx: Good catch... yes.

Btw. if you find more of those minor mess ups that may seem too small to fix individually, perhaps I'll just send one final PR that fixes them all together. Let me know which ones you want changed.

src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
jonorossi: Fair enough. I too usually go for `XBase`. `AbstractX` would come from the early days when most people were doing Java and building Castle.

src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
jonorossi: No problem. I've added a few comments as I go. Happy for you to make one PR with a bunch of fixes.

PR #550, opened 2021-01-04
Extract abstract base class `BaseInterfaceProxyGenerator`

stakx:
This is similar to my previous PR #549 which extracted a base class for class proxy generators. Only this time, we start off with extracting the base class, then `InterfaceProxyWithTargetGenerator` specifics are pushed into the subtype.

Again, the commits represent single edit steps (for easier reviewing), they can be squashed when merging.


jonorossi (commented):
Legend 🥇 . That mess has annoyed me for years, but I've never attempted to clean it up because it always seemed like there were so many small differences between each generator.

Thanks for making it so easy to code review.

stakx:
Same general feeling here. And there are probably similar opportunities remaining, in fact the two new base proxy generator classes share quite a few bits in common. It seems entirely possible to eventually merge them right into `BaseProxyGenerator`. My only worry is that the base class may become way too large (but that may be only an intermediate state), it would probably have to be restructured in _some_ way.

jonorossi:
Probably need more composition and less inheritance. The changes you made to separate one contributor into 3 for their specific tasks is probably more the direction things should go.

PR #551, opened 2021-01-04
Use `nameof` operator in argument validation

stakx:
Including code other than DynamicProxy's this time, since it's not much additional effort.


jonorossi (commented):
Nice clean up. I thought you might have found one out of place (i.e. misspelt).

PR #553, opened 2021-01-05
Split up instance contributors into 3 single-purpose contributors

stakx:
`ProxyInstanceContributor` and its subtypes currently have three responsibilities:

1. adding an implementation for `ISerializable`
2. adding an implementation for `IProxyTargetAccessor`
3. reproducing the proxied type's non-inheritable custom attributes on the proxy type

This PR takes that type apart into three single-purpose contributors `SerializableContributor`, `ProxyTargetAccessorContributor`, and `NonInheritableAttributesContributor`, where each one is responsible for the aspect it is named after.


src/Castle.Core/DynamicProxy/Contributors/ProxyTargetAccessorContributor.cs
```diff
@@ -0,0 +1,74 @@
...
+			{
+				dynProxySetTarget.CodeBuilder.AddStatement(
+					new AssignStatement(targetField,
+						new ConvertExpression(targetField.Fieldbuilder.FieldType, dynProxySetTarget.Arguments[0].ToExpression())));
+			}
+			else
+			{
+				dynProxySetTarget.CodeBuilder.AddStatement(
+					new ThrowStatement(typeof(InvalidOperationException), "Cannot change the target of the class proxy."));
+			}
```
stakx, line 64: Both the code comment and the exception message are somewhat misleading, since there are interface proxies that also don't support changing the target. While even interface proxies without a target have a `__target` field, setting it has no effect whatsoever.

This is the original implementation from `ProxyInstanceContributor` and I opted not to change it for now, but perhaps in the future, we want to formally disallow changing the target of interface proxies without a target...?

stakx (commented):
Just one thing for future consideration.

jonorossi (commented):
Very nice.

PR #554, opened 2021-01-06
Prevent double `GetObjectData` implementation for badly implemented serializable classes

### LINKED ISSUE #367: Proxying certain [Serializable] classes produces proxy types that fail PEVerify test
Using Castle Core 4.2.1, proxying the following type (or `System.Security.Claims.ClaimsIdentity`, if you want a more realistic example) produces a proxy class with incorrect IL:

```csharp
[Serializable]
public class BadSerializable
{
    public BadSerializable() { }
    public BadSerializable(SerializationInfo info, StreamingContext context) { }
    public virtual void GetObjectData(SerializationInfo info, StreamingContext context) { }
}

generator.CreateClassProxy<BadSerializable>(...);
```

When running PEVerify for the dynamically generated assembly, it complains about duplicate methods:

```
[MD]: Error: Method has a duplicate, token=0x06000004. [token:0x06000003]
[MD]: Error: Method has a duplicate, token=0x06000003. [token:0x06000004]
2 Error(s) Verifying Castle.Core\src\Castle.Core.Tests\bin\Debug\net461\CastleDynProxy2.dll
```

The mentioned methods belong to the generated proxy class `BadSerializableProxy`.

Below you'll find the two methods as ILSpy decompiles them back to C#:

```csharp
// Castle.Proxies.BadSerializableProxy (method token 06000003):
public override void GetObjectData(SerializationInfo info, StreamingContext context)
{
    BadSerializable_GetObjectData badSerializable_GetObjectData = new BadSerializable_GetObjectData(typeof(SerializableClassTestCase.BadSerializable), this, this.__interceptors, BadSerializableProxy.token_GetObjectData, new object[]
    {
        info,
        context
    });
    badSerializable_GetObjectData.Proceed();
}

// Castle.Proxies.BadSerializableProxy (method token 06000004):
public override void GetObjectData(SerializationInfo serializationInfo, StreamingContext streamingContext)
{
    Type type = Type.GetType("Castle.DynamicProxy.Serialization.ProxyObjectReference, Castle.Core, Version=0.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc", true, false);
    serializationInfo.SetType(type);
    serializationInfo.AddValue("__interceptors", this.__interceptors);
    string[] value = new string[0];
    serializationInfo.AddValue("__interfaces", value);
    serializationInfo.AddValue("__baseType", "Castle.DynamicProxy.Tests.SerializableClassTestCase+BadSerializable, Castle.Core.Tests, Version=0.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc");
    serializationInfo.AddValue("__proxyGenerationOptions", BadSerializableProxy.proxyGenerationOptions);
    serializationInfo.AddValue("__proxyTypeId", "class");
    serializationInfo.AddValue("__delegateToBase", false);
    MemberInfo[] array = FormatterServices.GetSerializableMembers(typeof(SerializableClassTestCase.BadSerializable));
    array = TypeUtil.Sort(array);
    object[] objectData = FormatterServices.GetObjectData(this, array);
    serializationInfo.AddValue("__data", objectData);
}
```

This error does not happen if either:

* the class is not decorated with `[Serializable]`
* implements the `ISerializable` interface
* `GetObjectData` is not declared `virtual`

stakx:
Fixes #367. As an additional bonus, gets rid of this:

https://github.com/castleproject/Core/blob/9143de2e453347f533df8ebc9129d9cb6e05e5a5/src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs#L112

See commit messages and code comments for details.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
@@ -44,7 +44,37 @@ namespace Castle.DynamicProxy.Contributors
 
 		public override void CollectElementsToProxy(IProxyGenerationHook hook, MetaType model)
 		{
-			delegateToBaseGetObjectData = VerifyIfBaseImplementsGetObjectData(targetType, model);
+			delegateToBaseGetObjectData = VerifyIfBaseImplementsGetObjectData(targetType, model, out var getObjectData);
+
+			// This contributor is going to add a `GetObjectData` method to the proxy type.
+			// If a method with the same name and signature exists in the proxied class type,
+			// and another contributor has decided to proxy it, we need to tell it not to.
+			// Otherwise, we'll end up with two implementations!
+
+			if (getObjectData == null)
+			{
+				// `VerifyIfBaseImplementsGetObjectData` only searches for `GetObjectData`
+				// in the implementation map for `ISerializable`. In the best case, it was
+				// already found there. If not, we need to look again, since *any* method
+				// with the same signature is a problem.
+
+				var getObjectDataMethod = targetType.GetMethod(
+					"GetObjectData",
+					BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance,
+					null,
+					new[] { typeof(SerializationInfo), typeof(StreamingContext) },
+					null);
+
+				if (getObjectDataMethod != null)
+				{
+					getObjectData = model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod);
+				}
+			}
+
+			if (getObjectData != null && getObjectData.Proxyable)
+			{
+				getObjectData.Ignore = true;
+			}
 		}
 
 		public override void Generate(ClassEmitter @class)
@@ -161,8 +191,10 @@ namespace Castle.DynamicProxy.Contributors
 			ctor.CodeBuilder.AddStatement(new ReturnStatement());
 		}
 
-		private bool VerifyIfBaseImplementsGetObjectData(Type baseType, MetaType model)
+		private bool VerifyIfBaseImplementsGetObjectData(Type baseType, MetaType model, out MetaMethod getObjectData)
 		{
+			getObjectData = null;
+
 			if (!typeof(ISerializable).IsAssignableFrom(baseType))
 			{
 				return false;
@@ -191,11 +223,7 @@ namespace Castle.DynamicProxy.Contributors
 				throw new ArgumentException(message);
 			}
 
-			var getObjectData = model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod);
-			if (getObjectData != null && getObjectData.Proxyable)
-			{
-				getObjectData.Ignore = true;
-			}
+			getObjectData = model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod);
 
 			serializationConstructor = baseType.GetConstructor(
 				BindingFlags.Instance | BindingFlags.Public |
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
@@ -19,9 +19,11 @@ namespace Castle.DynamicProxy.Contributors
 	using System;
 	using System.Collections.Generic;
 	using System.Diagnostics;
+	using System.Linq;
 	using System.Reflection;
 	using System.Runtime.Serialization;
 
+	using Castle.DynamicProxy.Generators;
 	using Castle.DynamicProxy.Generators.Emitters;
 	using Castle.DynamicProxy.Generators.Emitters.CodeBuilders;
 	using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
@@ -30,17 +32,19 @@ namespace Castle.DynamicProxy.Contributors
 
 	internal class ClassProxySerializableContributor : SerializableContributor
 	{
-		private readonly bool delegateToBaseGetObjectData;
+		private bool delegateToBaseGetObjectData;
 		private ConstructorInfo serializationConstructor;
 		private readonly IList<FieldReference> serializedFields = new List<FieldReference>();
 
-		public ClassProxySerializableContributor(Type targetType, IList<MethodInfo> methodsToSkip, Type[] interfaces,
-		                                     string typeId)
+		public ClassProxySerializableContributor(Type targetType, Type[] interfaces, string typeId)
 			: base(targetType, interfaces, typeId)
 		{
 			Debug.Assert(targetType.IsSerializable, "This contributor is intended for serializable types only.");
+		}
 
-			delegateToBaseGetObjectData = VerifyIfBaseImplementsGetObjectData(targetType, methodsToSkip);
+		public override void CollectElementsToProxy(IProxyGenerationHook hook, MetaType model)
+		{
+			delegateToBaseGetObjectData = VerifyIfBaseImplementsGetObjectData(targetType, model);
 		}
 
 		public override void Generate(ClassEmitter @class)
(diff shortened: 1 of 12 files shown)


src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
```diff
@@ -30,28 +31,22 @@ namespace Castle.DynamicProxy.Contributors
...
 		public ClassProxySerializableContributor(Type targetType, IList<MethodInfo> methodsToSkip, Type[] interfaces,
 		                                     string typeId)
 			: base(targetType, interfaces, typeId)
 		{
-			if (targetType.IsSerializable)
-			{
-				implementISerializable = true;
-				delegateToBaseGetObjectData = VerifyIfBaseImplementsGetObjectData(targetType, methodsToSkip);
-			}
+			Debug.Assert(targetType.IsSerializable, "This contributor is intended for serializable types only.");
```
stakx, line 42: This is fine because of this prior check:

https://github.com/castleproject/Core/blob/9143de2e453347f533df8ebc9129d9cb6e05e5a5/src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs#L184-L189

src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
```diff
@@ -58,7 +55,7 @@ protected override IEnumerable<MembersCollector> CollectElementsToProxyInternal(
 		protected override MethodGenerator GetMethodGenerator(MetaMethod method, ClassEmitter @class,
 		                                                      OverrideMethodDelegate overrideMethod)
 		{
-			if (methodsToSkip.Contains(method.Method))
+			if (method.Ignore)
 			{
 				return null;
 			}
```
stakx, line 61: Such checks are missing in all interface target contributors, likely because they were never needed in the past. So this is a minor inconsistency, and perhaps we'll fix that someday... I chose to leave those other contributors as is for now.

src/Castle.Core/DynamicProxy/Generators/MetaMethod.cs
```diff
@@ -52,6 +52,8 @@ public string Name
 			get { return name; }
 		}
 
+		public bool Ignore { get; internal set; }
+
 		public bool Proxyable { get; private set; }
 
 		public bool Standalone { get; private set; }
```
stakx, line 59: As an aside unrelated to this PR, we have multiple `bool` flags in `MetaMethod`. I haven't checked how the runtime lays this type out in memory, but I'm assuming /hoping that the flags are packed efficiently. If they ever turn out to blow up this type's size in memory, we can always combine them into a single flags enum.

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
```diff
@@ -187,7 +191,11 @@ private bool VerifyIfBaseImplementsGetObjectData(Type baseType, IList<MethodInfo
 				throw new ArgumentException(message);
 			}
 
-			methodsToSkip.Add(getObjectDataMethod);
+			if (model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod) is { } method &&
+				method.Proxyable)
+			{
+				method.Ignore = true;
+			}
```
stakx: This was the _only_ place where anything ever got added to `methodsToSkip`, btw.

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
```diff
@@ -29,29 +32,55 @@ namespace Castle.DynamicProxy.Contributors
...
+				var getObjectDataMethod = targetType.GetMethod(
+					"GetObjectData",
+					BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance,
+					null,
+					new[] { typeof(SerializationInfo), typeof(StreamingContext) },
+					null);
+
+				if (getObjectDataMethod != null)
+				{
+					getObjectData = model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod);
```
jonorossi, line 70: I know we've had a few cases of people reporting performance problems when they've got classes with hundreds or thousands of methods, yep some crazy people, but it'd be pretty quick to add a `MethodInfo` index something like this. At minimum you'd get rid of the Linq extension method usage which hurts performance.

```csharp
internal class MetaType
{
    private readonly Dictionary<MethodInfo, MetaMethod> methodsIndex = new Dictionary<MethodInfo, MetaMethod>();

    public MetaMethod FindMethod(MethodInfo info)
    {
        methodsIndex.TryGetValue(info, out var MetaMethod method);
        return method;
    }
}
```

jonorossi (changes requested):

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
```diff
@@ -187,7 +191,11 @@ private bool VerifyIfBaseImplementsGetObjectData(Type baseType, IList<MethodInfo
 				throw new ArgumentException(message);
 			}
 
-			methodsToSkip.Add(getObjectDataMethod);
+			if (model.Methods.FirstOrDefault(m => m.Method == getObjectDataMethod) is { } method &&
```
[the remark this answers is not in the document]
stakx: First, note that this line of code goes away in a later commit so any changes will only remain in a historical commit.

That said, I'm happy to change to a different syntax. I don't yet know your preference regarding pattern matching syntax, so this is probably a good opportunity to briefly discuss it.

I chose the syntax for brevity. I needed a way to name the queried `MetaMethod` because it's subsequently needed (`method.Ignore = true`), thus the `is … method` syntax. I chose `is { }` over `is MetaMethod` for brevity, and I chose it over `is var` for correctness as the latter also matches `null`.

* The most terse syntax would be (I think) `… is { Proxyable: true } method`. I didn't think of that at the time.
* The classic way would be: `var method = model.Methods.FirstOrDefault(…); if (method != null && method.Proxyable) …` or `… if (method?.Proxyable == true)`.

Let me know which you prefer.

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
stakx: Done. We could go even further and merge `methodsIndex` with `methods` (which is a `TypeElementCollection`). Noone actually ever queries `MetaType.Methods`, it's only there to prevent method name conflicts across all contributors. So the type of collection used internally doesn't really matter.

However I suggest we do that separately, there are a few other places in DP where we could optimize the type of collection used.

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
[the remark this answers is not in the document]
jonorossi: Oops, yer I meant to remove that comment, I had added it on a commit but it didn't show up when I got to the full diff to remove it.

I haven't used pattern matching much, too many clients still with either older toolsets or teams not using it, so I've avoided being fancy. I see exactly why you used it now, thanks for the explanation, I usually put the Linq extension methods query into a local on the line before an if-statement (the classic way) because the intent is clear especially because the lambda makes the if-statement already long, but I'm sure I could get used to this.

src/Castle.Core/DynamicProxy/Contributors/ClassProxySerializableContributor.cs
jonorossi: Wasn't line 70 meant to use `FindMethod` too?

stakx: Uhh, yes, but then confusion struck. 🤦 Sorry about that. Fixed.

jonorossi (approved):

PR #555, opened 2021-01-06
Remove `Castle.Core.Pair<,>`

stakx:
Given `System.ValueTuple<,>`, `System.Tuple<,>`, and recent C# language developments, I don't think we should encourage people to use our `Pair<,>` class.

OK with you, @jonorossi?


jonorossi (approved):
Agreed. There's `struct KeyValuePair` too, but tuple is definitely the recommendation. I'm not sure why we've even got it. My guess is that it pre-dates .NET 2.0 and was a class with 2 `object` fields which was converted to generic after .NET 2.0, but kept because someone wanted a class rather than a struct, `Tuple<,>` came later.

Someone will have to fix Windsor, there is some implementation usage and one public API that uses it for diagnostics.

PR #557, opened 2021-01-06
Make `ModuleScope` internal

### LINKED ISSUE #389: Reduce public API surface

stakx:
Following up from our discussion in https://github.com/castleproject/Core/issues/389#issuecomment-515923094, I'm proposing to make `ModuleScope` internal. This affects mainly `SaveAssembly` and `LoadAssemblyIntoCache`, as everything else is already internal. Those methods need new public entrypoints.

### Loading a previously generated assembly into the type cache

This would now be done via `IProxyBuilderWithCache.LoadAssemblyIntoCache(Assembly)`, which is implemented in `DefaultProxyBuilder`.

Introducing a new interface `IProxyBuilderWithCache` feels like a clean solution as long as `IProxyBuilder` is a public abstraction... there could be user implementations out there (unlikely, I know) that don't have a type cache. But it might also be overkill. Perhaps it's OK if such implementations simply throw `NotSupportedException` for the new load method; then the method could sit directly in `IProxyBuilder`.

### Saving a generated assembly

This can already be done through `PersistentProxyBuilder` (on platforms supporting `AssemblyBuilder.Save`).

However, there would now be no way to specify the desired target file path, nor the assembly's internal name. We might want to expose those capabilities through different overloads, e.g.

```diff
-string SaveAssembly();
+void SaveAssembly(string path = ModuleScope.DEFAULT_FILE_NAME);
+void SaveAssembly(string name, string path = ModuleScope.DEFAULT_FILE_NAME);
```

Closes #389. There's still a few bits we could remove from the public API (in some cases we already have separate issues), but by and large, we're done.

P.S.: This assumes `IProxyBuilder` stays public. It if shouldn't, everything here would have to move one level up (into `IProxyGenerator`, where it admittedly doesn't fit quite as nicely from a conceptual POV).


src/Castle.Core/DynamicProxy/IProxyBuilderWithCache.cs
```diff
@@ -0,0 +1,46 @@
...
+	using System.Reflection;
+
+	using Castle.DynamicProxy.Serialization;
+
+	/// <summary>
+	///   A <see cref="IProxyBuilder"/> that caches generated proxy types for reuse.
+	/// </summary>
+	public interface IProxyBuilderWithCache : IProxyBuilder
+	{
+#if FEATURE_SERIALIZATION
```
stakx, line 26: Due to the ordering of this `#if`, targets that don't support full binary serialization still get this interface (but it will be empty). I chose this&mdash;rather than erasing the interface as a whole&mdash;because it is true on all targets to say that "`DefaultProxyBuilder` is a builder with a cache", even when you cannot batch-populate it. It also gives us more API uniformity, as `DefaultProxyBuilder` will implement the same interfaces for all targets.

Issue #497, opened 2020-05-06
Move from AppVeyor to GitHub Actions


jonorossi:
I'm using GitHub Actions for a couple of private projects and have been happy. The integration into the GitHub UI is nicer, we'd be able to have more concurrent jobs which does get annoying with AppVeyor being limited to just one. I haven't tried out GitHub Packages, but I'd be interested in seeing if that would work well for CI builds.

Today we build and test on .NET Framework and .NET Core on Windows, and Mono and .NET Core on Linux, however it would be good to build our .NET Standard package once, and test it across all 3 platforms. GitHub Actions jobs can store files for another job for matrix builds.

Thoughts?

stakx:
While fixing our AppVeyor CI build, I was about to suggest moving to GitHub Actions, too. I ended up not to, because in the short term, I'd rather invest my time in getting the next 1-2 versions of Castle.Core ready, rather than rewriting the CI build. 😄

Once we have Castle.Core 5 done and published, I think moving to GitHub Actions would be a good idea. When we moved our Linux build from Travis to Appveyor, I thought it would be nice that Appveyor automatically manages & updates Mono for us. As the current breakage shows, that isn't so nice after all, and having a little bit more control over build-time dependencies would be good. GitHub Actions might give us that. (Alternatively, we could just reduce our use of AppVeyor as a Docker trampoline.)

TL;DR: I'm all for GitHub Actions, but would be glad if we could get Castle.Core v5 released first.

jonorossi:
[Dustin Moris Gorski's blog post](https://dusted.codes/github-actions-for-dotnet-core-nuget-packages) yesterday might be useful, he even has the Giraffe project builds publishing to GitHub Packages.

PR #559, opened 2021-01-12
Run build and tests on GitHub Actions

stakx:
Let's get started on migrating our CI build from Appveyor to GitHub Actions (#497). I haven't got very far yet. At the moment, this includes the build and test runs&mdash;with the addition of .NET Core test runs on MacOS&mdash;but no NuGet deployment-specific build steps.

You can [see this in action over at my fork](https://github.com/stakx/Castle.Core/actions?query=workflow%3ABuild).

I haven't yet found an easy way to only build once, then reuse the build output for different OS builds. The reason is that a build will not copy all required binaries from NuGet packages to the output folder; one would need to `dotnet publish --self-contained` to get complete output. A self-contained publish however produces runtime & platform specific output which is not what we want. Only speeding up the NuGet package restore by caching the package folder across OS builds doesn't help a lot time-wise due to the relatively large data volume (> 0.5 GB). Due to all of this, I suggest we do what's most straightforward and perform a complete restore, build, and test run cycle for every OS.

What's still missing? (Feel free to add to this list.)

* [ ] Running our ExplicitVersions tool to sync the auxiliary packages' versions with the main package's version.
* [ ] Anything else NuGet deployment related.
* [ ] Further tests to ensure that the build fails under certain conditions:
      - failed unit test
      - failed PEVerify execution
* [x] Structuring the workflow definition file a little nicer
* [x] ~~Possibly separating this commit into several commits~~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,12 +46,6 @@ jobs:
       # Build
       # -----
 
-      # This step is a workaround for dotnet/core#5881.
-      # It was originally described in actions/setup-dotnet#155.
-      # Without it, the NuGet package restore may fail unpredictably.
-      - name: Clean
-        run: dotnet clean ./Castle.Core.sln --configuration Release && dotnet nuget locals all --clear
-
       - name: Restore NuGet packages
         run: dotnet restore
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,78 @@
+name: Build
+
+on: push
+
+env:
+  MONO_TAG: "6.0.0.334"
+
+jobs:
+
+  build-and-test:
+    name: Build and test
+
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Several .NET Core versions will be used during the test run.
+      # The lowest version gets installed first in order to prevent
+      # "a newer version is already installed" install errors.
+
+      - name: Install .NET Core 2.1
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 2.1.x
+
+      - name: Install .NET Core 3.1
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1.x
+
+      # Building requires an up-to-date .NET SDK.
+
+      - name: Install .NET 5.0
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
+
+      # -----
+      # Build
+      # -----
+
+      # This step is a workaround for dotnet/core#5881.
+      # It was originally described in actions/setup-dotnet#155.
+      # Without it, the NuGet package restore may fail unpredictably.
+      - name: Clean
+        run: dotnet clean ./Castle.Core.sln --configuration Release && dotnet nuget locals all --clear
+
+      - name: Restore NuGet packages
+        run: dotnet restore
+
+      - name: Build all targets
+        run: dotnet build -c Release --no-restore
+
+      # ----
+      # Test
+      # ----
+
+      - name: Test on .NET Core 2.1
+        run: dotnet test -c Release -f netcoreapp2.1 --no-build --no-restore -l "console;verbosity=detailed"
+
+      - name: Test on .NET Core 3.1
+        run: dotnet test -c Release -f netcoreapp3.1 --no-build --no-restore -l "console;verbosity=detailed"
+
+      - name: Test on .NET Framework 4.6.1 (Windows only)
+        if: matrix.os == 'windows-latest'
+        run: dotnet test -c Release -f net461 --no-build --no-restore -l "console;verbosity=detailed"
+
+      - name: Test on .NET Framework 4.6.1 using Mono (Linux only)
+        if: matrix.os == 'ubuntu-latest'
+        run: |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -26,17 +26,21 @@ Browse the [contributing section](https://github.com/castleproject/Home#its-comm
 
 | Platforms       | Build Status | NuGet Feed |
 |-----------------|-------------:|------------|
-| Windows & Linux | [![Build status](https://ci.appveyor.com/api/projects/status/49a6i0ydiku56r5b/branch/master?svg=true)](https://ci.appveyor.com/project/castleproject/core/branch/master) | [Preview Feed](https://ci.appveyor.com/nuget/core-0mhe40ifodk8)
+| Windows & Linux | [![GitHub Actions](https://github.com/castleproject/Core/workflows/Build%20and%20test/badge.svg?branch=github-actions)](https://github.com/castleproject/Core/actions?query=workflow%3A%22Build%22) | :construction: **TODO** :construction:
```
jonorossi, line 29: I'd just get rid of the build status because it is integrated into the GitHub UI with the little coloured icons against commits and branches. This just becomes misleading because it is hardcoded to a specific branch.

.github/workflows/build.yml
```diff
@@ -0,0 +1,54 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
jonorossi, line 9: Any reason you aren't building other branches? i.e. feature branches

.github/workflows/build.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/checkout@v2
+      - name: Install .NET Core 2.1
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 2.1.x
+      - name: Install .NET Core 3.1
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1.x
+      - name: Install .NET Core 5.0
```
jonorossi, line 33:
```suggestion
      - name: Install .NET 5.0
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 5.0.x
+      - name: Clean
+        run: dotnet clean ./Castle.Core.sln --configuration Release && dotnet nuget locals all --clear
```
jonorossi, line 38: Do we need to "clean" in a CI build? If this is to run it locally, then `dotnet nuget locals all --clear` doesn't sound like something you really want to run, on the CI server won't it also meant you'll blow away the cache of framework packages extending the build time?

.github/workflows/build.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Test on .NET Core 3.1
+        run: dotnet test -c Release -f netcoreapp3.1 --no-build --no-restore -l "console;verbosity=detailed"
+      - name: Test on .NET Framework 4.6.1 (Windows only)
+        run: dotnet test -c Release -f net461 --no-build --no-restore -l "console;verbosity=detailed"
+        if: matrix.os == 'windows-latest'
+      - name: Test on .NET Framework 4.6.1 using Mono
+        run: |
+          docker run --rm -v "$PWD":'/project' -w='/project' mono:$MONO_TAG mono ./src/Castle.Core.Tests/bin/Release/net461/Castle.Core.Tests.exe
+          docker run --rm -v "$PWD":'/project' -w='/project' mono:$MONO_TAG mono ./src/Castle.Core.Tests.WeakNamed/bin/Release/net461/Castle.Core.Tests.WeakNamed.exe
+        if: matrix.os == 'ubuntu-latest'
```
jonorossi, line 54: Can you run the 2 of these in the same container which should speed it up.

jonorossi (approved):
I'm happy with the direction. I wouldn't mind if this goes in sooner rather than later, even if we still release from AppVeyor for now. That allows us to add to it over time and transition when ready.

README.md
jonorossi: If you want to merge this soon, I'd just get rid of the AppVeyor status badge since we've AppVeyor sends commit status details so this is unneeded. Just keep the preview feed link until we look into GitHub Packages.

.github/workflows/build.yml
stakx: This is indeed required. Without this `dotnet clean`, the NuGet package restore can sometimes fail. This problem has been documented elsewhere, and &ndash; at least at the time when I prepared this &ndash; the `dotnet clean` appears to be the best-known solution. I can probably locate the issue again that mentions all this.

README.md
stakx: Done.

I agree with you that merging this sooner rather, then successively iterating on the GitHub Actions workflow(s), is probably better than trying to get everything perfect right from the start.

.github/workflows/build.yml
stakx: No particular reason, I think I simply copied that from a template without thinking too much about it. I've simplified it to just `on: [ push, pull_requests ]` without mentioning any specific branch names.

stakx:
@jonorossi, I've been able to improve on the issues you raised above. Still far from complete, but I think it's reasonably clean now.

I'm definitely skipping a few things (for instance: versioning, and publishing packages) at this time. It appears that Appveyor is currently both faster and more reliable overall than GitHub Actions. I've witnessed several random workflow failures e.g. because of intermittent network connection break-offs.

It would perhaps be wise to gain a little bit more everyday experience with GitHub Actions in this project before going all in and investing a lot more time. (That being said, I still think moving to GitHub Actions is a step in the right direction.)

jonorossi (approved):
Unfortunate to hear about broken network connections, is that connecting to nuget.org I've not had any random failures on a Windows instance of GitHub Actions I'm using for a private project, however I'm subscribed to GitHub's status page and they've definitely had a lot of outages in the last 1-2 years compared to previously, I think the rapid scale out with growth is a challenge at the moment.

Definitely agree that it would be good get to some more real world experience. Merge if you are ready, this looks great.

PR #563, opened 2021-01-19
Remove `Castle.Core.Internal.CollectionExtensions`

### LINKED ISSUE #560: Please make Castle.Core.Internal.CollectionExtensions internal because it shows up in IntelliSense
Both Visual Studio natively and ReSharper suggest extension methods they found in unimported namespaces from all directly or indirectly referenced packages. There are parentheses showing that they will add a `using` directive for that namespace if you select that extension method, but if you are moving quickly, you can end up with `using Castle.Core.Internal;` without noticing. Sometimes the first time you notice is many commits later or after submitting a PR. It's not pleasant to be stepping around these gotcha entries in IntelliSense.

stakx:
Closes #560.

Its five methods are dealt with as follows:

 * `Find`, `IsNullOrEmpty`: unused.

 * `FindAll`: used only once, inlined.

 * `GetContentsHashCode`, `AreEquivalent`: used to implement equality contract in `ProxyGenerationOptions`. Moved there, mostly unchanged.


src/Castle.Core/DynamicProxy/ProxyGenerationOptions.cs
```diff
@@ -296,8 +293,61 @@ public override int GetHashCode()
 			result = 29*result + (Selector != null ? 1 : 0);
 			result = 29*result + MixinData.GetHashCode();
 			result = 29*result + (BaseTypeForInterfaceProxy != null ? BaseTypeForInterfaceProxy.GetHashCode() : 0);
-			result = 29*result + CollectionExtensions.GetContentsHashCode(AdditionalAttributes);
+			result = 29*result + GetAdditionalAttributesHashCode();
```
stakx, line 296: `GetHashCode` does not have to be precise (only fast), so we could just use `AdditionalAttributes.Count` instead of summing the attributes' hash codes. I've not done so because there are some unit tests affected by it; someone in the past must have thought  it important that `options.GetHashCode` differ for different attributes.

jonorossi (approved):
That was easy 😄

PR #566, opened 2021-01-25
Simplify SimpleAST by converting `Expression`, `Statement` to interfaces

stakx:
`ExpressionStatement` is an adapter type that turns an `Expression` into a `Statement`. Its mere existence demonstrates that a strict distinction between expressions and statements is more of a stumbling block than it is helpful. Some AST node types (such as `MethodInvocationExpression`) need to act as either.

We can model this at the type system level by converting `Expression` and `Statement` to interfaces. AST node types can then implement both interfaces, which renders adapter types `ExpressionStatement` and `ReferenceExpression` redundant. Removing those from AST trees means fewer allocations and thus slightly less GC pressure.

### Outlook

I'd like to follow this up with another PR that will introduce an interface `ILocation` for modelling storage locations. The `Load` / `Store` methods of `Reference` would end up there. I expect this would both simplify by-ref handling and indirect loads / stores.


src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/BlockStatement.cs
```diff
@@ -17,23 +17,18 @@ namespace Castle.DynamicProxy.Generators.Emitters.SimpleAST
 	using System.Collections.Generic;
 	using System.Reflection.Emit;
 
-	internal class MultiStatementExpression : Expression
+	internal class BlockStatement : IStatement
```
jonorossi, line 20: `MultiStatementExpression`, haha, yer that shows things were not quite right 👍

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/Reference.cs
```diff
@@ -45,14 +45,9 @@ public virtual void Generate(ILGenerator gen)
 		{
 		}
 
-		public virtual Expression ToAddressOfExpression()
+		void IExpressionOrStatement.Emit(IMemberEmitter member, ILGenerator gen)
```
jonorossi, line 48: Why does this need to be explicitly implemented?

jonorossi (approved):
Looks great. That definitely tidies up the code, especially that the stack based runtime rarely cares about the difference between C# level statements and expressions.

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/Reference.cs
stakx: Because of... uh, important reasons that I can't recall. 😁 Fixed now.

PR #570, opened 2021-01-30
Emit delegate types in same module as their associated proxy type

### LINKED ISSUE #569: DynamicProxy generates two modules when proceeding from a class proxy's protected method to the target
This is very similar to #327. (Back then we found out that DynamicProxy sometimes puts a generated invocation type in another module than the generated proxy type depending on it, causing verification failure due to the proxy type module being incomplete. This got fixed in #346.)

It appears there is another case where the same problem occurs, namely proceeding to a class proxy's target from an intercepted protected method. Because you cannot directly call protected methods from outside the target, DynamicProxy does so indirectly via a generated delegate type. Such delegate types can also end up in the wrong module:

```csharp
public class Foo
{
    public string InvokeMethod()
    {
        return Method();
    }

    protected virtual string Method()
    {
        return "...";
    }
}

var builder = new PersistentProxyBuilder();
var generator = new ProxyGenerator(builder);
_ = generator.CreateClassProxyWithTarget(new Foo());
builder.SaveAssembly();
```

The last line triggers this exception:

> ```
> System.InvalidOperationException: Both a strong-named and a weak-named assembly have been generated.
> ```

stakx:
Fixes #569. See commit messages for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Bugfixes:
 - Proxying certain `[Serializable]` classes produces proxy types that fail PEVerify test (@stakx, #367)
 - `private protected` methods are not intercepted (@CrispyDrone, #535)
 - `System.UIntPtr` unsupported (@stakx, #546)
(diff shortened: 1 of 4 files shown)


stakx:
Nothing spectacular to be seen here, this simply applies a previous bug fix to another location that was overlooked back then.

PR #571, opened 2021-01-30
Forward non-intercepted methods on class proxies to target

### LINKED ISSUE #67: Not proxied methods on class proxy with target should forward to the target [DYNPROXY-183]

### LINKED ISSUE #536: Virtual property is not proxied in DynamicProxy if ProxyGenerationHook returns false for ShouldInterceptMethod
**Expected:** the property to be proxied but not intercepted.

**Actual:** the property is generated in the proxy but does not pass the value to the underlying instance

```
public class TestCacheProxyGenerationHook : AllMethodsHook
        {
            public override bool ShouldInterceptMethod(Type type, MethodInfo methodInfo)
            {
                return false;
            }
        }

public interface ITestCacheInterface
        {
            int InstanceProperty { get; set; }
        }

        public class TestClassForCache : ITestCacheInterface
        {
            public virtual int InstanceProperty { get; set; }
        }

        [Fact]
        public async Task DynamicProxy_NonIntercepted_Property_Leaked()
        {
            var instance = new TestClassForCache();
            var toProxy = instance.GetType();

            var proxyGenerationOptions = new ProxyGenerationOptions(new TestCacheProxyGenerationHook());

            var proxy = generator.CreateClassProxyWithTarget(toProxy,
                instance,
                proxyGenerationOptions);

            var accessor = (ITestCacheInterface)proxy;
            accessor.InstanceProperty = 1;

            Assert.Equal(accessor.InstanceProperty, instance.InstanceProperty); // Fails here 1 != 0
        }

```

stakx:
This is a redo of my previous PR #450, which got stalled because I was under the impression that forwarding to non-public methods on the target would require special handling (and thus a lot of additional work). I've since discovered that the necessary bits are already in place:

When one proceeds from an intercepted non-public method of a class proxy to the target, the target method will likely be non-public, too, and therefore not directly callable from the outside. DynamicProxy solves this by calling the non-public target method through a delegate. This present PR simply piggybacks on that mechanism.

I found an issue with delegate type generation while working on this; see #570, which is a prerequisite for this PR and needs to be merged first. <del>(#570's commits are included here to avoid unrelated test failures; I'll rebase them away once it has been merged. For now, please ignore the first three commits of this PR.)</del>  <ins>**Update:** I've merged #570 and updated this PR.</ins>

Fixes #67, fixes #536.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 Enhancements:
 - .NET Standard 2.0 and 2.1 support (@lg2de, #485)
+- Non-intercepted methods on a class proxy with target are now forwarded to the target (@stakx, #571)
 
 Bugfixes:
 - Proxying certain `[Serializable]` classes produces proxy types that fail PEVerify test (@stakx, #367)
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
@@ -56,14 +56,23 @@ namespace Castle.DynamicProxy.Contributors
 				return null;
 			}
 
-			if (IsDirectlyAccessible(method) == false)
-			{
-				return IndirectlyCalledMethodGenerator(method, @class, overrideMethod);
-			}
+			var methodIsDirectlyAccessible = IsDirectlyAccessible(method);
 
 			if (!method.Proxyable)
 			{
-				return new ForwardingMethodGenerator(method, overrideMethod, (c, m) => c.GetField("__target"));
+				if (methodIsDirectlyAccessible)
+				{
+					return new ForwardingMethodGenerator(method, overrideMethod, (c, m) => c.GetField("__target"));
+				}
+				else
+				{
+					return IndirectlyCalledMethodGenerator(method, @class, overrideMethod, skipInterceptors: true);
+				}
+			}
+
+			if (!methodIsDirectlyAccessible)
+			{
+				return IndirectlyCalledMethodGenerator(method, @class, overrideMethod);
 			}
 
 			var invocation = GetInvocationType(method, @class);
@@ -136,7 +145,8 @@ namespace Castle.DynamicProxy.Contributors
 		}
 
 		private MethodGenerator IndirectlyCalledMethodGenerator(MetaMethod method, ClassEmitter proxy,
-		                                                        OverrideMethodDelegate overrideMethod)
+		                                                        OverrideMethodDelegate overrideMethod,
+		                                                        bool skipInterceptors = false)
 		{
 			var @delegate = GetDelegateType(method, proxy);
 			var contributor = GetContributor(@delegate, method);
@@ -144,7 +154,7 @@ namespace Castle.DynamicProxy.Contributors
 				.Generate(proxy, namingScope)
 				.BuildType();
 			return new MethodWithInvocationGenerator(method,
-			                                         proxy.GetField("__interceptors"),
+			                                         skipInterceptors ? NullExpression.Instance : proxy.GetField("__interceptors"),
 			                                         invocation,
 			                                         (c, m) => c.GetField("__target"),
 			                                         overrideMethod,
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
@@ -56,15 +56,14 @@ namespace Castle.DynamicProxy.Contributors
 				return null;
 			}
 
-			if (!method.Proxyable)
-			{
-				return new MinimialisticMethodGenerator(method,
-				                                        overrideMethod);
-			}
-
 			if (IsDirectlyAccessible(method) == false)
 			{
 				return IndirectlyCalledMethodGenerator(method, @class, overrideMethod);
+			}
+
+			if (!method.Proxyable)
+			{
+				return new ForwardingMethodGenerator(method, overrideMethod, (c, m) => c.GetField("__target"));
 			}
 
 			var invocation = GetInvocationType(method, @class);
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -4,11 +4,15 @@
 
 Enhancements:
 - .NET Standard 2.0 and 2.1 support (@lg2de, #485)
+- Non-intercepted methods on a class proxy with target are now forwarded to the target (@stakx, #571)
 
 Bugfixes:
+- Not proxied methods on class proxy with target should forward to the target (DYNPROXY-183) (@jonorossi, #67)
 - Proxying certain `[Serializable]` classes produces proxy types that fail PEVerify test (@stakx, #367)
 - `private protected` methods are not intercepted (@CrispyDrone, #535)
+- Virtual property is not proxied if `IProxyGenerationHook.ShouldInterceptMethod` returns false (@xumix, #536)
```
stakx, line 13: We probably don't need three separate entries, but I wasn't sure about the changelog entry format for this situation.

src/Castle.Core/DynamicProxy/Contributors/MembersCollector.cs
```diff
@@ -194,6 +194,18 @@ private MetaMethod AddMethod(MethodInfo method, IProxyGenerationHook hook, bool
...
+		}
+
+		/// <summary>
+		///   Performs some basic screening to filter out non-interceptable methods.
+		/// </summary>
+		/// <remarks>
+		///   The <paramref name="hook"/> will get invoked for non-interceptable method notification only;
+		///   it does not get asked whether or not to intercept the <paramref name="method"/>.
+		/// </remarks>
+		protected bool AcceptMethodPreScreen(MethodInfo method, bool onlyVirtuals, IProxyGenerationHook hook)
```
stakx, line 208: I am splitting off the pre-screening from `AcceptMethod` so that `WrappedMembersCollector` has a way of distinguishing between non-interceptable methods (e.g. `object.Finalize`, `object.MemberwiseClone`), and methods that the hook wants filtered out. It is only the latter group of methods that should be forwarded to the target; the former group of methods must not be proxied at all!

(In my original PR, I made this distinction via `AllMethodsHook.SkippedTypes`, which wasn't ideal; see https://github.com/castleproject/Core/pull/450#discussion_r303399484.)

src/Castle.Core/DynamicProxy/Contributors/ClassProxyWithTargetTargetContributor.cs
```diff
@@ -141,17 +149,19 @@ private Type GetInvocationType(MetaMethod method, ClassEmitter @class)
...
+		                                                        bool skipInterceptors = false)
 		{
 			var @delegate = GetDelegateType(method, proxy);
 			var contributor = GetContributor(@delegate, method);
 			var invocation = new CompositionInvocationTypeGenerator(targetType, method, null, false, contributor)
 				.Generate(proxy, namingScope)
 				.BuildType();
 			return new MethodWithInvocationGenerator(method,
-			                                         proxy.GetField("__interceptors"),
+			                                         skipInterceptors ? NullExpression.Instance : proxy.GetField("__interceptors").ToExpression(),
```
stakx, line 161: TBH this may be a bit of a hack: It relies on `AbstractInvocation.Proceed` calling `InvokeMethodOnTarget` right away when `interceptors == null` ([see here](https://github.com/castleproject/Core/blob/d1310e06b52dba1f7d5a35fd4d47482878460fd0/src/Castle.Core/DynamicProxy/AbstractInvocation.cs#L102-L107)), so we get the needed behavior without having to implement a new type of `MethodGenerator`. But the generated proxied method body in this case perhaps does too muchwork, and could be optimized further.

At this time, I don't yet understand the `MethodGenerator` part of DP well enough to add a new variant. I may pick this up in the future, when the code base has been cleaned up a little more and when I'm more familiar with it. I'm hoping this will do in the meantime.

src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
```diff
@@ -34,20 +34,28 @@ internal class MethodWithInvocationGenerator : MethodGenerator
...
 		                                     GetTargetExpressionDelegate getTargetTypeExpression,
 		                                     OverrideMethodDelegate createMethod, IInvocationCreationContributor contributor)
+			: this(method, interceptors.ToExpression(), invocation, getTargetExpression, getTargetTypeExpression, createMethod, contributor)
+		{
+		}
+
+		public MethodWithInvocationGenerator(MetaMethod method, Expression interceptors, Type invocation,
+		                                     GetTargetExpressionDelegate getTargetExpression,
+		                                     GetTargetExpressionDelegate getTargetTypeExpression,
+		                                     OverrideMethodDelegate createMethod, IInvocationCreationContributor contributor)
```
stakx, line 58: This new overload is identical to the preceding one except that `interceptors` is an `Expression`, not a `Reference`. Internally, this class needs an `Expression` anyway, and this ctor allows us to pass in a `NullExpression` (see review comment above).

stakx (commented):
A few explanations on some of the trickier bits.

CHANGELOG.md
jonorossi: Yer, definitely don't need all 3. I'd say this is an enhancement not a defect at this stage, and just keep your issue number, my name and the last one shouldn't be mentioned, neither of us contributed the change.

jonorossi (approved):

CHANGELOG.md
stakx: OK, will do. I'd still want to have a trace back to the original issues, but I suppose it might be enough if they're in the PR.

PR #573, opened 2021-02-02
Speed up interface proxy w/o target type generation significantly

stakx:
DynamicProxy's runtime performance is directly tied to the amount of System.Reflection.Emit-ting done. If we do less dynamic type generation, DynamicProxy becomes faster.

I've recently realised that the main reason why DynamicProxy generates dedicated invocation types for every proxied method is because the invocation's `InvokeMethodOnTarget` implementation depends on the proxied method's signature.

However, if an invocation cannot proceed to a (method on a) target, we don't need a custom implementation for `InvokeMethodOnTarget`, and thus no custom invocation type; we can use a pre-defined invocation type instead!

This PR is a realization of this idea. I'm starting with ~~non-generic~~ methods on interface proxies without target. It should be possible to cover more methods &mdash; ~~generic ones (I think), and~~ abstract methods on class proxies without target &mdash; but those will take more work.

I expect this to be very beneficial to some core DynamicProxy usage scenarios, i.e. mocking libraries used in unit testing, where interface mocks are frequently backed by interfaces proxies without target.

I've tested this PR against Moq 4's test suite. Test execution time has dropped by a factor of 2.5x for both the .NET Framework and .NET Core. Another simple `Stopwatch` benchmark suggests roughly the same speed-up.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
@@ -20,6 +20,7 @@ namespace Castle.DynamicProxy.Contributors
 
 	using Castle.DynamicProxy.Generators;
 	using Castle.DynamicProxy.Generators.Emitters;
+	using Castle.DynamicProxy.Internal;
 
 	internal class InterfaceProxyWithoutTargetContributor : CompositeTypeContributor
 	{
@@ -60,6 +61,15 @@ namespace Castle.DynamicProxy.Contributors
 
 		private Type GetInvocationType(MetaMethod method, ClassEmitter emitter)
 		{
+			var methodInfo = method.Method;
+
+			if (canChangeTarget == false && methodInfo.IsAbstract && methodInfo.IsGenericMethod == false && methodInfo.IsGenericMethodDefinition == false)
+			{
+				// We do not need to generate a custom invocation type because no custom implementation
+				// for `InvokeMethodOnTarget` will be needed (proceeding to target isn't possible here):
+				return typeof(InterfaceMethodWithoutTargetInvocation);
+			}
+
 			var scope = emitter.ModuleScope;
 			Type[] invocationInterfaces;
 			if (canChangeTarget)
@@ -70,14 +80,14 @@ namespace Castle.DynamicProxy.Contributors
 			{
 				invocationInterfaces = new[] { typeof(IInvocation) };
 			}
-			var key = new CacheKey(method.Method, CompositionInvocationTypeGenerator.BaseType, invocationInterfaces, null);
+			var key = new CacheKey(methodInfo, CompositionInvocationTypeGenerator.BaseType, invocationInterfaces, null);
 
 			// no locking required as we're already within a lock
 
 			return scope.TypeCache.GetOrAddWithoutTakingLock(key, _ =>
-				new CompositionInvocationTypeGenerator(method.Method.DeclaringType,
+				new CompositionInvocationTypeGenerator(methodInfo.DeclaringType,
 				                                       method,
-				                                       method.Method,
+				                                       methodInfo,
 				                                       canChangeTarget,
 				                                       null)
 				.Generate(emitter, namingScope)
--- a/src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
@@ -63,7 +63,7 @@ namespace Castle.DynamicProxy.Contributors
 		{
 			var methodInfo = method.Method;
 
-			if (canChangeTarget == false && methodInfo.IsAbstract && methodInfo.IsGenericMethod == false && methodInfo.IsGenericMethodDefinition == false)
+			if (canChangeTarget == false && methodInfo.IsAbstract)
 			{
 				// We do not need to generate a custom invocation type because no custom implementation
 				// for `InvokeMethodOnTarget` will be needed (proceeding to target isn't possible here):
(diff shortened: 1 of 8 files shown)


stakx:
@jonorossi, I realise you probably don't have time to review all my recent PRs... there have been quite a few lately. But this very short one may be worth a look.

(A remark about my other refactoring-only PRs: I'll aim at leaving each one those open for at least a couple of days, in case that you find some time to review, or that I may have second thoughts about some aspects. If neither of those two things occurs, I hope I'll be forgiven for going ahead and merging them, so that I can continue with subsequent code clean-ups, I find they often depend upon one another.)

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
```diff
@@ -62,6 +63,15 @@ protected override IEnumerable<MembersCollector> CollectElementsToProxyInternal(
...
 		private Type GetInvocationType(MetaMethod method, ClassEmitter emitter)
 		{
+			var methodInfo = method.Method;
+
+			if (canChangeTarget == false && methodInfo.IsAbstract && methodInfo.IsGenericMethod == false && methodInfo.IsGenericMethodDefinition == false)
+			{
+				// We do not need to generate a custom invocation type because no custom implementation
+				// for `InvokeMethodOnTarget` will be needed (proceeding to target isn't possible here):
+				return typeof(InterfaceMethodWithoutTargetInvocation);
+			}
```
stakx, line 73: Generics only matter because there are some asserts somewhere else that trigger for generic methods; I'll need to look at those more closely before allowing generic methods here.

`IsAbstract` may seem unnecessary now, but it becomes important once we support default interface implementations.

jonorossi: Do generic methods not already work? Looks like `SetGenericMethodArguments` would be called on this invocation instance if you removed the check.

jonorossi (approved):
Top work Dom 👍 . Definitely want to see this one in the changelog.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
```diff
@@ -62,6 +63,15 @@ protected override IEnumerable<MembersCollector> CollectElementsToProxyInternal(
 
 		private Type GetInvocationType(MetaMethod method, ClassEmitter emitter)
 		{
+			var methodInfo = method.Method;
+
+			if (canChangeTarget == false && methodInfo.IsAbstract && methodInfo.IsGenericMethod == false && methodInfo.IsGenericMethodDefinition == false)
+			{
+				// We do not need to generate a custom invocation type because no custom implementation
+				// for `InvokeMethodOnTarget` will be needed (proceeding to target isn't possible here):
+				return typeof(InterfaceMethodWithoutTargetInvocation);
```
jonorossi, line 72: I think it would be good to have at least one unit test that verifies this, since this return type is now part of the public API and to verify the performance improvement is working into the future.

I can also see more performance optimisations on the horizon doing better invocation type reuse, so verifying that will also be important.

stakx: I'll see what I can come up with.

(Speaking briefly about future optimizations: I believe it would be entirely feasible to completely stop generating invocation types, in favor of generating a sibling `InvokeOnTarget` method for each proxied method... directly in the proxy type containing the proxied method. We could then have just one single sealed `Invocation` type that receives a delegate to the appropriate `InvokeOnTarget` method. While that would imply a need to allocate delegates, I suspect this would still be a net win&ndash;in the time that it takes to emit a dynamic type, you can probably do a thousand delegate calls.)

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
stakx: I think they'd work just like that. It's just that existing code (`Assert`) assumes that either a `CompositionInvocation`- or `InheritanceInvocation`-derived type will come along the way, and AFAIK those carry the proxied method's type parameters at the type level. I wouldn't want to just remove those (probably superfluous) asserts without checking if perhaps there _is_ some case I haven't considered for the new invocation type.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
stakx: I have added a single unit test (for now...more to come if we introduce more invocation type reusing). I could have added more tests, but thought it would make most sense to only write tests for cases where specific invocation types get chosen (there's currently only one such case), rather than also testing when they do _not_ get chosen. The latter could easily be misunderstood to be a specification of desired behavior, when in actual fact it would only document the status quo.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
stakx: I added the above change in a slightly modified form, so generic methods are not covered, too.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
jonorossi: > thought it would make most sense to only write tests for cases where specific invocation types get chosen, rather than also testing when they do _not_ get chosen

Completely agree.

> I also refrained from writing any timing-related tests as those can easily become fragile.

Also agree. If we want to do timing tests we should have a separate benchmarks project and use BenchmarkDotNet or something, they can't be run on cloud hosted build servers anyway.

PR #574, opened 2021-02-06
Upgrade log4net to v2.0.12

stakx:


stakx:
The .NET Core test run on Ubuntu is now failing; it appears that this is due to an [issue in log4net 2.0.10's `MemoryAppender`](https://issues.apache.org/jira/projects/LOG4NET/issues/LOG4NET-649?filter=allopenissues). While not documented, it appears to have been resolved in 2.0.12.

We have the following options:

1. don't update log4net at all
2. update to 2.0.10 and disable the failing tests
3. update to 2.0.12

(2), i.e. picking a buggy version and skipping tests because of that, seems like the worst option to me.

jonorossi:
@stakx I'd go with option 3, upgrade to 2.0.12. 2.0.10 was released 5 months ago with .11 a month later and .12 another month later, close enough that it is unlikely to be a problem. NuGet will pretty much treat it as a patch version anyway, 2.0.12 nearly has 3 times the number of downloads, and we are making a major release next too.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -19,6 +19,16 @@
 		<StartupObject>Program</StartupObject>
 	</PropertyGroup>
 
+	<ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
+		<None Include="config\netfx\App.config" />
+		<None Remove="config\netcoreapp\App.config" />
+	</ItemGroup>
+
+	<ItemGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
```
jonorossi, line 27: Should we handle .NET 5 now so this works when we upgrade the test project in the future.

Looks like you can use `GetTargetFrameworkIdentifier` which sounds like it'll always return `.NETCoreApp` on .NET Core and .NET 5:
https://docs.microsoft.com/en-us/dotnet/core/compatibility/msbuild/5.0/targetframework-name-change

stakx: I didn't know about that yet! Done.

PR #579, opened 2021-02-21
Simplify SimpleAST further

stakx:
* Remove `BindDelegateExpression` and `LoadArrayElementExpression` (both unused), as well as `NopStatement` (which serves no useful purpose).
* Replace `ConstReference` with `LiteralIntExpression` (preexisting), as well as `LiteralBoolExpression` and `LiteralStringExpression` (both new).
* Convert static helper method `ArgumentsUtil.EmitLoadOwnerAndReference` to instance method `Reference.Emit`.
* Stop passing `IMemberEmitter member` to every AST node (only `Return` ever used it, but doesn't actually need it), then remove `IMemberEmitter` interface altogether.
* Remove some duplicate / redundant code paths.


src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/ReturnStatement.cs
```diff
@@ -37,10 +38,7 @@ public void Emit(IMemberEmitter member, ILGenerator gen)
 			}
 			else
 			{
-				if (member.ReturnType != typeof(void))
-				{
-					OpCodeUtil.EmitLoadOpCodeForDefaultValueOfType(gen, member.ReturnType);
```
stakx, line 52: Note that this is the only location where the `IMemberEmitter member` parameter gets used (vs. just being passed on to other `Emit` calls).

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/Reference.cs
```diff
@@ -47,12 +47,7 @@ public virtual void Generate(ILGenerator gen)
...
 		{
-			this.EmitLoadOwnerAndReference(gen);
-		}
-
-		public void EmitLoadOwnerAndReference(ILGenerator gen)
-		{
-			OwnerReference?.EmitLoadOwnerAndReference(gen);
+			OwnerReference?.Emit(member, gen);
 			LoadReference(gen);
 		}
```
stakx: I'm not entirely sure that nothing important is lost (in terms of semantics) when the (outward) distinction between `Emit` and `EmitLoadOwnerAndReference` is erased.

(It's true that `reference.Emit` always did the same as `EmitLoadOwnerAndReference`, but because it appears to never have been called directly, call sites read differently up until now.)

stakx:
Based on the observation that `ConstReference` is not a reference, just having discovered that there are more `Reference` subclasses that probably also shouldn't be references, and https://github.com/castleproject/Core/pull/579#discussion_r580599684, I am going to close this and redo the same work as a bunch of smaller, more focused PRs.

Issue #602, opened 2021-11-29
CastleCore using an old version of System.Net.Http which is vulnerable to "DoS", "Spoofing", "Privilege Escalation", "Authentication Bypass" and "Information Exposure"


sydseter:
The following vulnerable libraries were found:  System.Net.Http@4.3.0 and System.Text.RegularExpressions@4.3.0

All issues for System.Net.Http@4.3.0 have been fixed in 4.3.4.
All issues for System.Text.RegularExpressions@4.3.0 have been fixed in 4.3.1

These are the vulnerabilities associated vulnerable paths:

✗ Denial of Service (DoS) [High Severity]https://snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60045 in System.Net.Http@4.3.0
introduced by:
Castle.Core@4.4.0 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0
This issue was fixed in versions: 4.1.2, 4.3.2
✗ Improper Certificate Validation [High Severity]https://snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60046 in System.Net.Http@4.3.0
introduced by:
Castle.Core@4.4.0 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0
This issue was fixed in versions: 4.1.2, 4.3.2
✗ Privilege Escalation [High Severity]https://snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60047 in System.Net.Http@4.3.0
introduced by:
Castle.Core@4.4.0 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0
This issue was fixed in versions: 4.1.2, 4.3.2
✗ Authentication Bypass [Medium Severity]https://snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-60048 in System.Net.Http@4.3.0
introduced by:
Castle.Core@4.4.0 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0
This issue was fixed in versions: 4.1.2, 4.3.2
✗ Information Exposure [High Severity]https://snyk.io/vuln/SNYK-DOTNET-SYSTEMNETHTTP-72439 in System.Net.Http@4.3.0
introduced by:
Castle.Core@4.4.0 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0
[...]

PR #604, opened 2021-12-13
Update vulnerable libraries

eskildsen:
This is meant as an update for the current 4.4.1 version. Suggested version is 4.4.2 as it is only minor changes.

Aim is to update the dependencies System.Net.Http and System.Text.RegularExpressions to versions with security fixes. It is resolved in the new version 5 on master, however, with no suggested release date, this acts as an intermediary fix.

Feel free to comment, if you see this as a wrong approach, or if code needs to be altered.

### DIFF
--- a/src/Castle.Core/Castle.Core.csproj
+++ b/src/Castle.Core/Castle.Core.csproj
@@ -43,6 +43,11 @@
 		<PackageReference Include="System.ComponentModel" Version="4.3.0" />
 		<PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
 	</ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)'!='net35'AND'$(TargetFramework)'!='net40'">
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
+  </ItemGroup>
 
 	<ItemGroup>
 		<Folder Include="Properties\" />
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## 4.4.2 (2021-12-13)
+
+Bugfixes:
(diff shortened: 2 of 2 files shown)


eskildsen:
This is meant to be branch from https://github.com/castleproject/Core/tree/v4.4.1, so should probably not go to master, but rather a new branch

src/Castle.Core/Castle.Core.csproj
```diff
@@ -43,6 +43,11 @@
 		<PackageReference Include="System.ComponentModel" Version="4.3.0" />
 		<PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
 	</ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)'!='net35'AND'$(TargetFramework)'!='net40'">
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
```
Havunen, line 48: These packages should probably be defined with open upper version so application layer can have higher version installed

Havunen:
Wouldn't it be way better to add netstandard2.0 / net5.0 / net6.0 compilation targets and not have any dependencies for those?

Havunen:
I created an issue: https://github.com/castleproject/Core/issues/606 but I dont know is anybody maintaining this project?
New nuget system makes it difficult to patch dependencies of dependencies...

eskildsen:
> Wouldn't it be way better to add netstandard2.0 / net5.0 / net6.0 compilation targets and not have any dependencies for those?

Yes, it would certainly be better. However, this is meant as a small non-breaking change for the V4 branch. It seems the work on V5 will resolve this issues, so this is only until then. From recent issues and comments in this repository, this is something getting more attention. And it would be easier to get dependant projects (e.g. AutoMoq) to update to a minor version.

Havunen:
Hmm, yeah but adding new target frameworks should not be breaking change if old target frameworks are ny removed. Change is easy to do, but getting maintainers attention to publish new version is difficult :(

jonorossi:
I'm going to close this until it is clear in #602 why this change should go into Castle.

Havunen:
Can you add modern target frameworks to official castle core package and upload it to nuget please? It would be nice to avoid larger refactoring away from castle.core

Issue #473, opened 2019-09-10
Support Collectible Dynamic Assemblies


jonorossi:
**.NET Core 3.0 introduces support for unloading assemblies after loading them into a custom `AssemblyLoadContext`** via the new `Unload` method and constructor with `isCollectible` parameter. The custom `AssemblyLoadContext` is set up to only contain the assemblies you want with everything else being in the default context. Obviously this is a .NET Core 3.0 feature, and .NET Framework will never support it; .NET Standard also does not define `System.Runtime.Loader` classes. Unfortunately it looks like dynamic assemblies were missed (https://github.com/dotnet/corefx/issues/38426) and support for them won't make the cut for .NET Core 3.0.

- https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability-howto
- https://github.com/dotnet/samples/tree/master/core/tutorials/Unloading
- https://github.com/dotnet/coreclr/issues/552

**.NET Framework 4.0 introduced `AssemblyBuilderAccess.RunAndCollect`** which allows dynamic assemblies to be garbage collected when no reference is held. `RunAndCollect` has a whole heap of restrictions on what can be emitted into the assembly (e.g. no COM interop), but I think we could fairly trivially add opt-in for this per `ModuleScope` to make it disposable.

- https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/collectible-assemblies

I don't know how useful this feature would actually be since we don't get any requests for it, but logging this as an enhancement to raise the topic since it is possible. It sounds like it `RunAndCollect` could be helpful for #472.

stakx:
> I don't know how useful this feature would actually be since we don't get any requests for it [...]

There has been at least [one StackOverflow question](https://stackoverflow.com/q/57934193/240733) which seems related to just this.

> we could fairly trivially add opt-in for this per `ModuleScope` to make it disposable [...]. It sounds like it RunAndCollect could be helpful for #472.

I haven't thought it through all the way, one thing that perhaps would have to be pointed out in the documentation is that dynamic assemblies could be collected only once their associated `ProxyGenerator` is no longer referenced, nor any proxy objects created by that generator.

jonorossi:
> I haven't thought it through all the way, one thing that perhaps would have to be pointed out in the documentation is that dynamic assemblies could be collected only once their associated `ProxyGenerator` is no longer referenced, nor any proxy objects created by that generator.

[...]

PR #610, opened 2022-03-21
Fix generated assembly to be collectible

m08pvv:
Fixed issue #473


jonorossi:
Sorry for the delay in reviewing this.

I expected the unit tests to have failed with the [Restrictions on collectible assemblies](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/collectible-assemblies) saying these assemblies are "transient (that is, it cannot be saved)". I assume that the unit tests are following the previous branch which uses `RunAndSave` and so this has no unit tests.

With the number of restrictions on collectable dynamic assemblies I think this will probably need to be an opt-in selection. I can't quite understand the details but there sounds like a big gotchas where non-collectable assemblies can't reference collectable assemblies.

Could you please share some details about the testing you've done. Thanks.

jonorossi:
Closing as stale.

PR #615, opened 2022-04-07
Update copyright year to 2022

Jevonius:
Small change but touches many files - includes a couple of places where the copyright was still 2019.

Has there been any discussion around whether this is required in every single file, or whether it could just be in the README/LICENSE file in the root?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2004-2021 Castle Project - http://www.castleproject.org/
+Copyright 2004-2022 Castle Project - http://www.castleproject.org/
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Debugging symbols are available in symbol packages in the AppVeyor build artifac
 
 ## License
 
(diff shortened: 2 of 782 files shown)


stakx:
@Jevonius:

> Has there been any discussion around whether this is required in every single file, or whether it could just be in the README/LICENSE file in the root?

See https://github.com/castleproject/Core/pull/540#pullrequestreview-570961267. It appears we didn't reach a final decision on how exactly to deal with years in the copyright notices; but @jonorossi stated that he would have preferred _not_ updating every single file when there's no actual change.

jonorossi:
Ah this chestnut, yer let's not do that again, makes a royal mess of git history 😃 .

I do still have the #540 email in my inbox tagged to come back to it, but legal research/review isn't really that fun. If someone has looked into this before happy to hear what they've come up with.

I'll close this for now without merging and hope to do some research on what we are required to do legally. Legally the current code is fine, it's copyright when it was significantly changed, some people say you only need a starting year anyway.

jonorossi:
Thanks anyway @Jevonius

Jevonius:
No worries 👍  Didn't take any time to do, and agree reviewing it would be a better option.

PR #616, opened 2022-04-08
Add `net6.0` as a target framework

Jevonius:
Adds `net6.0` as a target framework, including handling cases where functions are Windows-only for `net6.0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Debugging symbols are available in symbol packages in the AppVeyor build artifac
 
 ## License
 
-Castle Core is &copy; 2004-2021 Castle Project. It is free software, and may be redistributed under the terms of the [Apache 2.0](http://opensource.org/licenses/Apache-2.0) license.
+Castle Core is &copy; 2004-2022 Castle Project. It is free software, and may be redistributed under the terms of the [Apache 2.0](http://opensource.org/licenses/Apache-2.0) license.
 
 ## Contributing
 
@@ -36,9 +36,9 @@ build.cmd
 
 Compilation requires a C# 9 compiler, an up-to-date .NET Core SDK, and MSBuild 15+ (which should be included in the former).
 
-Running the unit tests additionally requires the .NET Framework 4.6.2+ as well as the .NET Core 2.1 and 3.1 runtimes to be installed. (If you do not have all of those installed, you can run the tests for a specific target framework using `dotnet test -f <framework>`.)
+Running the unit tests additionally requires the .NET Framework 4.6.2+ as well as the .NET Core 2.1, 3.1 and 6.0 runtimes to be installed. (If you do not have all of those installed, you can run the tests for a specific target framework using `dotnet test -f <framework>`.)
 
-These requirements should be covered by Visual Studio 2019 and the .NET 5 SDK.
+These requirements should be covered by Visual Studio 2022 and the .NET 6 SDK.
 
 ### On Linux
 
@@ -48,7 +48,7 @@ These requirements should be covered by Visual Studio 2019 and the .NET 5 SDK.
 
 Compilation requires an up-to-date .NET Core SDK.
 
-Running the unit tests additionally requires the .NET Core 3.1 runtime to be installed, as well as either Docker or Mono. For the latter, we recommend Mono 5.10+, though older versions (4.6.1+) might still work as well.
+Running the unit tests additionally requires the .NET Core 3.1 and 6.0 runtimes to be installed, as well as either Docker or Mono. For the latter, we recommend Mono 5.10+, though older versions (4.6.1+) might still work as well.
 
 :information_source: **Mono runtime support:** Castle Core runs with minor limitations and defects on Mono 4.0.2+ (however 4.6.1+ is highly recommended, or 5.10+ if your code uses new C# 7.x language features such as `in` parameters).
 
--- a/README.md
+++ b/README.md
@@ -66,8 +66,6 @@ Symbol                              | .NET 4.6.2         | .NET Standard 2.x
 `FEATURE_ASSEMBLYBUILDER_SAVE`      | :white_check_mark: | :no_entry_sign:
 `FEATURE_SERIALIZATION`             | :white_check_mark: | :no_entry_sign:
 `FEATURE_SYSTEM_CONFIGURATION`      | :white_check_mark: | :no_entry_sign:
----                                 |                    |
-`DOTNET462`                         | :white_check_mark: | :no_entry_sign:
 
 * `FEATURE_APPDOMAIN` - enables support for features that make use of an AppDomain in the host.
 * `FEATURE_ASSEMBLYBUILDER_SAVE` - enabled support for saving the dynamically generated proxy assembly.
--- a/ref/Castle.Core-net6.0.cs
+++ b/ref/Castle.Core-net6.0.cs
@@ -205,6 +205,7 @@ namespace Castle.Components.DictionaryAdapter
         public System.Collections.Generic.IDictionary<string, Castle.Components.DictionaryAdapter.PropertyDescriptor> Properties { get; }
         public void AddCopyStrategy(Castle.Components.DictionaryAdapter.IDictionaryCopyStrategy copyStrategy) { }
     }
+    [System.Diagnostics.DebuggerDisplay("Type: {Type.FullName,nq}")]
     public class DictionaryAdapterMeta
     {
         public DictionaryAdapterMeta(System.Type type, System.Type implementation, object[] behaviors, Castle.Components.DictionaryAdapter.IDictionaryMetaInitializer[] metaInitializers, Castle.Components.DictionaryAdapter.IDictionaryInitializer[] initializers, System.Collections.Generic.IDictionary<string, Castle.Components.DictionaryAdapter.PropertyDescriptor> properties, Castle.Components.DictionaryAdapter.IDictionaryAdapterFactory factory, System.Func<Castle.Components.DictionaryAdapter.DictionaryAdapterInstance, Castle.Components.DictionaryAdapter.IDictionaryAdapter> creator) { }
@@ -567,6 +568,8 @@ namespace Castle.Components.DictionaryAdapter
     {
         public KeySubstitutionAttribute(string oldValue, string newValue) { }
     }
+    [System.Diagnostics.DebuggerDisplay("Count = {Count}, Adapter = {Adapter}")]
+    [System.Diagnostics.DebuggerTypeProxy(typeof(Castle.Components.DictionaryAdapter.ListProjectionDebugView<T>))]
     public class ListProjection<T> : Castle.Components.DictionaryAdapter.IBindingListSource, Castle.Components.DictionaryAdapter.IBindingList<T>, Castle.Components.DictionaryAdapter.ICollectionAdapterObserver<T>, Castle.Components.DictionaryAdapter.ICollectionProjection, System.Collections.Generic.ICollection<T>, System.Collections.Generic.IEnumerable<T>, System.Collections.Generic.IList<T>, System.Collections.ICollection, System.Collections.IEnumerable, System.Collections.IList, System.ComponentModel.IBindingList, System.ComponentModel.ICancelAddNew, System.ComponentModel.IChangeTracking, System.ComponentModel.IEditableObject, System.ComponentModel.IRaiseItemChangedEvents, System.ComponentModel.IRevertibleChangeTracking
     {
         public ListProjection(Castle.Components.DictionaryAdapter.ICollectionAdapter<T> adapter) { }
@@ -659,6 +662,7 @@ namespace Castle.Components.DictionaryAdapter
         public object NewValue { get; }
         public object OldValue { get; }
     }
+    [System.Diagnostics.DebuggerDisplay("{Property.DeclaringType.FullName,nq}.{PropertyName,nq}")]
     public class PropertyDescriptor : Castle.Components.DictionaryAdapter.IDictionaryBehavior, Castle.Components.DictionaryAdapter.IDictionaryKeyBuilder, Castle.Components.DictionaryAdapter.IDictionaryPropertyGetter, Castle.Components.DictionaryAdapter.IDictionaryPropertySetter
     {
         protected System.Collections.Generic.List<Castle.Components.DictionaryAdapter.IDictionaryBehavior> dictionaryBehaviors;
@@ -1119,7 +1123,9 @@ namespace Castle.Components.DictionaryAdapter.Xml
     }
     public static class Try
     {
+        [System.Diagnostics.DebuggerHidden]
         public static bool Failure<T>(out T result) { }
+        [System.Diagnostics.DebuggerHidden]
         public static bool Success<T>(out T result, T value) { }
     }
     public static class TypeExtensions
@@ -1978,6 +1984,7 @@ namespace Castle.Core.Logging
         public override Castle.Core.Logging.ILogger CreateChildLogger(string loggerName) { }
         protected override void Log(Castle.Core.Logging.LoggerLevel loggerLevel, string loggerName, string message, System.Exception exception) { }
     }
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
     public class DiagnosticsLogger : Castle.Core.Logging.LevelFilteredLogger, System.IDisposable
     {
         public DiagnosticsLogger(string logName) { }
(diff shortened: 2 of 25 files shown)


Jevonius:
Having tweaked how `MONO_EVENTLOG_TYPE` is being set in `DiagnosticsLoggerTestCase`, I briefly considered adding logic in a static constructor for `DiagnosticsLogger` to automatically set `MONO_EVENTLOG_TYPE` when using Mono. Realised that probably warrants some discussion though, given it would be a change to default behaviour and could be undesirable.
Code would have been along the lines of:
``` C#
static DiagnosticsLogger()
{
	if (RunningOnMono() && string.IsNullOrEmpty(Environment.GetEnvironmentVariable("MONO_EVENTLOG_TYPE")))
	{
		Environment.SetEnvironmentVariable("MONO_EVENTLOG_TYPE", "local:" + Environment.CurrentDirectory);
	}
}

private static bool RunningOnMono()
{
	// taken from http://www.mono-project.com/FAQ:_Technical
	return Type.GetType("Mono.Runtime") != null;
}
```

Jevonius:
Not entirely sure why for commit [e078044](https://github.com/castleproject/Core/pull/616/commits/e0780442c376b1c7d3b9c04029da1f5fc14a8b04) the [GitHub Ubuntu build](https://github.com/castleproject/Core/runs/5892963347?check_suite_focus=true) failed to respect `SupportedOSPlatform`, whereas the [AppVeyor Ubuntu build](https://ci.appveyor.com/project/castleproject/core/builds/43182665) did, but have switched to the NUnit Platform attribute which is more consistent with elsewhere in the codebase anyway.

There are a few places I've had to make code "Windows only" which are probably worthy of review - I've generally limited this to when on `net6.0` but there is a change in `DynamicProxy` around COM interop that it would be good to get a second opinion on. I tried checking the Mono source and limiting to Windows here is _probably_ correct, but `Marshal.GetIUnknownForObject`'s behaviour on Mono is unclear to me (see [IsComObject](https://github.com/mono/mono/blob/main/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs#L840-L848), and [GetIUnknownForObject](https://github.com/mono/mono/blob/main/mcs/class/corlib/System.Runtime.InteropServices/Marshal.cs#L609-L618)).

Jevonius:
Need to update README and CHANGELOG.

src/Castle.Core/Castle.Core.csproj
```diff
@@ -3,7 +3,15 @@
...
 	<PropertyGroup>
-		<TargetFrameworks>net462;netstandard2.0;netstandard2.1</TargetFrameworks>
+		<TargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0</TargetFrameworks>
+
+		<NoWarn>$(NoWarn);SYSLIB0003</NoWarn>
+		<!--
+			Code Access Security is not available in Core, and obsolete in net5.0+
+			This stops warnings about SYSLIB0003 (https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0003)
+			Depending on how these are deprecated in a future dotnet release, may need to wrap relevant code with "#if NET462_OR_GREATER" or similar instead.
+		-->
```
stakx, line 13: (As above, I'm surprised this is necessary, and that we still have CAS code sticking around.)

src/Castle.Core/DynamicProxy/ProxyGenerator.cs
```diff
@@ -558,7 +556,7 @@ public object CreateInterfaceProxyWithTarget(Type interfaceToProxy, object targe
 
 			if (target != null)
 			{
-				if (Marshal.IsComObject(target))
+				if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Marshal.IsComObject(target))
```
stakx, line 559: I am somewhat doubtful that this additional runtime check is truly needed, just because we're adding the `net6.0` TFM.

I understand that improved diagnostics now produce a _compile-time_ warning that wasn't there before, and that's why you're adding this additional _run-time_ check.

However, AFAIK we haven't actually gotten run-time errors in the past due to calling `Marshal.IsComObject` when we should not, so I would tend to simply suppressing that new compiler warning instead of introducing an additional NuGet dependency.

Other opinions?

ref/Castle.Core-net6.0.cs
```diff
@@ -205,6 +205,7 @@ public class DictionaryAdapterInstance
         public System.Collections.Generic.IDictionary<string, Castle.Components.DictionaryAdapter.PropertyDescriptor> Properties { get; }
         public void AddCopyStrategy(Castle.Components.DictionaryAdapter.IDictionaryCopyStrategy copyStrategy) { }
     }
+    [System.Diagnostics.DebuggerDisplay("Type: {Type.FullName,nq}")]
```
stakx, line 208: (Seeing this change while reviewing the PR commit by commit:)

This is strange. Updating stuff in the test project shouldn't change the `ref/` file for the main library. Should this change (and the other ones in the same file) have been included in an earlier commit?

src/Castle.Core.Tests/PublicApiTestCase.cs
```diff
@@ -31,6 +31,7 @@ namespace Castle
 	///   the assertion will obviously fail if someone changes the public API without also including the change to the ref/*.cs files
 	/// </summary>
 	[TestFixture]
+	[System.Runtime.Versioning.SupportedOSPlatform("windows")]
```
stakx: (Looking at this change while reviewing this PR commit by commit:)

~~The commit message claims this will lead to the code only getting executed on Windows. I thought this attribute only affects compilation warnings; in order to actually filter a test by platform, wouldn't we typically use NUnit's `[Platform]` attribute?~~

(Ah, you're going to replace with `[Platform(...)]` in the next commit. I suppose you could squash those two commits to clean things up a little, but I'm fine with it either way.)

src/Castle.Core.Tests/PublicApiTestCase.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright 2004-2021 Castle Project - http://www.castleproject.org/
+﻿// Copyright 2004-2022 Castle Project - http://www.castleproject.org/
```
stakx, line 1: (You could perhaps squash the two "update copyright header" commits into a single one. But I'm fine with it either way.)

stakx (commented):
This looks mostly good to me! :+1: Apart from some comments, I really only have a single request:

Please check whether the Mono runtime guard for `Marshal.IsComObject`, and the related additional NuGet dependency on `net462` is truly necessary. I would personally have gone with simply suppressing the compiler warning instead.

Also, it seems that you're using a different username / email in your commits than on GitHub, is that intentional or did you want to change the commit author to match your GitHub profile?

stakx:
> I briefly considered adding logic in a static constructor for `DiagnosticsLogger` to automatically set `MONO_EVENTLOG_TYPE` when using Mono. Realised that probably warrants some discussion though, given it would be a change to default behaviour and could be undesirable.

I'd keep such a change out of this PR, since it isn't required for adding the `net6.0` target.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -3,7 +3,14 @@
...
 	<PropertyGroup>
-		<TargetFrameworks>net462;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+		<TargetFrameworks>net462;netcoreapp2.1;netcoreapp3.1;net6.0</TargetFrameworks>
+
+		<NoWarn>$(NoWarn);SYSLIB0003</NoWarn>
+		<!--
+			Code Access Security is not available in Core, and obsolete in net5.0+
+			This stops warnings about SYSLIB0003 (https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0003)
+			Depending on how these are deprecated in a future dotnet release, may need to wrap relevant code with "#if NET462_OR_GREATER" or similar at some point.
+		-->
```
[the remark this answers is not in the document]
jonorossi: I agree with @stakx, surprised to see this.

Could this be changed to `#pragma warning disable` around the code rather than here so that we can see exactly what's using it and so more code isn't added.

I guess it'll at least be this unit test:
https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core.Tests/DynamicProxy.Tests/AttributesToAvoidReplicatingTestCase.cs#L77-L80

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -59,11 +65,15 @@
 	<ItemGroup Condition="'$(TargetFramework)'=='netcoreapp2.1'OR'$(TargetFramework)'=='netcoreapp3.1'">
 		<PackageReference Include="System.Security.Permissions" Version="4.7.0" />
 	</ItemGroup>
+	<ItemGroup Condition="'$(TargetFramework)'=='net6.0'">
+		<PackageReference Include="System.Security.Permissions" Version="6.0.0" />
+	</ItemGroup>
```
jonorossi, line 70: Is there a reason we can't just use 6.0.0 on all targets? Looks like it supports our targets.

src/Castle.Core/Castle.Core.csproj
jonorossi: Same as above, change to pragma.

src/Castle.Core/Castle.Core.csproj
```diff
@@ -40,6 +49,10 @@
 		<PackageReference Include="System.Diagnostics.EventLog" Version="4.7.0" />
 	</ItemGroup>
 
+	<ItemGroup Condition="'$(TargetFramework)'=='net6.0'">
+		<PackageReference Include="System.Diagnostics.EventLog" Version="6.0.0" />
+	</ItemGroup>
+
```
jonorossi, line 55: Similar to above, can we not use 6.0.0 for all?

src/Castle.Core/DynamicProxy/ProxyGenerator.cs
jonorossi: `IsComObject` probably works on any platform as I think it just inspects the .NET type.

I too would prefer a pragma to suppress the warning and leave this without the runtime check.

src/Castle.Core/Core/Logging/DiagnosticsLogger.cs
```diff
@@ -23,6 +23,9 @@ namespace Castle.Core.Logging
 	/// </summary>
 #if FEATURE_SERIALIZATION
 	[Serializable]
+#endif
+#if NET6_0_OR_GREATER
+	[System.Runtime.Versioning.SupportedOSPlatform("windows")]
 #endif
```
jonorossi, line 29: Should this be `NET5_0_OR_GREATER` since `SupportedOSPlatformAttribute` was added in .NET 5? Obviously not that important because we don't build a .NET 5 build.

README.md
```diff
@@ -36,9 +36,9 @@ build.cmd
 
 Compilation requires a C# 9 compiler, an up-to-date .NET Core SDK, and MSBuild 15+ (which should be included in the former).
 
-Running the unit tests additionally requires the .NET Framework 4.6.2+ as well as the .NET Core 2.1 and 3.1 runtimes to be installed. (If you do not have all of those installed, you can run the tests for a specific target framework using `dotnet test -f <framework>`.)
+Running the unit tests additionally requires the .NET Framework 4.6.2+ as well as the .NET Core 2.1, 3.1 and 6.0 runtimes to be installed. (If you do not have all of those installed, you can run the tests for a specific target framework using `dotnet test -f <framework>`.)
```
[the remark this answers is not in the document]
jonorossi: IIRC we needed to target netcoreapp2.1+ to use the new NUnit runner, it didn't work on 2.0.

jonorossi (changes requested):

Jevonius:
Apologies for the delay getting back to this - have been swamped. Hoping to get on it again later this week.

README.md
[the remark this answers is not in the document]
Jevonius: I noticed the tests weren't being run against 2.1, which effectively means the `netstandard2.0` target isn't tested. Didn't want to get distracted in this branch though.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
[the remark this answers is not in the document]
Jevonius: There's one instance in `Castle.Core`, and a few in `Castle.Core.Tests`:
https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs#L35

https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core.Tests/DynamicProxy.Tests/AttributesToAvoidReplicatingTestCase.cs#L74

https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core.Tests/DynamicProxy.Tests/AttributesToAvoidReplicatingTestCase.cs#L86

https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core.Tests/DynamicProxy.Tests/AttributesToAvoidReplicatingTestCase.cs#L89

Happy to move the suppression to be `#pragma warning disable SYSLIB0001` if that's preferable - for the test file, would you want it for each instance, or just once at the top of the file?

src/Castle.Core.Tests/PublicApiTestCase.cs
Jevonius: I tend to favour full history, so haven't had much experience with squashing commits - happy to give it a shot if that's the preference here though.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
Jevonius: It does, but unfortunately the 6.0.0 package adds a target file for `netcoreapp2.0` which causes a build error (not just a warning) for EOL targets. See https://nuget.info/packages/System.Diagnostics.EventLog/6.0.0 -> `buildTransitive\netcoreapp2.0\System.Diagnostics.EventLog.targets`

src/Castle.Core/Core/Logging/DiagnosticsLogger.cs
Jevonius: I went with 6 because 5 was never targeted.

ref/Castle.Core-net6.0.cs
Jevonius: I'm struggling to recreate this now (i.e. the version of the file without this addition - I remember being confused by it at the time, but I can't remember the cause. Quite likely it is related to a previous commit, and I didn't rerun all the tests at the time, so found this change at a later point.

Jevonius:
> Please check whether the Mono runtime guard for `Marshal.IsComObject`, and the related additional NuGet dependency on `net462` is truly necessary. I would personally have gone with simply suppressing the compiler warning instead.
I've tweaked the guards that were using the nuget package to work with preprocessor directives instead.

> Also, it seems that you're using a different username / email in your commits than on GitHub, is that intentional or did you want to change the commit author to match your GitHub profile?
This wasn't intentional, just an oversight, but I've added that email as a secondary on my account so it should all link up properly now - thanks for the prompt.

I think I've covered what's been raised, except for squashing commits - not sure if that was just a preference for future reference, or a request for this PR?

jonorossi:
> Apologies for the delay getting back to this - have been swamped. Hoping to get on it again later this week.

Cheers. Sorry for not getting to reviewing this again until now, last week was crazy.

> I've tweaked the guards that were using the nuget package to work with preprocessor directives instead. Keen to keep the Windows checks in place for net6+ so that cross-platform consumers don't get unexpected errors.

Sounds good.

> I think I've covered what's been raised, except for squashing commits - not sure if that was just a preference for future reference, or a request for this PR?

I'll have a browse through it again now. Just a preference, @stakx's commits are a clean masterpiece hiding all of his mess of back and forth 😆 .

src/Castle.Core.Tests/Castle.Core.Tests.csproj
jonorossi: Cheers. All good.

ref/Castle.Core-net6.0.cs
jonorossi: The class has the attribute, so resolving this comment:

https://github.com/castleproject/Core/blob/38298f3ce58902a354412ad6be136c367f12a95e/src/Castle.Core/Components.DictionaryAdapter/DictionaryAdapterMeta.cs#L23-L24

README.md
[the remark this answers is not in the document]
jonorossi: Resolving this comment for now. Let's keep this pull request for .NET 6 moving.

jonorossi (approved):

Issue #447, opened 2019-04-21
Default interface methods


stakx:
The upcoming .NET Core 3.0 along with C# 8 will support default interface methods.

It is likely that DynamicProxy won't support those properly (e.g., `Proceed`?) without some changes. This is a reminder to look into this.

thomaslevesque:
I just looked into it. Proceed indeed doesn't work:

> NotImplementedException: This is a DynamicProxy2 error: The interceptor attempted to 'Proceed' for method 'Int32 get_Area()' which has no target. When calling method without target there is no implementation to 'proceed' to and it is the responsibility of the interceptor to mimic the implementation (set return value, out arguments etc)

Interface methods with default implementation have `IsAbstract == false`

thomaslevesque:
The only workaround I found so far for FakeItEasy is to call the default interface method non-virtually, by emitting IL with a `call` opcode instead of `callvirt`.
https://gist.github.com/thomaslevesque/a4310070de3d44f81144aafd136bea7c
Hopefully it can be done in a cleaner fashion in Castle.Core...

stakx:
@thomaslevesque, I've wanted to  look into this earlier in the past week, now that .NET Core 3 and .NET Standard 2.1 have been released. Unfortunately, I realised quickly that VS2019 as an IDE isn't up to the task: Compilation is very unreliable and haphazard, I get different compilation errors (often bogus ones, too) on every single build. That makes it difficult to properly develop and test against .NET Core 3. I'll see if I'm more lucky with Rider IDE. Otherwise, it'll be the command line, or another wait until VS2019 is actually usable. :disappointed:

thomaslevesque:
> Unfortunately, I realised quickly that VS2019 as an IDE isn't up to the task: Compilation is very unreliable and haphazard, I get different compilation errors (often bogus ones, too) on every single build.

Really? It seems to be working fine for me. Have you tried cleaning everything (bin/obj/.vs) and killing the build server processes?

thomaslevesque:
Anyway, I wonder what the correct behavior should be...
Should `Proceed` call the default implementation automatically? Or should it be explicit?
I'm leaning toward the former, but it might have side effects that I'm not seeing...

stakx:
> Have you tried cleaning everything (bin/obj/.vs) and killing the build server processes?

There's no way on Earth that I am going to do these contortions every time I need to use VS. I'd rather stay with VS1017. :laughing:

[...]

PR #621, opened 2022-05-17
Fix `MissingMethodException` when proxying interfaces containing sealed methods

stakx:
This is a first small step towards support for default interface methods (#447).

Not certain whether this warrants a changelog entry on its own, since what is being fixed here never got reported as an issue. Perhaps we just wait with a changelog entry until we have fully addressed #447?

P.S.: marking this as a draft as it may be incomplete; the same issue may still be present with proxies that have a target.


jonorossi (approved):

stakx:
This may not be totally complete (the same problem may persist for other proxy types) but it's still a step in the right direction. So instead of letting this rot away, I'll merge it. I'll likely pick this topic up again as I'm still working on better support for new interface capabilities.

PR #624, opened 2022-05-18
Add documentation about DynamicProxy internals

stakx:
This is work in progress. I'm planning on adding much more information, especially about the structure and significance of all those similarly-named internal classes in the `Castle.DynamicProxy` sub-namespaces. I'll also try to add links from the generated proxy type code exampels to the parts of DynamicProxy that are responsible for creating those bits of code.

[View it rendered here.](https://github.com/stakx/Castle.Core/blob/documentation/dynamicproxy-internals/docs/dynamicproxy-internals.md)

I'll gladly accept any suggestions and corrections (but it's probably still a little too early to review right now).


stakx:
@304NotModified, I was thinking the same earlier. But I should finish with a rough outline of what's still missing, instead of just stopping. 😁

PR #625, opened 2022-05-20
Use standard invocation type for class proxy methods w/o target

stakx:
This follows in the footsteps of #573, where we introduced a standard `IInvocation` implementation type and re-used it for all methods of interface proxies without target. This present PR does the same for abstract methods of class proxies without target: those would always get a custom-built invocation type with identical `InvokeMethodOnTarget` implementations; so they, too, are eligible for invocation type reuse.

The tests cover all combinations of (a) `abstract` vs. `virtual`, (b) `public` vs. `protected`, and (c) generic vs. non-generic method:

 * (a) is in fact the only distinction that really matters.
 * I've included (b) because of [`EmitCallThrowOnNoTarget` apparently depends on the presence of a callback method](https://github.com/castleproject/Core/blob/0176c319ea9feecf911e2fcf35319eaead866ea8/src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs#L109-L114), and `protected` would regularly lead to callback methods being emitted.
 * I've included (c) i.e. generics mainly because I did so in #573 (even though TBH I cannot quite remember why I thought it necessary back then).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+Enhancements:
+- Performance improvement with proxy type generation for class proxies (without target). Abstract class methods now reuse a predefined invocation type (like methods of interface proxies without target; see explanation below at version 5.0.0 enhancements) (@stakx, #625)
+
 Bugfixes:
 - DynamicProxy emits invalid metadata for redeclared event (@stakx, #590)
 
--- a/ref/Castle.Core-net462.cs
+++ b/ref/Castle.Core-net462.cs
@@ -2788,6 +2788,12 @@ namespace Castle.DynamicProxy.Internal
         protected abstract override void InvokeMethodOnTarget() { }
     }
     [System.Serializable]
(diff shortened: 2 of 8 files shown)


src/Castle.Core/DynamicProxy/Internal/ClassMethodWithoutTargetInvocation.cs
```diff
@@ -0,0 +1,42 @@
...
+	using System;
+	using System.ComponentModel;
+	using System.Diagnostics;
+	using System.Reflection;
+
+#if FEATURE_SERIALIZATION
+	[Serializable]
+#endif
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public sealed class ClassProxyMethodWithoutTargetInvocation : InheritanceInvocation
```
stakx, line 26: I'm probably going to rename this new type to `InheritanceInvocationWithoutTarget`. Far less specific to a single use case than the current name (and thus more likely to be reusable for other purposes), while still perfectly accurate.

stakx:
Oops, GitHub automatically closed this PR when I renamed my branch. Will reopen the PR. Sorry about that.

Issue #593, opened 2021-07-14
Inheritance of Attributes from Interface Members


bepursuant:
Hello team - This might be an easy one to put to rest, I think I just need a castle expert to look at it.

I have [an issue](https://github.com/aloneguid/config/issues/107) with a separate project, and I think it comes down to something in the Castle proxy generation.

The root of the question is - _can Castle allow my generated proxy to inherit attributes from its source Interface?_  The linked issue has an example project demonstrating the disconnect.

stakx:
Hi @bepursuant. I don't remember all the implementation details without looking them up, but off the top of my head:

As a rule of thumb, **DynamicProxy only replicates non-inherited custom attributes.** There is currently no mechanism or option to change that behavior. The only thing you _can_ do is globally specify (via [`Castle.DynamicProxy.Generators.AttributesToAvoidReplicating.Add(Type)`](https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs#L39)) custom attribute types that should not be replicated... but that is the opposite of what you need.

Your [`DescriptionAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.descriptionattribute)s will not be replicated because `DescriptionAttribute`s are inherited attributes. Let's check its definition:

```csharp
[AttributeUsage(AttributeTargets.All)]
public class DescriptionAttribute : Attribute
```

Its `[AttributeUsage]` does not explicitly specify `Inherited = false`, so the [default](https://docs.microsoft.com/en-us/dotnet/api/system.attributeusageattribute.inherited#property-value) (`Inherited = true`) applies.

Knowing this, you could try subclassing `DescriptionAttribute` and then reapply `[AttributeUsage]` with `Inherited = false`. If you then use that custom attribute type, DynamicProxy should replicate it on the generated proxy type.

Let's try this:

```csharp
using System;
using System.ComponentModel;
using System.Reflection;

using Castle.DynamicProxy;

var generator = new ProxyGenerator();

var foo = generator.CreateInterfaceProxyWithoutTarget<IFoo>();
var fooAttr = foo.GetType().GetProperty("Property").GetCustomAttribute<DescriptionAttribute>();
Console.WriteLine(fooAttr?.Description);

var bar = generator.CreateInterfaceProxyWithoutTarget<IBar>();
[...]

PR #633, opened 2022-09-17
Support replication of attributes defined in AttributesToAlwaysReplicate list

### LINKED ISSUE #607: [FEATURE] Allow replication of attributes on mixins or attributes on inherited interfaces
I have a scenario where the attribute I want on the proxy's property is not being replicated over from a mixin.
The attribute itself is sealed and I can't modify the AttributeUsage to set Inherited to false.

The feature itself was considered once before in
https://github.com/castleproject/Core/blob/d88e94447150996973d947a150ffda5efc0c3c1c/src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs#L119
Quoting the comment

> ```
> // Later, there might be more special cases requiring attribute replication,
> // which might justify creating a `SpecialCaseAttributeThatShouldBeReplicated`
> // method and an `AttributesToAlwaysReplicate` class. For the moment, `Param-
> // ArrayAttribute` is the only special case, so keep it simple for now:
> ```

Also it would be nice if we could also apply Attributes to proxy members and not only the class using ProxyGenerationOptions :>)

shoaibshakeel381:
- This is complementary to `AttributesToAvoidReplicating`
- It will allow consumers to specify attributes which should always be copied to generated proxy. It is especially useful where we do not own target interfeces and/or cannot override attributes, e.g. attributes used by various frameworks. See #593 for more context
- Similar effect could by achieved in Castle.Core Version 4.4.1 by simply extending various generator classes but since 5.0 all those classes were marked internal, forcing client code to override half the library. This fix offers an alternative path while keeping to code conventions already in place.
- Fixes #607

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
@@ -53,7 +53,7 @@ namespace Castle.DynamicProxy.Generators
 
 		internal static bool ShouldAdd(Type attribute)
 		{
-			return attributes.Any(attr => attr == attribute);
+			return attributes.Any(attr => attr.IsAssignableFrom(attribute));
 		}
 	}
 }
--- a/src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
@@ -1,0 +1,59 @@
+// Copyright 2004-2021 Castle Project - http://www.castleproject.org/
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Generators
+{
+	using System;
+	using System.Collections.Generic;
+	using System.Linq;
+
+	public static class AttributesToAlwaysReplicate
+	{
+		private static readonly object lockObject = new object();
+
+		private static IList<Type> attributes;
+
+		static AttributesToAlwaysReplicate()
+		{
+			attributes = new List<Type>()
+			{
+				typeof(ParamArrayAttribute)
+			};
+		}
+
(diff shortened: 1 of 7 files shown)


src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Castle.DynamicProxy.Generators
+{
+	using System;
+	using System.Collections.Generic;
+	using System.Linq;
+
+	public static class AttributesToAlwaysReplicate
```
shoaibshakeel381, line 21: see [AttributesToAvoidReplicating](https://github.com/castleproject/Core/blob/master/src/Castle.Core/DynamicProxy/Generators/AttributesToAvoidReplicating.cs) for comparison

jonorossi (approved):

src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		public static bool Contains(Type attribute)
+		{
+			return attributes.Contains(attribute);
+		}
+
+		internal static bool ShouldAdd(Type attribute)
+		{
+			return attributes.Any(attr => attr == attribute);
```
stakx, line 56: I suspect this should be `attr.IsAssignableFrom(attribute)` (like in `AttributesToAvoidReplicating`) instead of `attr == attribute`, so that the two facilities work in the same manner.

When we only had to check for `ParamArrayAttribute` (in `AttributeUtil.cs`) , using `==` was good enough and in fact more efficient, since that FCL type is `sealed` and thus cannot have subclasses. But when we allow any types to be registered, we no longer have such a guarantee.

stakx (changes requested):
Sorry it took me so abysmally long to review this.

I'm in favor of this change, however there's one detail that might need adjusting before we merge... see review comment below.

P.S. could you please also add a new entry to `CHANGELOG.md` for this new feature?

```diff
 # Castle Core Changelog

 # Unreleased
 
 Enhancements:
 - ...
+- <Short description of the new feature goes here> (@shoaibshakeel381, #633)

 ## 5.1.1 (2022-12-30)
```

src/Castle.Core/DynamicProxy/Generators/AttributesToAlwaysReplicate.cs
shoaibshakeel381: I have updated my branch with this change.

shoaibshakeel381:
@stakx  I have applied all requested changes. Please review

stakx (approved):

stakx:
Looks good to me! Thanks for your contribution @shoaibshakeel381, and sorry again for letting you wait so long.

PR #639, opened 2022-12-12
Enable replication of nested custom attribute types

### LINKED ISSUE #638: Nested custom attribute types do not get replicated
#637 incidentally revealed that custom attributes do not get replicated if their custom attribute type is a nested type. Given these two tests:

```csharp
namespace Castle.DynamicProxy.Tests;

using System;

using NUnit.Framework;

[TestFixture]
public class Tests : BasePEVerifyTestCase
{
    [Test]
    public void Nested_non_inherited_attribute_is_replicated()
    {
        var proxy = generator.CreateInterfaceProxyWithoutTarget<IHasNestedNonInheritedAttribute>();
        var proxyType = proxy.GetType();
        var customAttributesOnProxyType = proxyType.GetCustomAttributes(typeof(NestedNonInheritedAttribute), inherit: false);
        CollectionAssert.IsNotEmpty(customAttributesOnProxyType);
    }

    [Test]
    public void Non_nested_non_inherited_attribute_is_replicated()
    {
        var proxy = generator.CreateInterfaceProxyWithoutTarget<IHasNonNestedNonInheritedAttribute>();
        var proxyType = proxy.GetType();
        var customAttributesOnProxyType = proxyType.GetCustomAttributes(typeof(NonNestedNonInheritedAttribute), inherit: false);
        CollectionAssert.IsNotEmpty(customAttributesOnProxyType);
    }

    [NestedNonInherited]
    public interface IHasNestedNonInheritedAttribute { }

    [NonNestedNonInherited]
    public interface IHasNonNestedNonInheritedAttribute { }

    [AttributeUsage(AttributeTargets.Interface, Inherited = false)]
    public sealed class NestedNonInheritedAttribute : Attribute { }
}

// NOTE: it is relevant to the above tests that the following type not be a nested type!

[AttributeUsage(AttributeTargets.Interface, Inherited = false)]
public sealed class NonNestedNonInheritedAttribute : Attribute { }
```

Only of of the tests fails, more specifically, `Nested_non_inherited_attribute_is_replicated`.

The problem is likely caused here:

https://github.com/castleproject/Core/blob/c6d7a165ac3050ebfc70debabe52ee6624a97bdf/src/Castle.Core/DynamicProxy/Internal/AttributeUtil.cs#L188

We should probably augment this test with an additional condition `... && attribute.IsNestedPublic == false`.

... or was there a specific reason against replicating nested custom attribute types?

stakx:
Fixes #638.

TL;DR, this boils down to the following change:

```diff
 private static bool ShouldSkipAttributeReplication(Type attribute, bool ignoreInheritance)
 {
-    if (attribute.IsPublic == false)
+    if (attribute.IsPublic == false && attribute.IsNestedPublic == false)
 ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Castle Core Changelog
 
+## Unreleased
+
+Bugfixes:
+- Nested custom attribute types do not get replicated (@stakx, #638)
+
 ## 5.1.0 (2022-08-02)
 
 Enhancements:
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ClassWithAttributesTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ClassWithAttributesTestCase.cs
@@ -113,6 +113,53 @@ namespace Castle.DynamicProxy.Tests
 		}
 
 		[Test]
+		public void EnsureProxyHasNestedAttributesOnClassAndMethods()
+		{
+			var instance = (HasNestedNonInheritableAttribute)generator.CreateClassProxy(typeof(HasNestedNonInheritableAttribute), new StandardInterceptor());
+
+			object[] attributes = instance.GetType().GetCustomAttributes(typeof(Nested.NonInheritableAttribute), false).ToArray();
+			Assert.AreEqual(1, attributes.Length);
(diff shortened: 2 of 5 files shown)


stakx:
I am going to merge this without any further review, as it's a pretty straightforward change.

PR #641, opened 2022-12-28
Update `FrameworkDisplayName` in `ref/` contract files

stakx:
It appears that the .NET compilers previously left this custom attribute property empty, but have now started setting it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,12 +35,17 @@ jobs:
         with:
           dotnet-version: 3.1.x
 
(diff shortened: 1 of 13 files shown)


stakx:
What is this new devilry?

Took me a moment to figure out what went wrong here: apparently, .NET tooling started to provide a value for `FrameworkDisplayName`, _except_ the .NET 6 SDK, which left it empty. This has apparently been fixed in https://github.com/dotnet/sdk/pull/24656, and if we use the .NET 7 SDK when targeting .NET 6, the problem goes away.

PR #619, opened 2022-05-13
Add support for covariant returns

stakx:
This should fix #601.

~~Right now, there are only two very basic unit tests.~~ Some more tests may be needed to test various aspects of support for covariant returns. ~~For example:~~

* [x] ~~interplay between covariant returns and generics (if that's a feasible scenario at all)~~
* [x] ~~ensuring DynamicProxy always picks the "most derived" method for proxying~~

Some more things to think about (I'll expand this list as further concerns show up):

* [x] ~~Do we need to detect runtime support for covariant returns, or are the required changes compatible with older platforms that don't support them?~~

Issue #601, opened 2021-11-11
Proxies using Records with a base class broken  using .NET 6 compiler


ajcvickers:
See original issue here: https://github.com/dotnet/efcore/issues/26602

Attempt is made to proxy this:

```C#
public record MyRecord : MyBaseRecord
{
    public int MyInt { get; init; }
}

public record MyBaseRecord
{
}
```

Comment from Jared Parsons on the C# compiler team:

> The C# compiler changed the emit strategy for records in .NET 6 to take advantage of covariant returns. The Clone method now always has a return type that matches the containing type. Covariant returns were added to the runtime and language in .NET 5 but records didn't take advantage of them (just ran out of time). In .NET 6 though we finished off that feature and added it to records.

Full repro: https://github.com/Jejuni/Net6RecordEfCoreProxies

Stack trace:

```
Unhandled exception. System.TypeLoadException: Return type in method 'Castle.Proxies.MyRecordProxy.<Clone>$()' on type 'Castle.Proxies.MyRecordProxy' from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' is not compatible with base type method 'Net6RecordEfCoreProxies.M
yRecord.<Clone>$()'.
   at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
   at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.CreateType(TypeBuilder type)
   at Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.BuildType()
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateType(String name, Type[] interfaces, INamingScope namingScope)
   at Castle.DynamicProxy.Generators.ClassProxyGenerator.<>c__DisplayClass1_0.<GenerateCode>b__0(String n, INamingScope s)
   at Castle.DynamicProxy.Generators.BaseProxyGenerator.<>c__DisplayClass33_0.<ObtainProxyType>b__0(CacheKey _)
   at Castle.Core.Internal.SynchronizedDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Castle.DynamicProxy.Generators.BaseProxyGenerator.ObtainProxyType(CacheKey cacheKey, Func`3 factory)
[...]

PR #643, opened 2022-12-28
Improve support for covariant returns (where generic types are involved)

### LINKED ISSUE #632: Proxies using Records derived from a base generic record broken using .NET 6 compiler
Similar to what occurred in #601, when trying to mock (with Moq) records derived from a base generic record, an exception is thrown.

Repro code:

```csharp
var mock1 = new Mock<MyDerivedRecord>();
var mock2 = new Mock<MyDerivedGenericRecord>();

var a = mock1.Object;
var b = mock2.Object;

public abstract record MyBaseRecord
{
}

public abstract record MyBaseGenericRecord<T>
{
	public T Prop { get; set; }
}

public record MyDerivedRecord : MyBaseRecord
{
}

public record MyDerivedGenericRecord : MyBaseGenericRecord<int>
{
}
```

Variable `a` holds the mocked object as expected, but when initializing `b`, the following is thrown:

```
System.ArgumentException
  HResult=0x80070057
  Message=Type to mock (MyDerivedGenericRecord) must be an interface, a delegate, or a non-sealed, non-static class.
  Source=Moq
  StackTrace:
   at Moq.CastleProxyFactory.CreateProxy(Type mockType, IInterceptor interceptor, Type[] interfaces, Object[] arguments) in C:\projects\moq4\src\Moq\Interception\CastleProxyFactory.cs:line 66
   at Moq.Mock`1.InitializeInstance() in C:\projects\moq4\src\Moq\Mock`1.cs:line 307
   at Moq.Mock`1.OnGetObject() in C:\projects\moq4\src\Moq\Mock`1.cs:line 326
   at Moq.Mock`1.get_Object() in C:\projects\moq4\src\Moq\Mock`1.cs:line 281
   at Program.<Main>$(String[] args) in C:\Users\Cesar\source\repos\ConsoleApp1\Program.cs:line 11

  This exception was originally thrown at this call stack:
    System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
    System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
    Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.BuildType()
    Castle.DynamicProxy.Generators.BaseClassProxyGenerator.GenerateType(string, Castle.DynamicProxy.Generators.INamingScope)
    Castle.Core.Internal.SynchronizedDictionary<TKey, TValue>.GetOrAdd(TKey, System.Func<TKey, TValue>)
    Castle.DynamicProxy.Generators.BaseProxyGenerator.GetProxyType()
    Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(System.Type, System.Type[], Castle.DynamicProxy.ProxyGenerationOptions, object[], Castle.DynamicProxy.IInterceptor[])
    Moq.CastleProxyFactory.CreateProxy(System.Type, Moq.IInterceptor, System.Type[], object[]) in CastleProxyFactory.cs

Inner Exception 1:
TypeLoadException: Return type in method 'Castle.Proxies.MyDerivedGenericRecordProxy_2.<Clone>$()' on type 'Castle.Proxies.MyDerivedGenericRecordProxy_2' from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' is not compatible with base type method 'MyDerivedGenericRecord.<Clone>$()'.
```

_Originally posted by @CesarD in https://github.com/castleproject/Core/issues/601#issuecomment-1213625907_

stakx:
Fixes #632.

It turns out that we only added support for _non-generic_ covariant return types in #619; however there are two additional cases that we need to support, too:

* covariant return types where one of them is generic but the other is not
* covariant return types where both of them are generic

(I'm counting those as two distinct cases due to the code structure found inside `MethodSignatureComparer.EqualSignatureTypes`, which is the method that needs to be augmented.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 Bugfixes:
+- Proxies using records derived from a base generic record broken using .NET 6 compiler (@CesarD, #632)
 - Failure proxying type that has a non-inheritable custom attribute type applied where `null` argument is given for array parameter  (@stakx, #637)
 - Nested custom attribute types do not get replicated (@stakx, #638)
 
--- a/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
@@ -16,6 +16,7 @@ namespace Castle.DynamicProxy.Generators
 {
 	using System;
 	using System.Collections.Generic;
+	using System.Diagnostics;
 	using System.Reflection;
 
 	internal class MethodSignatureComparer : IEqualityComparer<MethodInfo>
@@ -136,6 +137,9 @@ namespace Castle.DynamicProxy.Generators
 
 			static bool IsCovariantReturnTypes(Type x, Type y, MethodInfo xm, MethodInfo ym)
 			{
+				Debug.Assert((xm == null && ym == null)
+				          || (xm != null && ym != null && x == xm.ReturnType && y == ym.ReturnType));
+
 				// This enables covariant method returns for .NET 5 and newer.
 				// No need to check for runtime support, since such methods are marked with a custom attribute;
 				// see https://github.com/dotnet/runtime/blob/main/docs/design/features/covariant-return-methods.md.
--- a/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
@@ -92,7 +92,7 @@ namespace Castle.DynamicProxy.Generators
 			}
 			else if (x.IsGenericType != y.IsGenericType)
 			{
-				return false;
+				return IsCovariantReturnTypes(x, y, xm, ym);
 			}
 
 			if (x.IsGenericParameter)
@@ -129,19 +129,24 @@ namespace Castle.DynamicProxy.Generators
 			{
 				if (!x.Equals(y))
 				{
-					// This enables covariant method returns for .NET 5 and newer.
-					// No need to check for runtime support, since such methods are marked with a custom attribute;
-					// see https://github.com/dotnet/runtime/blob/main/docs/design/features/covariant-return-methods.md.
-					if (preserveBaseOverridesAttribute != null)
-					{
-						return (xm != null && xm.IsDefined(preserveBaseOverridesAttribute, inherit: false) && y.IsAssignableFrom(x))
-						    || (ym != null && ym.IsDefined(preserveBaseOverridesAttribute, inherit: false) && x.IsAssignableFrom(y));
-					}
-
-					return false;
+					return IsCovariantReturnTypes(x, y, xm, ym);
 				}
 			}
 			return true;
+
+			static bool IsCovariantReturnTypes(Type x, Type y, MethodInfo xm, MethodInfo ym)
+			{
+				// This enables covariant method returns for .NET 5 and newer.
+				// No need to check for runtime support, since such methods are marked with a custom attribute;
+				// see https://github.com/dotnet/runtime/blob/main/docs/design/features/covariant-return-methods.md.
+				if (preserveBaseOverridesAttribute != null)
+				{
+					return (xm != null && xm.IsDefined(preserveBaseOverridesAttribute, inherit: false) && y.IsAssignableFrom(x))
+					    || (ym != null && ym.IsDefined(preserveBaseOverridesAttribute, inherit: false) && x.IsAssignableFrom(y));
+				}
+
+				return false;
+			}
 		}
 
 		public bool Equals(MethodInfo x, MethodInfo y)
--- a/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
@@ -121,9 +121,12 @@ namespace Castle.DynamicProxy.Generators
 					return false;
 				}
 
-				for (var i = 0; i < xArgs.Length; ++i)
+				if (IsCovariantReturnTypes(x, y, xm, ym) == false)
 				{
-					 if(!EqualSignatureTypes(xArgs[i], yArgs[i])) return false;
+					for (var i = 0; i < xArgs.Length; ++i)
+					{
+						 if(!EqualSignatureTypes(xArgs[i], yArgs[i])) return false;
+					}
 				}
 			}
 			else
--- a/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
@@ -16,7 +16,6 @@ namespace Castle.DynamicProxy.Generators
 {
 	using System;
 	using System.Collections.Generic;
-	using System.Diagnostics;
 	using System.Reflection;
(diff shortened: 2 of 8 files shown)


stakx:
@jonorossi, I'd appreciate your input on one particular detail here. In principle, this PR is an extension of #619, whose code patterns it duplicates for the most part. After this PR, the `IsCovariantReturnTypes` check will happen in three code branches inside `MethodSignatureComparer.EqualSignatureTypes`, instead of just in one. I've tried to keep those checks out of the "hot" code paths as much as possible in order to not affect runtime performance. If you have any suggestions how they could be relegated to run even less frequently, I'd be interested to hear them.

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
```diff
@@ -129,19 +129,24 @@ private bool EqualSignatureTypes(Type x, Type y, MethodInfo xm = null, MethodInf
...
-						    || (ym != null && ym.IsDefined(preserveBaseOverridesAttribute, inherit: false) && x.IsAssignableFrom(y));
-					}
-
-					return false;
+					return IsCovariantReturnTypes(x, y, xm, ym);
 				}
 			}
 			return true;
+
+			static bool IsCovariantReturnTypes(Type x, Type y, MethodInfo xm = null, MethodInfo ym = null)
```
jonorossi, line 141: Probably don't need default values for `xm` and `ym` since the function needs them to do something useful.

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
```diff
@@ -120,28 +121,39 @@ private bool EqualSignatureTypes(Type x, Type y, MethodInfo xm = null, MethodInf
 					return false;
 				}
 
-				for (var i = 0; i < xArgs.Length; ++i)
+				if (IsCovariantReturnTypes(x, y, xm, ym) == false)
 				{
-					 if(!EqualSignatureTypes(xArgs[i], yArgs[i])) return false;
+					for (var i = 0; i < xArgs.Length; ++i)
+					{
+						 if(!EqualSignatureTypes(xArgs[i], yArgs[i])) return false;
```
jonorossi, line 128: `IsCovariantReturnTypes` gets called for every generic parameter. That'll obviously do nothing as the `MethodInfo`s are null, probably nothing that could be improved without it being a premature micro-optimisation though.

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
```diff
@@ -120,28 +121,39 @@ private bool EqualSignatureTypes(Type x, Type y, MethodInfo xm = null, MethodInf
 					return false;
 				}
 
-				for (var i = 0; i < xArgs.Length; ++i)
+				if (IsCovariantReturnTypes(x, y, xm, ym) == false)
```
jonorossi, line 124: This additional check causes a covariant return type generic parameters not to be checked. Is that always handled by `IsAssignableFrom`?

jonorossi (approved):

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
stakx: It just ocurred to me that the obvious thing to do is to move the covariant return type check from `EqualSignatureTypes` to `EqualReturnTypes`. Then it will be performed _only_ for return types, and we only need it in one location (instead of in three).

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 Bugfixes:
+- Proxies using records derived from a base generic record broken using .NET 6 compiler (@CesarD, #632)
```
jonorossi, line 6: Shouldn't this have your username and likely the pull request since you fixed it?

jonorossi (approved):

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
stakx: I'm not sure I understand. But I'll start with the TL;DR: If you can think of a test case where uninstantiated generic parameters end up being used as the return types, I'm glad to add it to the test suite. I cannot think of such a test case, and I suspect that this might never actually occur in practice.

> This additional check causes a covariant return type generic parameters not to be checked.

When talking about generic parameters, shouldn't we be looking at these bits of code?:

https://github.com/castleproject/Core/blob/451c58fc9510c9faecaf5579fdafa8d65a53e978/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs#L89-L92

https://github.com/castleproject/Core/blob/451c58fc9510c9faecaf5579fdafa8d65a53e978/src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs#L98-L104

I haven't explicitly covered these code branches so far because I wasn't sure that they're actually relevant. Since DynamicProxy only proxies fully instantiated generic types, I suspect we don't actually see (uninstantiated) generic type parameters being used as return types. At least I can't currently think of a test case that would trigger this scenario.

I've now moved my `IsAssignableFrom`-based covariant return type check over to `EqualReturnTypes`, so it will now run even for this scenario (if it actually ever ocurrs). However as you seem to suspect, that won't do much good, because...

> Is that always handled by `IsAssignableFrom`?

`IsAssignableFrom` seems to return `false` when given two generic type parameters, even when they are constrained (via `where T : ...`) such that they are guaranteed to be assignment-compatible.

CHANGELOG.md
stakx: I am generally quite happy to give credits to the person & issue reporting a problem, as opposed to my PR fixing it. So I did the same thing here that I did with #601:

https://github.com/castleproject/Core/blob/451c58fc9510c9faecaf5579fdafa8d65a53e978/CHANGELOG.md#L17

That being said, I'm actually unsure about the original intent behind the changelog's "bugfixes" section: should it reference the bug reports, or the PRs fixing them?

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
jonorossi: Yer, no idea of a test case to include, because as you said it likely couldn't happen. Happy to resolve the conversation and forget I posted it 😉. Thanks for digging into it, you're always so thorough.

CHANGELOG.md
jonorossi: No worries, I think I've included both the reporter (especially when they've provided a good repro) and the fixer before. There was a time when I only listed external contributors. Often bug fixes won't have an issue number, just the pull request.

Happy for you to decide what makes sense.

stakx: OK, in that case I'll leave it as is... for now. Still undecided about what makes the _most_ sense, but it probably doesn't matter all that much, since links on GitHub pages make the relationships between issues and PRs easily discoverable.

src/Castle.Core/DynamicProxy/Generators/MethodSignatureComparer.cs
stakx: ... which is mostly thanks to the detailed questions you ask in your code reviews. I've learnt a lot from those. :+1:

I'll ignore generic type parameters for now. If someone finds an actual real-world case where they become a problem, I'll be happy to fix it.

PR #645, opened 2023-01-14
Refactor `MethodFinder.GetAllInstanceMethods`

stakx:
See commit messages for further details.


src/Castle.Core/DynamicProxy/Internal/TypeUtil.cs
```diff
@@ -17,13 +17,17 @@ namespace Castle.DynamicProxy.Internal
...
+	using System.Linq;
 	using System.Reflection;
 	using System.Reflection.Emit;
 
+	using Castle.DynamicProxy.Generators;
 	using Castle.DynamicProxy.Generators.Emitters;
 
 	public static class TypeUtil
 	{
+		private static readonly Dictionary<Type, MethodInfo[]> instanceMethodsCache = new Dictionary<Type, MethodInfo[]>();
```
jonorossi, line 29: `TypeUtil.GetAllInstanceMethods` is called by `InvocationHelper.ObtainMethod` which is in turn called by `InheritanceInvocation.MethodInvocationTarget`. I know the old code worked this way, but do you think it would be worth changing this to `ConcurrentDictionary` or `SynchronizedDictionary`?

We use `SynchronizedDictionary` quite a bit for these caches in DP, we can probably retire that class since we no longer support .NET Framework 3.5 and Silverlight 😃.

jonorossi (approved):
Nice clean up!

src/Castle.Core/DynamicProxy/Internal/TypeUtil.cs
stakx: True! How about I do another PR where I replace them all with `ConcurrentDictionary<,>`?

stakx: Oh, scratch that idea. It's not as easy as that. I forgot that `SynchronizedDictionary<,>` is basically how we ensure that there is no concurrent Reflection.Emit-ting going on into the same `AssemblyBuilder`... so we can't just replace that with a lock-free dictionary.

I'll still check where we might safely replace `SynchronizedDictionary<,>` with `ConcurrentDictionary<,>` (probably everywhere except `ModuleBuilder.TypeCache`) and if there are several such locations, I'll do a separate PR. Otherwise I might just add a commit to this PR for this single location.

stakx: There are currently 4 sites across the project where `SynchronizedDictionary<,>` gets used, 3 of which could likely be replaced by `ConcurrentDictionary<,>` very easily. I think I'll tackle that in a separate PR.

PR #660, opened 2023-08-28
Stop testing for .NET Core 2.1/3.1 and update some nuget packages

Romfos:
Motivation:
Soon .NET 8 will be released and this PR is part of preparation for this. Small cleanup of build\test system

Changes:
- Stop testing for .NET Core 2.1 and .NET Core 3.1
- Update some nuget packages for testing projects
- Set lang version as 11

This pr do not change product by itself. Only test\build infrastructure. No need to release new nuget package version

.Notes:
.NET Core 2.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-2-1-will-reach-end-of-support-on-august-21-2021/

.NET Core 3.1 is out of support
https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,16 +25,6 @@ jobs:
       # The lowest version gets installed first in order to prevent
       # "a newer version is already installed" install errors.
 
-      - name: Install .NET Core 2.1
-        uses: actions/setup-dotnet@v1
-        with:
-          dotnet-version: 2.1.x
-
-      - name: Install .NET Core 3.1
-        uses: actions/setup-dotnet@v1
-        with:
-          dotnet-version: 3.1.x
-
       - name: Install .NET 6.0
         uses: actions/setup-dotnet@v1
         with:
@@ -60,12 +50,6 @@ jobs:
       # ----
       # Test
       # ----
-
-      - name: Test on .NET Core 2.1
-        run: dotnet test -c Release -f netcoreapp2.1 --no-build --no-restore -l "console;verbosity=detailed"
-
-      - name: Test on .NET Core 3.1
-        run: dotnet test -c Release -f netcoreapp3.1 --no-build --no-restore -l "console;verbosity=detailed"
 
       - name: Test on .NET 6.0
         run: dotnet test -c Release -f net6.0 --no-build --no-restore -l "console;verbosity=detailed"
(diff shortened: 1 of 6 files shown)


stakx:
I tend to agree that testing on these now-unsupported platforms, by itself, isn't that important anymore... however, since this library targets `net462`, `net6.0`, but also `netstandard2.0` and `netstandard2.1`, we need to figure out what the last two TFMs mean for running tests.

IIRC, it is for the .NET Standard TFMs that we're currently testing on .NET Core 2.1 and 3.1.

Given that .NET 5 has started moving away from .NET Standard starting with version 5 (see [the official docs](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0)), one might wonder why we are even keeping these TFMs around at all. .NET 4.6.2 and 6 already cover .NET Standard 2.x, and we're not testing our library for any other platforms (Xamarin, UWP, aunity, etc.); Mono being the sole exception, but that falls under the `net462` umbrella.

The main reason to keep targeting `netstandard2.x` is perhaps because downstream libraries do so, too, and maybe we do not want to force them into abandoning those TFMs?

I suppose dropping `netstandard2.x` support might be the technically honest tging to do, but it might be an unpopular decision. And if we're keeping those TFMs around, perhaps we should also test them in some way.

Maybe I'm overthinking this... any opinions? @jonorossi?

jonorossi:
@stakx I think you've got it right.

Microsoft has the following recommendations for when to keep using .NET Standard, and that's for libraries that want to share between .NET Framework and .NET [Core], which is still useful for libraries that depend on interfaces defined in `Castle.Core`:

> .NET Standard not deprecated
>
> .NET Standard is still needed for libraries that can be used by multiple .NET implementations. We recommend you target .NET Standard in the following scenarios:
>
> * Use netstandard2.0 to share code between .NET Framework and all other implementations of .NET.
> * ...

I agree we should only run unit tests on supported .NET platforms, and the build warnings need to be cleaned up, e.g.:

> .NET Core 2.1 is no longer supported and will not receive security updates in the future.
> (https://github.com/castleproject/Core/actions/runs/5998689016)

stakx:
@jonorossi, cleaning up the build warnings would be easy enough by simply suppressing them (I'll look into it!), but are you suggesting we go further and drop the `netcoreapp*` targets for our test projects? That would leave the `netstandard*` assemblies untested since we have no other runtime we could use for them. (I could probably live with that, esp. because we have no conditional compilation that distinguishes them from the `net6` assembly.)

jonorossi:
@stakx I think best to come back to this after your list of areas to address post the next release. Some parts of the library that'll get pulled out could just target .NET Standard only.

PR #661, opened 2023-08-31
Add support for default interface members in inheritance-based proxy types

stakx:
This is another step towards completing #447. I'm limiting this PR to inheritance-based proxy types (i. e. those without target) due to time constraints &ndash; having to adjust only 2 out of 5 implementations is a lot less work! &ndash; and because these are likely the most important proxy types for downstream testing libraries such as Moq, NSubstitute, etc.

 * [X] methods
 * [X] generic methods
 * [X] properties and events (i. e. where the accessor methods are not "standalone" ones) &ndash; only testing properties at this time, which should be enough
 * [X] different accessibilities (e. g. `protected` ~~, `protected internal`, `internal`~~)
 * [X] mixins
 * [X] additional interfaces
 * [X] `static` members
 * [X] ~~`static abstract` members~~ (requires targeting .NET 7 or newer & a major architectural expansion to include non-instance members; let's do this in a future PR)

Testing all possible combinations of these aspects would be prohibitive, I've therefore restricted tests to a subset of these combinations that seemed reasonable and justifiable by the nature of the actual code changes / extensions made to DynamicProxy.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
@@ -42,6 +42,11 @@ namespace Castle.DynamicProxy.Contributors
 			var targetItem = new ClassMembersCollector(targetType) { Logger = Logger };
 			yield return targetItem;
 
+			foreach (var @interface in targetType.GetAllInterfaces())
+			{
+				yield return new InterfaceMembersWithDefaultImplementationCollector(@interface, targetType);
+			}
+
 			foreach (var @interface in interfaces)
 			{
 				var item = new InterfaceMembersOnClassCollector(@interface, true,
--- a/src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
@@ -44,6 +44,11 @@ namespace Castle.DynamicProxy.Contributors
 
 			foreach (var @interface in targetType.GetAllInterfaces())
 			{
+				if (@interface.HasAnyOverridableDefaultImplementations() == false)
+				{
+					continue;
+				}
+
 				yield return new InterfaceMembersWithDefaultImplementationCollector(@interface, targetType);
 			}
 
--- a/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
@@ -62,7 +62,7 @@ namespace Castle.DynamicProxy.Contributors
 				return null;
 			}
 
-			return new MetaMethod(method, methodOnTarget, true, proxyable, !method.IsAbstract);
+			return new MetaMethod(method, methodOnTarget, isStandalone, proxyable, !method.IsAbstract);
 		}
 	}
 }
--- a/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
@@ -38,10 +38,13 @@ namespace Castle.DynamicProxy.Contributors
 
 		protected override MetaMethod GetMethodToGenerate(MethodInfo method, IProxyGenerationHook hook, bool isStandalone)
 		{
-			if (method.IsAbstract)
+			if (method.IsAbstract || method.IsFinal || method.IsVirtual == false)
 			{
-				// This collector is only interested in methods with default implementations.
+				// This collector is only interested in overridable methods with default implementations.
 				// All other interface methods should be dealt with in other contributors.
+				//
+				// (Note this is the opposite check of that in `TypeUtil.HasAnyOverridableDefaultImplementations`,
+				// which is the method used to filter out whole interfaces before this collector gets run for them.)
 				return null;
 			}
 
--- a/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
+++ b/src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
@@ -38,6 +38,13 @@ namespace Castle.DynamicProxy.Contributors
 
 		protected override MetaMethod GetMethodToGenerate(MethodInfo method, IProxyGenerationHook hook, bool isStandalone)
 		{
+			if (method.IsAbstract)
+			{
+				// This collector is only interested in methods with default implementations.
+				// All other interface methods should be dealt with in other contributors.
+				return null;
+			}
+
 			var index = Array.IndexOf(map.InterfaceMethods, method);
 			Debug.Assert(index >= 0);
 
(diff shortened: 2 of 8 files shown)


src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
```diff
@@ -42,6 +42,16 @@ protected override IEnumerable<MembersCollector> GetCollectors()
 			var targetItem = new ClassMembersCollector(targetType) { Logger = Logger };
 			yield return targetItem;
 
+			foreach (var @interface in targetType.GetAllInterfaces())
```
stakx, line 45: I have been wondering whether adding a cache for `type.GetAllInterfaces()` might be a good idea. It'll now be called at least twice per class proxy type generation (and IIRC it's already being called twice per interface proxy type generation). Probably not the most urgent optimization, but might be good looking into that at some point.

src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright 2004-2022 Castle Project - http://www.castleproject.org/
```
stakx, line 1: Should be `2004-2023`, not `2022`. (I cherry-picked this code from an older code draft of mine.)

src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
```diff
@@ -0,0 +1,59 @@
...
+			if (methodOnTarget.DeclaringType.IsInterface == false)
+			{
+				return null;
+			}
+
+			var proxyable = AcceptMethod(method, true, hook);
+			if (!proxyable)
+			{
+				return null;
+			}
```
stakx, line 54: It might be possible to optimize this further.

 * The `proxyable = AcceptMethod(...)` check could be run first, which would mean that the `map` lookups do not need to run in the negative case.

 * If `method.IsAbstract == true` (i. e. an interface method _without_ a default implementation), then `methodOnTarget.DeclaringType.IsInterface == false`, because implementing classes are then not allowed to omit declaring that method. So we could check `method.IsAbstract` first and then possibly skip the `map` lookups in the positive case.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
```diff
@@ -63,23 +65,32 @@ private Type GetInvocationType(MetaMethod method, ClassEmitter emitter)
...
-			if (canChangeTarget)
-			{
-				invocationInterfaces = new[] { typeof(IInvocation), typeof(IChangeProxyTarget) };
-			}
-			else
-			{
-				invocationInterfaces = new[] { typeof(IInvocation) };
-			}
+			Type[] invocationInterfaces = new[] { typeof(IInvocation), typeof(IChangeProxyTarget) };
 			var key = new CacheKey(methodInfo, CompositionInvocationTypeGenerator.BaseType, invocationInterfaces, null);
```
stakx, line 94: Mostly unrelated to this PR, but as a reminder to myself, I ought to look up & remind myself why interface proxies generally use `CompositionInvocation`s... and whether it's OK to introduce new code that then uses a `InheritanceInvocation` instead.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
```diff
@@ -93,5 +104,24 @@ private Type GetInvocationType(MetaMethod method, ClassEmitter emitter)
 				.Generate(emitter, namingScope)
 				.BuildType());
 		}
+
+		private MethodInfo CreateCallbackMethod(ClassEmitter emitter, MethodInfo methodInfo, MethodInfo methodOnTarget)
+		{
+			var targetMethod = methodOnTarget ?? methodInfo;
+			var callBackMethod = emitter.CreateMethod(namingScope.GetUniqueName(methodInfo.Name + "_callback"), targetMethod);
+
+			if (targetMethod.IsGenericMethod)
```
stakx, line 113: There's currently no test for generic interface methods with default implementations. Might have to add at least one!

src/Castle.Core/DynamicProxy/Contributors/InterfaceMembersWithDefaultImplementationCollector.cs
stakx: I've only applied the second optimization; moving up the `AcceptMethod` check didn't seem so crucial anymore afterwards, and all the other collectors usually run that check last, too, so it's nice to stay consistent with them.

src/Castle.Core/DynamicProxy/Internal/TypeUtil.cs
```diff
@@ -134,6 +135,32 @@ public static Type GetTypeOrNull(object target)
...
+		{
+			Debug.Assert(interfaceType != null);
+			Debug.Assert(interfaceType.IsInterface);
+
+			var cache = hasAnyOverridableDefaultImplementationsCache;
+			lock (cache)
+			{
+				if (!cache.TryGetValue(interfaceType, out var result))
+				{
+					foreach (var method in interfaceType.GetAllInstanceMethods())
```
stakx, line 148: Note that while this operation (`GetAllInstanceMethods`) is relatively expensive, if we didn't pre-screen interfaces using this new `HasAnyOverridableDefaultImplementations` method, we'd still end up calling it for each interface (in `InterfaceMembersWithDefaultImplementationCollector`'s base class). So we're really just calling it a little earlier, and a later, repeated call then becomes much cheaper since it uses a cache internally.

stakx:
@jonorossi, I think I've finally got this in a reasonably understandable & mergable state. I'd appreciate if you could give this at least a cursory review, if you can spare some time.

As noted in the initial description above, this is only for proxies without targets, and I am not testing all possible combinations of DynamicProxy features. Do let me know if you think I have missed any crucial test cases.

Regarding the structure of this PR: most of the commits consist simply of new tests (usually grouped by some aspect listed in the PR's description). Most of those actually pass right away, without any code changes. In the few cases where tests don't pass, the test-only commit is immediately followed by one that makes them succeed. Those actual-code-change commits are sometimes followed by a little refactoring & clean-up work.

src/Castle.Core/DynamicProxy/Contributors/InterfaceProxyWithoutTargetContributor.cs
stakx: The answer is that interface proxies are either composition-based and thus rightly use `CompositionInvocation`, or when they're _not_ composition-based (i. e. interface proxy without target) we've previously replaced `InheritanceInvocation` with the much faster `InterfaceMethodWithoutTargetInvocation`.

src/Castle.Core/DynamicProxy/Contributors/ClassProxyTargetContributor.cs
jonorossi: You'd have to confirm the performance impact, as I believe the runtime team has made reflection improvements in more recent releases to better cache specific type information.

jonorossi (approved):
Looks good, sorry for taking so long to get to this. From a cursory review I don't see anything missing.

Nice work with the extremely neat and organised commits.

stakx:
I think I've let this sit for long enough. I looked over the code again with a fresh set of eyes, this should still be good to go.

Issue #663, opened 2023-09-01
Support by-ref-like (`ref struct`) parameter types such as `Span<T>` and `ReadOnlySpan<T>`


stakx:
DynamicProxy does not currently support those because by-ref-like values cannot be boxed to be put in the `object[] IInvocation.Arguments` array.

I haven't yet been able to think of a general way how _all_ by-ref-like types (including user-defined ones) because of their various limitations. If anyone has ideas, I'd be interested in hearing them!

However, it might be fairly easy to at least add support for `Span<T>` and `ReadOnlySpan<T>` specifically. Those types are becoming more and more common in the .NET FCL, so while not an ideal solution, it might still be sufficient for most use cases:

We could introduce a new type in DynamicProxy's namespace:

```csharp
public interface ISpanMarshaller
{
    object BoxSpan<T>(Span<T> span);
    object BoxReadOnlySpan<T>(ReadOnlySpan<T> readOnlySpan);
    ReadOnlySpan<T> UnboxReadOnlySpan<T>(object boxedReadOnlySpan);
    Span<T> UnboxSpan<T>(object boxedSpan);
}
```

Then we could introduce a new property to `ProxyGenerationOptions`, `ISpanMarshaller SpanMarshaller { get; set; }`, which generated proxies could use to transfer spans into and out of the `object[] IInvocation.Arguments` array.

I don't really like an addition that deals with two very specific types, but they are becoming more and more common, and I haven't yet thought of anything better.

Opinions, or alternate ideas, anyone?

stakx:
I've done some quick prototyping. The extension proposed above is feasible and works.

(I'd however perhaps make one change to the proposal: instead of adding `ISpanMarshaller SpanMarshaller { get; set; }` to `ProxyGenerationOptions`, it might be better to add a `Type SpanMarshallerType { get; set; }` instead. Proxies would then instantiate a marshaller instance themselves, instead of getting one injected. This has a few benefits:

 * Marshaller types do not need to be serializable.
 * It's clearer that they are supposed to be stateless (because it's out of the user's control when & how often they get instantiated).
 * It's easier to implement. :wink:)

stakx:
Scratch the above proposal, I think I can actually come up with one that works for arbitrary by-ref-like types. I'll post an updated feature proposal shortly.

stakx:
[...]

PR #664, opened 2023-09-02
Draft: Add support for by-ref-like (`ref struct`) parameter types such as `Span<T>` and `ReadOnlySpan<T>`

stakx:
This is a partially working proof of concept for #663, but it still needs quite a bit of work:

 * [X] perform value conversion for inbound by-re-flike values (in by-value and `in` method parameters)
 * [ ] perform value conversion for outbound by-ref-like values (in method return values, `ref` and `out` parameters)
 * [ ] add unit tests
 * [ ] add documentation on how to use `ProxyGenerationOptions.ByRefLikeConverterSelector`
 * [ ] ideally, support generic converter types

### Usage example 1

This shows how one would set up a converter for by-ref-like argument values. In this case, one gets set up such that `Span<byte>` arguments will get converted to a `byte[]` copy during invocation:

```csharp
// this will get invoked during interception:
public class CopyByteSpanConverter
{
    // a by-ref-like converter must be accesible from the dynamically generated assembly, and it must have...

    // (1) for inbound conversions, a public `Box` method
    //     with a single `in` parameter of a by-ref-like type, and return type `object`:
    public object Box(in Span<byte> span)
    {
        return span.ToArray();
    }

    // (2) for outbound conversions, a public `Unbox` method
    //     with an `object` parameter and a `ref` parameter of a by-ref-like type:
    public void Unbox(object argument, ref Span<byte> span)
    {
        var array = (byte[])argument;
        array.CopyTo(span);
    }
}

// this will get invoked during proxy type generation:
class ConverterSelector : IByRefLikeConverterSelector
{
    public Type SelectConverterType(MethodInfo _, int _, Type parameterType)
    {
        return parameterType == typeof(Span<byte>) ? typeof(CopyByteSpanConverter) : null;
    }
}

// this is what we want to proxy:
public interface IFoo
{
    void Method(in Span<byte> bytes);
}

var options = new ProxyGenerationOptions { ByRefLikeConverterSelector = new ConverterSelector() };
var proxy = generator.CreateInterfaceProxyWithoutTarget<IFoo>(options, ...);
var span = new byte[] { ... }.AsSpan();
proxy.Method(in span);  // this argument will end up in the invocation's `Arguments` as a `byte[]` array!
```

### Usage example 1

If no converter is set up, by-ref-like argument values no longer cause an exception during invocation, but now get converted to `null` references in the `IInvocation.Arguments` array.

```csharp
// this is what we want to proxy:
public interface IFoo
{
    void Method(in Span<byte> bytes);
}

var span = new byte[] { ... }.AsSpan();
proxy.Method(in span);  // this argument will end up in the interceptor's `Arguments` as a `null` reference
```


stakx:
Superseded by #712.

PR #670, opened 2023-09-28
Update DictionaryAdapter TypeKeyPrefix example

jerone:
Update DictionaryAdapter `TypeKeyPrefix` example to show where the namespace is coming from.


jonorossi:
Thanks

PR #685, opened 2024-08-31
Draft: Support collectible assemblies via a new `CollectibleProxyBuilder`

stakx:
This is an alternative to #679. Consider it a proof of concept that is to be discussed further, over in #473.

### DIFF
--- a/src/Castle.Core/DynamicProxy/ModuleScope.cs
+++ b/src/Castle.Core/DynamicProxy/ModuleScope.cs
@@ -60,7 +60,7 @@ namespace Castle.DynamicProxy
 		private readonly object moduleLocker = new object();
 
 		// Specified whether the generated assemblies are intended to be saved
-		private readonly bool savePhysicalAssembly;
+		private readonly AssemblyBuilderAccess assemblyBuilderAccess;
 		private readonly bool disableSignedModule;
 		private readonly INamingScope namingScope;
 
@@ -78,6 +78,16 @@ namespace Castle.DynamicProxy
 		/// <param name = "savePhysicalAssembly">If set to <c>true</c> saves the generated module.</param>
 		public ModuleScope(bool savePhysicalAssembly)
 			: this(savePhysicalAssembly, false)
+		{
+		}
+
+		/// <summary>
+		///   Initializes a new instance of the <see cref = "ModuleScope" /> class,
+		///   allowing to specify the <see cref="AssemblyBuilderAccess" /> to be used for generated assemblies.
+		/// </summary>
+		/// <param name = "assemblyBuilderAccess">The desired <see cref="AssemblyBuilderAccess" /> to be used for the generated assemblies.</param>
+		public ModuleScope(AssemblyBuilderAccess assemblyBuilderAccess)
+			: this(assemblyBuilderAccess, disableSignedModule: false, new NamingScope(), DEFAULT_ASSEMBLY_NAME, DEFAULT_FILE_NAME, DEFAULT_ASSEMBLY_NAME, DEFAULT_FILE_NAME)
 		{
 		}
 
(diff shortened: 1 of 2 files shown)


src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -133,8 +143,21 @@ public ModuleScope(bool savePhysicalAssembly, bool disableSignedModule, string s
 		internal ModuleScope(bool savePhysicalAssembly, bool disableSignedModule, INamingScope namingScope,
 		                     string strongAssemblyName, string strongModulePath,
 		                     string weakAssemblyName, string weakModulePath)
+			: this(
+#if FEATURE_ASSEMBLYBUILDER_SAVE
+				  assemblyBuilderAccess: savePhysicalAssembly ? AssemblyBuilderAccess.RunAndSave : AssemblyBuilderAccess.Run,
+#else
+				  assemblyBuilderAccess: AssemblyBuilderAccess.Run,
+#endif
```
stakx, line 151: This is the code location where we could add additional logic to default to `RunAndCollect` instead of `Run` depending on the current `AssemblyLoadContext`.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -81,6 +81,16 @@ public ModuleScope(bool savePhysicalAssembly)
...
 
+		/// <summary>
+		///   Initializes a new instance of the <see cref = "ModuleScope" /> class,
+		///   allowing to specify the <see cref="AssemblyBuilderAccess" /> to be used for generated assemblies.
+		/// </summary>
+		/// <param name = "assemblyBuilderAccess">The desired <see cref="AssemblyBuilderAccess" /> to be used for the generated assemblies.</param>
+		public ModuleScope(AssemblyBuilderAccess assemblyBuilderAccess)
+			: this(assemblyBuilderAccess, disableSignedModule: false, new NamingScope(), DEFAULT_ASSEMBLY_NAME, DEFAULT_FILE_NAME, DEFAULT_ASSEMBLY_NAME, DEFAULT_FILE_NAME)
+		{
+		}
```
stakx, line 92: This new constructor overload doesn't offer any other options for strong-naming and assembly save paths. We can always add more overloads, but for now it seems improbable that users would be much interested in those aspects on newer platforms where `AssemblyBuilderAccess.RunAndCollect` is functioning well... and if you're not after `RunAndCollect`, you have all the preexisting overloads already.

stakx: Decided to change this back to `internal` for now. We can always make it `public` later if desired.

src/Castle.Core/DynamicProxy/ModuleScope.cs
stakx: As mentioned in https://github.com/castleproject/Core/issues/473#issuecomment-2390642865, auto-detection logic may indeed not be the most pressing feature, so I am omitting any such logic, at least for now.

stakx:
@castleproject/committers, I'll let this PR sit for a few days, in case anyone wants to review. If there are no objections, I'll go ahead and merge it in a few days' time as the enhancement to DynamicProxy it contains is fairly trivial.

jonorossi (approved):
@stakx Looks good. Could we add a very basic unit test to ensure the this new proxy builder at least works?

stakx:
Thanks for reviewing.

You're right, I forgot to write a test, since this started out only as a proof of concept. I'll add one.

simonferquel:
Hi @stakx, I have some reserve about this approach as it will require all moqing frameworks to opt-into leverage this to be able to work without leaking Assembly Load Contexts in environments that leverage them.

The example of Unity 7 where we are moving to CoreCLR and take advantage of ALCs (AssemblyLoadContext) to allow users to iterate on their code without reloading the editor:
- We load user code and packages into an unloadable ALC. This user code may contain tests which uses any Moqing framework built on top of Castle.Core
- Whenever the user modify its code or update a package, we cleanup all references the engine has to user object, and unload the user ALC, compile new version of code, and reload the new ALC

If any non-unloadable ALC keep references to anything loaded within the User ALC, the unload will never complete (the GC will see outstanding references to some objects belonging to that ALC and thus will keep it alive). We call that ALC leaks in Unity.

With that PR, if a Moqing framework relying on Castle.Core wants to play nice with Unity (or any other environment that does leverage unloadable ALCs), it will need to explicitly opt-into using this new CollectibleProxyBuilder. I don't think it is the responsibility of the Moqing framework, but more about the hosting process to decide in which kind of ALC generated code should get loaded: that is why I first introduced a global static setting that a runtime environment could use to define this behavior (e.g. if Unity detects that Castle.Core is loaded, it would set the setting accordingly so that any moqing framework would generate its code in a way that plays nice with Unity runtime). 
The default setting value that would depend on whether Castle.Core itself is loaded in a Collectible ALC makes sense to me, as it means it as been loaded as a dependency of something that may be collected at some point in the process lifetime, and so whatever it generated dynamically should not be able to keep its ALC alive and should thus be collectible. But I could live with that logic not enabled.

So what I really think we should have to define if the builders should emit collectible types is this:
- A global setting that could let an hosting environment set the safe defaults (with optionally the "sensible defaults" I introduced in the original PR relying on detecting if Castle.Core assembly itself is in a collectible ALC)
- An explicit way to let direct users of Castle.Core override this global setting

That could be implemented by having: 

- `DefaultProxyBuilder` add a `protected virtual bool GeneratedCollectibleTypes` property, implemented by looking at the global setting
- Adding a `CollectibleProxyBuilder` and `NonCollectibleProxyBuilder` that would override the property to ignore the global setting

What would you think of that ?

PR #687, opened 2024-09-17
Make classes sealed, methods static & avoid array allocations where possible, dispose disposable, other optimizations & clean up usings

Henr1k80:
I was surprised to see how [popular this package is](https://www.nuget.org/stats) and I would like to try reduce the carbon impact of it.  
It might not be directly measurable, but the compound impact of all the executions hopefully makes a difference 🤞🏻

- Where possible:
   - Classes sealed
   - Methods static
   - Avoid unneeded array allocations  

- Dispose disposable
- Other minor optimizations™®©
- Clean up `using`s

There should be no breaking changes, unless you do reflection hacks and that should not be a supported use case.

### DIFF
--- a/ref/Castle.Core-net462.cs
+++ b/ref/Castle.Core-net462.cs
@@ -1945,10 +1945,11 @@ namespace Castle.Core.Internal
             where T : System.Attribute { }
         public static System.Type GetTypeConverter(System.Reflection.MemberInfo member) { }
     }
-    public static class InternalsVisible
+    public class InternalsVisible
     {
         public const string ToCastleCore = @"Castle.Core, PublicKey=002400000480000094000000060200000024000052534131000400000100010077F5E87030DADCCCE6902C6ADAB7A987BD69CB5819991531F560785EACFC89B6FCDDF6BB2A00743A7194E454C0273447FC6EEC36474BA8E5A3823147D214298E4F9A631B1AFEE1A51FFEAE4672D498F14B000E3D321453CDD8AC064DE7E1CF4D222B7E81F54D4FD46725370D702A05B48738CC29D09228F1AA722AE1A9CA02FB";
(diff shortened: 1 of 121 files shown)


stakx:
Hi @Henr1k80 and thanks for contributing! ✨

A couple points:

* Making classes `sealed` or declaring them `static` where possible is a change to the public contract. We could schedule such changes for the next major version release, but earlier than that I think that the cost (avoidable breaking changes) outweighs the benefit (cleaner and perhaps minutely faster code).
* DictionaryAdapter is pretty much legacy by now, I personally have no interest in developing this further, including doing code reviews on it. (Sorry.) Maybe someone else wants to invest their time here?
* Avoiding unnecessary array allocations sounds great! Let's do this.

So I think it would be great if you could split up this PR into several ones, each dealing with one specific aspect so that we can deal with them separately

* I'd be happy to review and merge one focusing on unnecessary array allocations.
* I'd be willing to do the same for one focusing on `sealed`/`static` etc. but only at a later time (when we're planning the next major version).
* I'm reluctant about a PR doing things to DictionaryAdapter. But if there's enough general interest and support for it, I could perhaps be talked into it.

OK for you to proceed like this? If so, I'd be closing this PR in favor of the upcoming one(s).

Henr1k80:
Hi @stakx 

> Making classes sealed or declaring them static where possible is a change to the public contract. We could schedule such changes for the next major version release, but earlier than that I think that the cost (avoidable breaking changes) outweighs the benefit (cleaner and perhaps minutely faster code).

I have been very meticulous securing that there are no changes to the public contract, only internal & private classes have been made sealed and static.  
The only exceptions are the public classes that only contains `const`s. There are no code changes needed for accessing these consts.

stakx:
@Henr1k80, see the changes in the `ref/` files. Those are indications that there are indeed changes to the public contract (API surface).

(My above points and the question re: whether to proceed as suggested still stands btw.)

Henr1k80:
@stakx ok, I understand your concerns and have removed the changes that affect the `ref/` files

stakx:
@Henr1k80, you have since submitted separate PRs addressing the same concerns raised in this PR, so I am going to close this one as it is now duplicated by the others.

PR #688, opened 2024-10-14
Optimize array usage, plus other minor tweaks

Henr1k80:
Avoid array allocations, where possible, part of https://github.com/castleproject/Core/pull/687 (excluding DictionaryAdapter)
Plus other minor tweaks

### DIFF
--- a/src/Castle.Core/Core/Resource/CustomUri.cs
+++ b/src/Castle.Core/Core/Resource/CustomUri.cs
@@ -39,7 +39,7 @@ namespace Castle.Core.Resource
 			{
 				throw new ArgumentNullException(nameof(resourceIdentifier));
 			}
-			if (resourceIdentifier == string.Empty)
+			if (resourceIdentifier.Length == 0)
 			{
 				throw new ArgumentException("Empty resource identifier is not allowed", nameof(resourceIdentifier));
 			}
--- a/src/Castle.Core/Core/Resource/AssemblyResource.cs
+++ b/src/Castle.Core/Core/Resource/AssemblyResource.cs
@@ -104,7 +104,7 @@ namespace Castle.Core.Resource
 			string nameFound = null;
 			foreach(string name in names)
 			{
-				if (string.Compare(resourcePath, name, StringComparison.OrdinalIgnoreCase) == 0)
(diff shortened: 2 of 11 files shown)


src/Castle.Core/Core/Resource/CustomUri.cs
```diff
@@ -39,7 +39,7 @@ public CustomUri(string resourceIdentifier)
 			{
 				throw new ArgumentNullException(nameof(resourceIdentifier));
 			}
-			if (resourceIdentifier == string.Empty)
+			if (resourceIdentifier.Length == 0)
```
stakx, line 42: This looks like an anti-optimization. Computing the length of a string requires processing all of its characters; comparing to the empty string means you can stop after at most 1 character.

Please revert.

stakx (changes requested):
Looks mostly good to me!

src/Castle.Core/Core/Resource/CustomUri.cs
Henr1k80: Length is an integer field on string, so it is not calculated, just read, and then it is simply comparing two integers.

Comparing to string.Empty will on the other hand:
1. Do a reference comparison, if the other string is the same as string.Empty, stop the comparison.
2. If they are not reference equals, it will do a value comparison
so likely two comparisons and branches instead of just one.

stakx: > Length is an integer field on string, so it is not calculated

While your reasoning here is flawed sup>(*)</sup>, the conclusion appears to be correct nevertheless. I did a brief research and `string.Length` indeed appears to get cached. So yeah, I was wrong and this should _not_ make performance worse.

In my opinion, this still isn't the best refactoring (less semantic, among other things), but then it's probably not worth losing any sleep over.

Let's go with your version.

---

<sup>(*)</dup> Just because something is integer-typed doesn't mean it is not calculated. `string.Length` is a property after all, not a plain field, so at least in theory it is very much possible that there would be a `strlen` of some kind hidden inside it... but luckily that doesn't appear to be the case.

stakx (approved):

stakx:
Thank you for contributing, @Henr1k80!

PR #692, opened 2024-11-14
Include DeclaringType.Name for nested types when generating explicit implementation names

JelleKerkstra:
This PR resolves the `DynamicProxyException` `Duplicate element: Castle.DynamicProxy.Generators.MetaMethod` when generating a proxy that implements the following interfaces due to the nested classes.
In the current implementation, only the namespace, type name and method name are considered. In this PR, the declaring type names are also used when generating explicit interface implementations due to matching method names.

```
public static class OuterWrapper
{
	public static class InnerWrapperA
	{
		public interface ISharedName
		{
			void M();
		}
	}

	public static class InnerWrapperB
	{
		public interface ISharedName
		{
			void M();
		}
	}

	public static class InnerWrapperC
	{
		public interface ISharedName
		{
			void M();
		}
	}
}
```

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/MetaTypeElement.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/MetaTypeElement.cs
@@ -62,23 +62,30 @@ namespace Castle.DynamicProxy.Generators
 					nameBuilder.Append(ns);
 					nameBuilder.Append('.');
 				}
-				AppendTypeName(nameBuilder, sourceType);
+				AppendTypeName(nameBuilder, sourceType, true);
 				nameBuilder.Append('.');
 				nameBuilder.Append(name);
 				this.name = nameBuilder.ToString();
 			}
 			else if (ns != null)
 			{
-				this.name = string.Concat(ns, ".", sourceType.Name, ".", name);
+				this.name = string.Concat(ns, ".", GetFullTypeName(sourceType), ".", name);
 			}
 			else
 			{
-				this.name = string.Concat(sourceType.Name, ".", name);
+				this.name = string.Concat(GetFullTypeName(sourceType), ".", name);
 			}
 
-			static void AppendTypeName(StringBuilder nameBuilder, Type type)
+			static void AppendTypeName(StringBuilder nameBuilder, Type type, bool fullName = false)
 			{
+				if (fullName)
+				{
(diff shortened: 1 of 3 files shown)


src/Castle.Core/DynamicProxy/Generators/MetaTypeElement.cs
```diff
@@ -62,23 +62,30 @@ protected void SwitchToExplicitImplementationName()
 					nameBuilder.Append(ns);
 					nameBuilder.Append('.');
 				}
-				AppendTypeName(nameBuilder, sourceType);
+				AppendTypeName(nameBuilder, sourceType, true);
```
stakx: Why not simply call `AppendFullTypeName` directly instead of going through `AppendTypeName` using that new parameter?

src/Castle.Core/DynamicProxy/Generators/MetaTypeElement.cs
```diff
@@ -62,23 +62,30 @@ protected void SwitchToExplicitImplementationName()
...
 				nameBuilder.Append('.');
 				nameBuilder.Append(name);
 				this.name = nameBuilder.ToString();
 			}
 			else if (ns != null)
 			{
-				this.name = string.Concat(ns, ".", sourceType.Name, ".", name);
+				this.name = string.Concat(ns, ".", GetFullTypeName(sourceType), ".", name);
```
stakx: If I remember correctly, the pre-existing code in `SwitchToExplicitImplementationName` was written the way it is based on the assumption that `string.Concat` would be faster than instantiating a `StringBuilder`. The first `if` clause is therefore the slow path, while the other two are supposedly fast paths.

Your code change introduces `StringBuilder`s in the fast paths, too, so from a performance perspective, they're now all the same and it seems likely that the distinction between three cases no longer needs to be made.

So if we go with your change, that whole method could perhaps be simplified to a single case where a `StringBuilder` is always instantiated at the start and there are various `if`s for the presence/absence of a namespace, for the presence/absence of generic type arguments, etc.

If we go with the existing code's structure, you'd have to find a way of *not* relying on `StringBuilder` on the fast paths.

I think I'd favor simplifying the code by rewriting the `if` conditions to something more obvious, even if it means we no longer have the `string.Concat` fast paths.

Issue #683, opened 2024-08-30
Should we drop support for .NET Core 2.1 and 3.1?


stakx:
On recent builds, we get a bunch of out-of-support warnings for .NET Core 2.1 and 3.1:

```
warning NETSDK1138: The target framework 'netcoreapp2.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
warning NETSDK1138: The target framework 'netcoreapp3.1' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
```

For .NET Core 2.1, we additionally get security vulnerability warnings:

```
warning NU1903: Package 'Microsoft.NETCore.App' 2.1.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-2xjx-v99w-gqf3
warning NU1902: Package 'Microsoft.NETCore.App' 2.1.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-3gp9-h8hw-pxpw
warning NU1903: Package 'Microsoft.NETCore.App' 2.1.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-3w5p-jhp5-c29q
warning NU1902: Package 'Microsoft.NETCore.App' 2.1.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-5633-f33j-c6f7
warning NU1903: Package 'Microsoft.NETCore.App' 2.1.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-6px8-22w5-w334
warning NU1903: Package 'Microsoft.NETCore.App' 2.1.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-g5vf-38cp-4px9
warning NU1902: Package 'Microsoft.NETCore.App' 2.1.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-vgwq-hfqc-58wv
warning NU1902: Package 'Microsoft.NETCore.App' 2.1.0 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-x5qj-9vmx-7g6g
```

Simply suppressing those warnings doesn't seem right, I think we should first decide whether or not we want to continue supporting those out-of-support frameworks at all.

### See also:

* The last discussion we had on this topic: https://github.com/castleproject/Core/pull/660
* .NET and .NET Core Support Policy: https://aka.ms/dotnet-core-support

[...]

PR #693, opened 2025-01-20
Fix closure

gao-artur:
`InvocationHelper.GetMethodOnType` is in on a hot path and creates relatively much garbage.
`ProxyUtil.AreInternalsVisibleToDynamicProxy` is not on a hot path but was easy to fix.
There is another closure in `BaseProxyGenerator.GetProxyType`, which is a bit harder to fix, but since it's not a hot path I left it as is. Let me know if you want to fix it too.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
+++ b/src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
@@ -50,7 +50,7 @@ namespace Castle.DynamicProxy.Internal
 
 			var cacheKey = new CacheKey(proxiedMethod, type);
 
-			return cache.GetOrAdd(cacheKey, ck => ObtainMethod(proxiedMethod, type));
+			return cache.GetOrAdd(cacheKey, static ck => ObtainMethod(ck.Method, ck.Type));
 		}
 
 		private static MethodInfo ObtainMethod(MethodInfo proxiedMethod, Type type)
--- a/src/Castle.Core/DynamicProxy/ProxyUtil.cs
+++ b/src/Castle.Core/DynamicProxy/ProxyUtil.cs
@@ -162,9 +162,9 @@ namespace Castle.DynamicProxy
 		/// <param name="asm">The assembly to inspect.</param>
 		internal static bool AreInternalsVisibleToDynamicProxy(Assembly asm)
 		{
-			return internalsVisibleToDynamicProxy.GetOrAdd(asm, a =>
+			return internalsVisibleToDynamicProxy.GetOrAdd(asm, static a =>
 			{
-				var internalsVisibleTo = asm.GetCustomAttributes<InternalsVisibleToAttribute>();
+				var internalsVisibleTo = a.GetCustomAttributes<InternalsVisibleToAttribute>();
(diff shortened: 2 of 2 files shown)


src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
```diff
@@ -50,7 +50,7 @@ public static MethodInfo GetMethodOnType(Type type, MethodInfo proxiedMethod)
 
 			var cacheKey = new CacheKey(proxiedMethod, type);
 
-			return cache.GetOrAdd(cacheKey, ck => ObtainMethod(proxiedMethod, type));
+			return cache.GetOrAdd(cacheKey, static ck => ObtainMethod(ck.Method, ck.Type));
```
stakx, line 53: (Theoretically generated code could be reduced by changing `ObtainMethod` such that it accepts a single `CacheKey` argument, then it could be used directly here in place of the lambda function wrapping it. But that's a different refactoring perhaps for another time.)

stakx (approved):

src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
gao-artur: Do you mean method group? I don't mind doing that if you like.

stakx:
If there are more such cases that are equally trivial to fix, then by all means, please go ahead.

However when the refactorings get more complicated / tim intensive, or when they negatively affect code legibility / make the code harder to understand, those are probably cases we should leave unmodified as the change wouldn't seem worth it.

I'm happy to merge this but if you want to add more, I can wait a little. Just let me know how you'd like to proceed.

gao-artur:
Fixing other places will require adding a third argument to `SynchronizedDictionary` similar to [TArg factoryArgument](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-9.0#system-collections-concurrent-concurrentdictionary-2-getoradd-1(-0-system-func((-0-0-1))-0)) in `ConcurrentDictionary.GetOrAdd` and passing all the dependencies in a value tuple. I don't think it is worth it.

src/Castle.Core/DynamicProxy/Internal/InvocationHelper.cs
stakx: I meant replacing the whole `ck => ObtainMethod(ck.Method, ck.Type)` with just `ObtainMethod`, which might be possible if that method were made to accept a single`CacheKey` argument instead of two. But I just mentioned in it passing, no need to change that.

stakx:
Looks like .NET broke our test suite once again. Will need to fix that in a separate PR before we can merge this one here.

jonorossi:
> Looks like .NET broke our test suite once again. Will need to fix that in a separate PR before we can merge this one here.

@stakx We had planned to drop .NET Core 2.1 and other old runtimes, and things have moved on from the discussion in #683.

snakefoot:
Guess I can made a dedicated pull-request using my [last commit](https://github.com/castleproject/Core/pull/694/commits/a891322463c65e94e36037106ff2ae9e61fbb6d9) from #694 (That fixes the build-pipeline by removing obsolete netcoreapp2.1 + netcoreapp3.1)

PR #696, opened 2025-07-12
Castle.Core modernization

Romfos:
Changed:
- Now target frameworks are: .NET 8, .NET Framework 4.6.2, .NET Standard 2.0
- Tests platforms now are: .NET Framework 4.6.2, .NET 8, .NET 9
- Lang version now is 12
- Dependencies were updated
- Github action are updated

Removed:
- .NET Core 2.1, .NET Core 3.1, .NET 6, and mono tests
- .NET Standard 2.1 tfm
- Castle.Core.Tests.WeakNamed project (was run only on mono)


Romfos:
Can someone assist with appveyor? Do we need it if we have github actions?

Romfos:
cc @stakx @jonorossi

.github/workflows/build.yml
```diff
@@ -28,31 +25,16 @@ jobs:
...
-        with:
-          dotnet-version: 6.0.x
-
-      # Building requires an up-to-date .NET SDK.
+      # -----
+      # Setup .NET sdk
+      # -----
 
-      - name: Install .NET 7.0
+      - name: Install dotntet sdk
```
stakx:
```suggestion
      - name: Install .NET SDKs
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Castle Core Changelog
 
+## Next
+
+Enhancements:
+- Now target frameworks are: .NET 8, .NET Framework 4.6.2, .NET Standard 2.0
+- Tests platforms now are: .NET Framework 4.6.2, .NET 8, .NET 9
```
stakx, line 7: It might be good to explicitly list the removed platforms in a separate "Deprecations:" section (see e.g. the changelog for version 5.0.0).

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+- Lang version now is 13
```
stakx: Not sure if this last bit ("Lang version...") needs to be in the changelog, as it typically won't affect consumers of the `Castle.Core` package.

It's a possibly important detail to contributors though, so we should mention the updated C# language version somewhere. Probably in the `README`'s "Building" section, which currently still says C# 9.

src/Castle.Core.Tests/Castle.Core.Tests.csproj
```diff
@@ -42,32 +42,22 @@
...
-	<ItemGroup Condition="'$(TargetFramework)'=='netcoreapp3.1'">
-		<PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
-		<PackageReference Include="System.Net.Primitives" Version="4.3.0" />
-	</ItemGroup>
 	
-	<ItemGroup Condition="'$(TargetFramework)'=='net6.0'">
-		<PackageReference Include="PublicApiGenerator" Version="10.1.2" />
+	<ItemGroup Condition="'$(TargetFramework)'=='net8.0'">
+		<PackageReference Include="System.Security.Permissions" Version="8.0.0" />
+		<PackageReference Include="PublicApiGenerator" Version="11.4.6" />
```
stakx, line 60: Asking out of curiousity: why not include PublicApiGenerator for the newest .NET version supported (9)? Is it because the actual library projects only target .NET 8 (unlike the test projects)?

src/Castle.Core.Tests/DynamicProxy.Tests/DefaultInterfaceMembersTestCase.cs
```diff
@@ -397,12 +397,12 @@ public void Can_proxy_interface_with_static_method()
...
-		public void Can_proxy_interface_with_static_abstract_method()
-		{
-			_ = generator.CreateInterfaceProxyWithoutTarget<IHaveStaticAbstractMethod>();
-		}
+		//[Test]
+		//[Ignore("Support for static abstract interface members has not yet been implemented.")]
+		//public void Can_proxy_interface_with_static_abstract_method()
+		//{
+		//	_ = generator.CreateInterfaceProxyWithoutTarget<IHaveStaticAbstractMethod>();
+		//}
```
stakx: I'm not sure I understand this change. Why comment out the test? Does `[Ignore]` no longer work?

src/Castle.Services.Logging.SerilogIntegration/Castle.Services.Logging.SerilogIntegration.csproj
```diff
@@ -25,7 +25,7 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Serilog" Version="2.0.0" />
+		<PackageReference Include="Serilog" Version="3.0.0" />
```
stakx: I'm not sure whether this dependency update should be mixed up with the changes to the targeted platforms.

Leave it as is for now... I may or may not get back to you about this later.

CHANGELOG.md
Romfos: removed

CHANGELOG.md
Romfos: added

.github/workflows/build.yml
Romfos: updated

src/Castle.Services.Logging.SerilogIntegration/Castle.Services.Logging.SerilogIntegration.csproj
Romfos: restored

src/Castle.Core.Tests/DynamicProxy.Tests/DefaultInterfaceMembersTestCase.cs
Romfos: this code just wasn't compiled and wasn't run before

image

before this PR tests were run <= .NET 6

src/Castle.Core.Tests/Castle.Core.Tests.csproj
Romfos: this is related to this code:
https://github.com/castleproject/Core/blob/4d62205a258dcdf4c14452b37840d6a2d64cf813/src/Castle.Core.Tests/PublicApiTestCase.cs#L15

Why this was implemented like that? - have no idea
Make sense to create a separate issue and remove these `#ifdef` sections

this is out of scope of my change, I just did version update for this

src/Castle.Core.Tests/DynamicProxy.Tests/DefaultInterfaceMembersTestCase.cs
stakx: Ah, I think I remember now: This test was written in anticipation of .NET 7 and some new features it would introduce, and the test was meant to explicitly document (in the form of an additional ignored test) that support for those would be missing.

Please revert this change, otherwise we lose this intended reminder.

Romfos: > Ah, I think I remember now: This test was written in anticipation of .NET 7 and some new features it would introduce, and the test was meant to explicitly document (in the form of an additional ignored test) that support for those would be missing.
> 
> Please revert this change, otherwise we lose this intended reminder.

I cannot do it because of error on then screenshot above. Now we will have tests for .NET 8/9 and this code just do not compile

this code should be uncommented when this feature will be implemented. Or I can remove this code at all

stakx: I see.

I think it would be good to keep the test for its intended purpose as an implementation reminder, but it also shouldn't hold you back at this moment.

How about we only comment out the erroneous bit and leave a note for later?:

```suggestion
		[Test]
		[Ignore("Support for static abstract interface members has not yet been implemented.")]
		public void Can_proxy_interface_with_static_abstract_method()
		{
			// NOTE: uncommented because of compilation error CS8920, to be reviewed when we implement support for static abstract interface members
			// _ = generator.CreateInterfaceProxyWithoutTarget<IHaveStaticAbstractMethod>();
		}
```

(If that should bring up yet another complication, we'll do it your way and comment out the whole thing... for now.)

Romfos: done

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
```diff
@@ -23,18 +23,11 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
-		<PackageReference Include="NUnit.Console" Version="3.11.1" />
-		<PackageReference Include="NUnit3TestAdapter" Version="3.16.1" />
-		<PackageReference Include="NUnitLite" Version="3.13.3" />
```
stakx, line 29: I am not sure whether I understand and agree with the removal of these package references. This is a test project, no? So shouldn't it explicitly reference the packages necessary to run its tests?

I am assuming that test execution still works after removing these is that they're indirectly referenced via the `Castle.Core.Tests.csproj` project reference. Or am I missing something else?

buildscripts/build.cmd
```diff
@@ -37,20 +37,12 @@ echo --------------------
...
-
-
-echo ---------------------------
-echo Running NET6.0 Tests
-echo ---------------------------
-
-dotnet .\src\Castle.Core.Tests\bin\%Configuration%\net6.0\Castle.Core.Tests.dll --result=Net60TestResults.xml;format=nunit3 || exit /b 1
-dotnet .\src\Castle.Core.Tests.WeakNamed\bin\%Configuration%\net6.0/Castle.Core.Tests.WeakNamed.dll --result=Net60WeakNamedTestResults.xml;format=nunit3 || exit /b 1
+dotnet .\src\Castle.Core.Tests\bin\%Configuration%\net8.0\Castle.Core.Tests.dll --result=Net80TestResults.xml;format=nunit3 || exit /b 1
+dotnet .\src\Castle.Core.Tests.WeakNamed\bin\%Configuration%\net8.0/Castle.Core.Tests.WeakNamed.dll --result=Net80WeakNamedTestResults.xml;format=nunit3 || exit /b 1
```
stakx, line 48: At first glance, it struck me as a little odd that we're running tests on .NET 8 but not on .NET 9 also.

That being said, I have no problem with it: after all we _are_ running tests on .NET 9 during the GitHub Actions build, and we're probably going to get rid of the Appveyor build altogether sometime soon, so I don't think it's worth investing too much time and effort here.

No further action required here.

stakx (approved):

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
Romfos: That is right. They are referenced as transitive dependencies from Castle.Core.Tests.csproj

PR #701, opened 2025-12-07
Draft: Enable saving dynamic assemblies to disk on .NET 9+ using `PersistentAssemblyBuilder` 

stakx:
This is the smallest code change I can think of off the top of my head to enable `PersistentAssemblyBuilder.SaveAssembly()` on .NET 9+. It represents only a code exploration to test feasibility, but it is far from an ideal solution and won't be merged in this form.

The approach chosen here works using the following approach:

* Dynamic types are generated using .NET 9's new `PersistedAssemblyBuilder`. The major obstacle here is that it does not support type activation. In order to activate a type, the assembly is emitted into a stream, which can then be loaded into the current assembly load context like any regular assembly.

* This means that the dynamic assembly is "baked" after every single proxy type activation, and it cannot be reused to emit further types. Because of that, the `ModuleScope` gets replaced with a new one after every single proxy type creation.

* Because of the `ModuleScope` recycling, we essentially lose the type cache functionality. In theory, this functionality could be restored by lifting the type cache out of `ModuleScope` into `DefaultProxyBuilder` (so that it can span several module scopes). This refactoring would be fairly involved if breaking changes in the public API were to be avoided.

Something similar to this *might* be good enough to run out test suite (including PE / IL verification of generated assemblies) on .NET 9, as we could likely do well without the type cache in this scenario.


src/Castle.Core/DynamicProxy/Generators/BaseClassProxyGenerator.cs
```diff
@@ -102,7 +102,9 @@ protected sealed override Type GenerateType(string name, INamingScope namingScop
 			// Crosses fingers and build type
 
 			var proxyType = emitter.BuildType();
+#if !NET9_0_OR_GREATER
 			InitializeStaticFields(proxyType);
+#endif
```
stakx, line 107: This must be skipped here because the generated type cannot be activated while tied to a `PersitedAssemblyBuilder`.

src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs
```diff
@@ -114,7 +124,28 @@ public Type CreateInterfaceProxyTypeWithoutTarget(Type interfaceToProxy, Type[]?
...
+			{
+				lastAssemblyGenerated?.Dispose();
+				var stream = new MemoryStream();
+				persistedAssemblyBuilder.Save(stream);
+				stream.Seek(0, SeekOrigin.Begin);
+				var assembly = AssemblyLoadContext.Default.LoadFromStream(stream);
+				type = assembly.GetType(type.FullName!)!;
+				lastAssemblyGenerated = stream;
+				lastScope = scope;
+				scope = scope.Recycle();
```
stakx, line 143: This would likely have to be made thread-safe.

src/Castle.Core/DynamicProxy/DefaultProxyBuilder.cs
```diff
@@ -20,6 +20,11 @@ namespace Castle.DynamicProxy
 	using System.Collections.Generic;
 	using System.Linq;
 	using System.Reflection;
+#if NET9_0_OR_GREATER
```
stakx, line 23: Instead of using `NET9_0_OR_GREATER` we'd probably want a `FEATURE_PERSISTEDASSEMBLYBUILDER` conditional compilation symbol.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -450,7 +465,11 @@ private AssemblyName GetAssemblyName(bool signStrongName)
 				}
 				assemblyBuilder = (AssemblyBuilder)StrongNamedModule.Assembly;
 				assemblyFileName = StrongNamedModuleName;
+#if NET9_0_OR_GREATER
+				assemblyFilePath = weakModulePath;
+#else
```
stakx, line 470: This alternative assignment may actually work also for .NET 4.6.2+, as `[Weak|Strong]NamedModule.FullyQualifiedName` is possibly equal to `[weak|strong]ModulePath`.

src/Castle.Core/DynamicProxy/ModuleScope.cs
```diff
@@ -472,7 +495,12 @@ private AssemblyName GetAssemblyName(bool signStrongName)
 			AddCacheMappings(assemblyBuilder);
 #endif
 
+#if FEATURE_ASSEMBLYBUILDER_SAVE
 			assemblyBuilder.Save(assemblyFileName);
+#elif NET9_0_OR_GREATER
+			var persistedAssemblyBuilder = (PersistedAssemblyBuilder)assemblyBuilder;
+			persistedAssemblyBuilder.Save(assemblyFilePath);
```
stakx, line 502: Looks like this is left-over code that (if it worked) would have to be called by the `PersistentProxyBuilder` on its `lastScope`. This code has been replaced with code directly in `PersistentProxyBuilder` however... probably not ideal.

src/Castle.Core/DynamicProxy/PersistentProxyBuilder.cs
```diff
@@ -43,7 +47,17 @@ public PersistentProxyBuilder() : base(new ModuleScope(true))
...
 		public string? SaveAssembly()
 		{
+#if NET9_0_OR_GREATER
+			var assemblyPath = lastScope.WeakNamedModule != null ? lastScope.WeakNamedModuleName : lastScope.StrongNamedModuleName;
+			using var file = File.Create(assemblyPath);
+			lastAssemblyGenerated.Seek(0, SeekOrigin.Begin);
+			lastAssemblyGenerated.CopyTo(file);
+			file.Flush();
+			file.Close();
+			return assemblyPath;
```
stakx, line 57: This should perhaps be replaced with `return lastScope.SaveAssembly();`.

PR #702, opened 2025-12-09
Switch to C# language version 14

### LINKED ISSUE #698: Compatibility issue with C# 14: `FieldReference.LoadReference()` throws ArgumentNullException
### Description

While working with Castle.Core and targeting **C# 14**, I encountered an issue in the `FieldReference` class that causes a failure when attempting to load a field.

Image

The exception occurs because the `Reference` property is referencing its backing field, but not the desired class field.

### Link to Source
src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/FieldReference.cs:56

### Steps to Reproduce
Use Castle.DynamicProxy to generate a proxy that includes a baked field (e.g., via newer C# 14 features).
Attempt to invoke a method that triggers FieldReference.LoadReference().
Observe the runtime failure due to ArgumentNullException.

### Expected Behavior
The proxy generation should emit compatible IL instructions without throwing runtime exceptions.

### Proposed Fix
Using `this.field` or `@field` to ensure the correct field reference is used.
This eliminates any assumptions regarding the specific field being referenced and conforms with baked fields introduced in newer C# versions.

### Environment
commit: 4d62205a258dcdf4c14452b37840d6a2d64cf813
.NET SDK: 8.0+
C# Language Version: 14

stakx:
Resolves #698.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,7 @@ jobs:
           dotnet-version: |
             8.0.x
             9.0.x
+            10.0.x
 
       # -----
       # Build
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,6 +18,12 @@ for:
     only:
     - image: Ubuntu2204
 
+  # install .NET 10 SDK
+  install:
+  - curl -L https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
+  - chmod +x ./dotnet-install.sh
+  - sudo ./dotnet-install.sh --channel 10.0 --install-dir /usr/share/dotnet
+
   # build and run tests
   build_script:
   - uname -a
@@ -45,6 +51,10 @@ for:
       {
           Update-AppveyorBuild -Version ($env:APPVEYOR_REPO_TAG_NAME).TrimStart("v")
       }
+
+  # install .NET 10 SDK
+  install:
+  - cmd: choco install dotnet-10.0-sdk
 
   # build and run tests
   build_script:


src/Castle.Core.Tests/DynamicProxy.Tests/CustomModifiersTestCase.cs
```diff
@@ -150,7 +150,7 @@ public void ReflectionReturnsCorrectModoptOnParamTypeForGeneratedType(string typ
 
 			var modopts = this.generatedTypes[typeName].GetMethod("Foo").GetParameters()[0].GetOptionalCustomModifiers();
 
-			CollectionAssert.AreEqual(expected: CustomModifiersTestCase.customModifiers[typeNameWithoutSuffix].Reverse(), actual: modopts);
+			CollectionAssert.AreEqual(expected: Enumerable.Reverse(CustomModifiersTestCase.customModifiers[typeNameWithoutSuffix]), actual: modopts);
```
stakx, line 153: This is necessary because apparently, C# 14 changed the method group overload resolution such that `array.Reverse()` now resolves to `Span<>.Reverse` (which reverses in-place and returns `void`) instead of `Enumerable.Reverse` (which returns a new, reversed sequence).

See e.g. https://medium.com/c-sharp-programming/csharp-14-reverse-array-bug-cfe0742870e3.

src/Castle.Core/DynamicProxy/Generators/Emitters/SimpleAST/FieldReference.cs
```diff
@@ -53,7 +53,7 @@ public FieldBuilder FieldBuilder
 
 		public FieldInfo Reference
 		{
-			get { return field; }
+			get { return @field; }
```
stakx, line 56: This is what fixes the issue mentioned above.

PR #707, opened 2025-12-15
Refactor `SimpleAST` for clear distinction between expressions and locations

stakx:
This is in preparation for #663, where it will be necessary to adjust many code locations dealing with conversion of arguments that get passed around both by-value and by-reference. I find DynamicProxy's handling of by-ref arguments especially hard to understand, and that is in part due to the structuring of the CIL syntax trees.

This PR does lots of little things, but the main goal is introducing a clear distinction in the `SimpleAST` syntax nodes between expressions/statements and references. The latter are being renamed "locations" to better match ECMA-335 terminology (see section I.8.3).

Way too many of the `Reference` node types aren't really locations at all, and are thus "down-gradeed" to expressions.

There's also some code duplication being removed here. Nodes that do too many different things are being replaced with simpler single-purpose nodes that compose better than the previous node types.

### DIFF
--- a/src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs
+++ b/src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs
@@ -46,12 +46,12 @@ namespace Castle.DynamicProxy.Generators
 		///   Generates the constructor for the class that extends
 		///   <see cref = "AbstractInvocation" />
 		/// </summary>
-		protected abstract ArgumentReference[] GetBaseCtorArguments(Type targetFieldType,
+		protected abstract ArgumentLocation[] GetBaseCtorArguments(Type targetFieldType,
 		                                                            out ConstructorInfo baseConstructor);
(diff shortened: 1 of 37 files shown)


src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs
```diff
@@ -149,7 +149,7 @@ protected virtual void ImplementInvokeMethodOnTarget(AbstractTypeEmitter invocat
 					}
 
 					invokeMethodOnTarget.CodeBuilder.AddStatement(new AssignStatement(localReference, localValue));
-					var byRefReference = new ByRefReference(localReference);
+					var byRefReference = new ByRefExpression(localReference);
```
stakx: `byRefExpression`, not `byRefReference`

stakx:
I'm not really happy yet with this PR. I'll probably split it up into a few smaller ones.

PR #712, opened 2025-12-19
Add support for byref-like (`ref struct`) parameter and return types such as `Span<T>` and `ReadOnlySpan<T>`

### LINKED ISSUE #651: ProxyGenerator for Stream not .NET7 compatible
The following repro sample unfortunately fails with the .NET SDK 7.0.200 (it worked like a charm with .NET 6).
I'm using Castle.Core 5.1.1.

```
using System.Text;
using System.Xml;
using System.Xml.Linq;
using Castle.DynamicProxy;

var xmlString = "<MyXml></MyXml>";
var stream = new MemoryStream(Encoding.UTF8.GetBytes(xmlString));

// THIS IS THE CULPRIT - COMMENTING LINE 12 MAKES IT WORK AGAIN
stream = proxyGenerator.CreateClassProxyWithTarget(stream, new MyInterceptor());

using var reader = XmlReader.Create(stream); // FAILS
var xdoc = XDocument.Load(reader);
Console.WriteLine(xdoc.Root!.Name);

class MyInterceptor : IInterceptor
{
  public void Intercept (IInvocation invocation)
  {
    invocation.Proceed();
  }
}
```

The exception is:
> Unhandled exception. System.InvalidProgramException: Common Language Runtime detected an invalid program.
   at Castle.Proxies.MemoryStreamProxy.Read(Span`1 buffer)
   at System.IO.Stream.ReadAtLeastCore(Span`1 buffer, Int32 minimumBytes, Boolean throwOnEndOfStream)
   at System.IO.Stream.ReadAtLeast(Span`1 buffer, Int32 minimumBytes, Boolean throwOnEndOfStream)
   at System.Xml.XmlTextReaderImpl.InitStreamInput(Uri baseUri, String baseUriStr, Stream stream, Byte[] bytes, Int32 byteCount, Encoding encoding)
   at System.Xml.XmlTextReaderImpl.FinishInitStream()
   at System.Xml.XmlReader.Create(Stream input)
   at Program.<Main>$(String[] args) in C:\Projects\ReproInvalidProgramExceptionXmlReader\ReproInvalidProgramExceptionXmlReader\Program.cs:line 13
   
 Please fix this soon, it prevents us from upgrading to .NET 7.
 
Best regards,
D.R.

### LINKED ISSUE #663: Support by-ref-like (`ref struct`) parameter types such as `Span<T>` and `ReadOnlySpan<T>`

stakx:
Unlike my earlier draft #664, this PR will not require DynamicProxy user code to set up any value converters for byref-like parameters.

Instead, any byref-like argument will automatically get substituted in `IInvocation` with a "reference" type:

* ~~`SpanArgument<T>`~~  ~~`SpanProxy<T>`~~  `SpanReference<T>` for `Span<T>` values
* ~~`ReadOnlySpanArgument<T>`~~  ~~`ReadOnlySpanProxy<T>`~~  `ReadOnlySpanReference<T>` for `ReadOnlySpan<T>` values
* ~~`ByRefLikeProxy`~~  `ByRefLikeReference` on .NET 8 for any non-span by-ref-like values
* ~~`ByRefLikeArgument<TByRefLike>`~~  ~~`ByRefLikeProxy<TByRefLike>`~~  `ByRefLikeReference<TByRefLike>` on .NET 9+ for any non-span by-ref-like values of type `TByRefLike`

Each of these (except the non-generic `ByRefLikeReference`) has a `ref`-returning `Value` property for accessing the actual value.

These class types are essentially references to the actual byref-like parameters. They use unmanaged pointers (`void*`) under the hood. I've added a big comment in the `ByRefLikeReference.cs` code file explaining why I think this is safe.

```csharp
public interface IFoo
{
    void A(Span<char> characters);
    void B(ref Span<char> characters);
}

public void FooAInterceptor : IInterceptor
{
    public void Intercept(IInvocation invocation);
    {
        var charactersRef = invocation.Arguments[0] as SpanReference<char>;
        Span<char> characters = charactersRef.Value;
    }
}

public void FooBInterceptor : IInterceptor
{
    public void Intercept(IInvocation invocation);
    {
        var charactersRef = invocation.Arguments[0] as SpanReference<char>;
        Span<char> characters = charactersRef.Value;
        charactersRef.Value = characters[0..1]);
    }
}
```

This should now be ready for merging:

* [x] add support for reading/writing byref-like argument values and propagating them through the interception pipeline
* [x] add support for byref-like return values
* [x] update / replace unit tests in `ByRefLikeTestCase` (which still expect byref-like parameters to default / get nullified)
* [x] write a documentation article on how to properly interact with byref-like arguments in `IInvocation` &rarr; preview it [here](https://github.com/stakx/Castle.Core/blob/c9f7a729387f9788c5a502e4b1119fc57e228d94/docs/dynamicproxy-byref-like-parameters.md)
* [x] update `ref/` contract files
* [x] update the changelog

Will fix #651 and close #663 once completed and merged.


src/Castle.Core/DynamicProxy/IByRefLikeProxy.cs
```diff
@@ -0,0 +1,87 @@
...
+	///   DynamicProxy replaces these unboxable values with <see cref="IByRefLikeProxy{TByRefLike}"/> references
+	///   (or alternatively, in the case of spans, with <see cref="ISpanProxy{T}"/> or <see cref="IReadOnlySpanProxy{T}"/>),
+	///   which grant you indirect read/write access to the actual values.
+	///   <para>
+	///     These references are only valid for the duration of the intercepted method call.
+	///     Any attempt to use it beyond that will result in a <see cref="AccessViolationException"/>.
+	///   </para>
+	/// </remarks>
+	/// <typeparam name="TByRefLike">A by-ref-like (<c>ref struct</c>) type.</typeparam>
+	public interface IByRefLikeProxy<TByRefLike> where TByRefLike : struct, allows ref struct
```
stakx: The name of this and the other interface types does not seem quite right. "Proxy" suggests a stand-in object with exactly the same public surface as the one being proxied (as is the case with DynamicProxy proxies). This isn't the case here: these "proxy" interface types bring their own distinct APIs.

I think "accessor" would probably be a better name category than "proxy", and DynamicProxy already has such a type with a pair of "get"/"set" methods: `IProxyTargetAccessor`.

(Note: in a previous draft, I called those types `ByRefLikeArgument`, `SpanArgument`, etc. &ndash; I moved away from those names because the types may also be used for the return parameter.)

I'm probably going to rename `IByRefLikeProxy` &rarr; `IByRefLikeAccessor`, `ISpanProxy` &rarr; `ISpanAccessor` etc.

src/Castle.Core/DynamicProxy/Internal/ByRefLikeProxy.cs
```diff
@@ -0,0 +1,237 @@
...
+	// As far as I can reason, `ByRefLikeProxy` et al. should be safe to use IFF they are never copied out from an
+	// `IInvocation`, and IFF DynamicProxy succeeds in destructing them and erasing them from the `IInvocation` right
+	// before the intercepted method finishes executing.
+
+	/// <summary>
+	///   Do not use! Only DynamicProxy internals may interact with this class type directly.
+	/// </summary>
+	[CLSCompliant(false)]
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	public unsafe class ByRefLikeProxy
```
stakx: I apparently forgot to commit the required `<AllowUnsafeBlocks>` setting in the project file.

stakx: I think I have a better name for the interface types (see other review comment), but I'm not sure about the class types' names. It doesn't seem right to call them "accessors". Perhaps "reference", e. g. `ByRefLikeProxy` &rarr; `ByRefLikeReference`?

stakx: I have been thinking about adding `[DebuggerTypeProxy]` and/or `[DebuggerBrowsable]` attributes to automatically display the referenced by-ref-like values in debugger windows instead of the wrapper objects. Looking at a memory address isn't exactly helpful.

src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
```diff
@@ -103,7 +105,10 @@ protected override MethodEmitter BuildProxiedMethodBody(MethodEmitter emitter, C
 
 			var argumentsMarshaller = new ArgumentsMarshaller(emitter, MethodToOverride.GetParameters());
 
-			argumentsMarshaller.CopyIn(out var argumentsArray, out var hasByRefArguments);
+			argumentsMarshaller.CopyIn(out var argumentsArray, out var hasByRefArguments, out var hasByRefLikeArguments);
+
+			// TODO: If the return type is by-ref-like, we should prepare a local variable and a `ByRefLikeProxy` for it
+			// and place it in `IInvocation.ReturnValue`, so that the interception pipeline has a means to return something.
```
stakx: This TODO needs to be implemented so we can return by-ref-like values, too.

src/Castle.Core/DynamicProxy/Internal/ByRefLikeProxy.cs
```diff
@@ -0,0 +1,237 @@
...
+
+			this.type = type;
+			this.ptr = (nint)ptr;
+		}
+
+		public void* GetPtr(Type checkType)
+		{
+			if (checkType != type)
+			{
+				throw new AccessViolationException();
```
stakx: Since we'll probably be adding a documentation page regarding by-ref-like values during interception, perhaps we could add exception messages with a link to the documentation page for more info...?

src/Castle.Core/DynamicProxy/Generators/MethodWithInvocationGenerator.cs
```diff
@@ -259,23 +270,49 @@ public void CopyIn(out LocalReference argumentsArray, out bool hasByRefArguments
...
-						// because it cannot be boxed. We need to replace it with some other value.
+						hasByRefLikeArguments = true;
 
-						// For now, we just erase it by substituting `null`:
+						// By-ref-like values live exclusively on the stack and cannot be boxed to `object`.
+						// Instead of them, we prepare instances of `ByRefLikeProxy` wrappers that reference them.
+
+#if NET9_0_OR_GREATER
+						// TODO: perhaps we should cache these `ConstructorInfo`s?
+						ConstructorInfo proxyCtor = typeof(ByRefLikeProxy<>).MakeGenericType(dereferencedArgumentType).GetConstructors().Single();
```
stakx: Should implement `ConstructorInfo` caching here, as the "todo" comment suggest.

Since `Span<T>` and `ReadOnlySpan<T>` are commonly encountered types in the FCL, perhaps constructed generic types deriving from these should also be in the cache.

src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs
```diff
@@ -333,17 +322,20 @@ public void CopyIn(LocalReference?[] byRefArguments)
...
-								InvocationMethods.SetArgumentValue,
-								new LiteralIntExpression(i),
-								NullExpression.Instance));
+							new AssignStatement(
+								new PointerReference(
+									new MethodInvocationExpression(
+										new MethodInvocationExpression(
+											ThisExpression.Instance,
+											InvocationMethods.GetArgumentValue,
+											new LiteralIntExpression(i)),
```
stakx, line 337: An alternative to calling `IInvocation.GetArgumentValue` would be to query `IInvocation.Arguments` once, and then access the returned arguments array directly for every by-ref-like parameter? That might be more efficient for methods having more than one by-ref-like-typed parameter.

src/Castle.Core/DynamicProxy/Generators/InvocationTypeGenerator.cs
```diff
@@ -361,25 +353,14 @@ public void CopyIn(LocalReference?[] byRefArguments)
...
-
-					// For now, we just erase it by substituting `null`:
-					method.CodeBuilder.AddStatement(new MethodInvocationExpression(
-						ThisExpression.Instance,
-						InvocationMethods.SetReturnValue,
-						NullExpression.Instance));
-				}
-				else
+				// TODO: For by-ref-like return values, we will need to read `IInvocation.ReturnValue`
+				// and set the return value via pointer indirection (`ByRefLikeProxy.GetPtr`).
```
stakx: Implement this "todo" for by-ref-like-typed return parameter support. And don't forget to also `Invalidate` the wrapper object for the return type.

src/Castle.Core/DynamicProxy/Internal/ByRefLikeProxy.cs
stakx: I am not certain whether (and how) one could get `[DebuggerTypeProxy]` to work for generic types. However, it occurred to me that instead of a `Get()` / `Set()` method pair, we could simply have a `ref`-returning `Value` property whose value could easily be displayed in the debugger, and also simplify usage.

src/Castle.Core/DynamicProxy/IByRefLikeProxy.cs
stakx: Renamed to `ByRefLikeReference`, `SpanReference`, and `ReadOnlySpanReference`, and got rid of the parallel interfaces.

src/Castle.Core/DynamicProxy/Internal/ByRefLikeProxy.cs
stakx: I indeed ended up with `ByRefLikeReference`.

PR #718, opened 2025-12-31
Restore ability on .NET 9 and later to save dynamic assemblies to disk

### LINKED ISSUE #697: Look into .NET 8's `PersistedAssemblyBuilder` to enable DynamicProxy tests on .NET Core
Not being able to save dynamically generated assemblies to disk on .NET Core has made it impossible to run most DynamicProxy tests on that runtime, meaning we had an increasingly large blink spot regarding new runtime & C# language features. The availability of something like `PersistedAssemblyBuilder` (if it is what it looks like) means we can finally close that gap and run all DP tests on .NET Core.

A prerequisite for this is enabling .NET 8 or later as a TFM. There's ongoing work towards that which should therefore be dealt with first.

stakx:
This PR adds a `PersistentProxyBuilder` implementation for .NET 9+ &ndash; similar, but not identical, to the one we already have for the .NET Framework. The minimum platform requirement for this new implementation is .NET 9 because it is based on SRE's `PersistedAssemblyBuilder`, which was introduced in that version of the runtime. Therefore I am also adding the `net9.0` TFM to the library.

I have an immediate need for this as an aid in improving by-ref-like types support further (#663), which is quite difficult when one cannot inspect the generated assemblies directly. While the proposed implementation should suffice for that specific purpose, it isn't thoroughly tested yet and it comes with certain caveats (as noted in its XML documentation comment). However, since it is an opt-in feature, it seems reasonably safe to me to publicly expose it.

What is left to be done at a later time is setting up IL verification during test execution on .NET 9 and later, even though that now seems to be within reach.

Closes #697.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Breaking Changes:
 
 Enhancements:
 - Minimally improved support for methods having `ref struct` parameter and return types, such as `Span<T>`: Intercepting such methods caused the runtime to throw `InvalidProgramException` and `NullReferenceException` due to forbidden conversions of `ref struct` values when transferring them into & out of `IInvocation` instances. To prevent these exceptions from being thrown, such values now get replaced with `null` in `IInvocation`, and with `default` values in return values and `out` arguments. When proceeding to a target, the target methods likewise receive such nullified values. (@stakx, #665)
+- Restore ability on .NET 9 and later to save dynamic assemblies to disk using `PersistentProxyBuilder` (@stakx, #718)
 - Dependencies were updated
 
 Bugfixes:


src/Castle.Core.Tests/DynamicProxy.Tests/PersistentProxyBuilderTestCase.cs
```diff
@@ -44,6 +53,107 @@ public void PersistentProxyBuilder_SavesSignedFile()
...
+				Assert.Null(twoAssembly.GetType(oneProxyType.FullName));
+				Assert.NotNull(twoAssembly.GetType(twoProxyType.FullName));
+			});
+		}
+
+		[Test]
+		public void TypeCacheWorks()
+		{
+			Assume.That(assemblyPaths, Is.Empty, "Test did not start with an empty assembly paths list.");
+
```
stakx: Duplicate logic. This has since moved into a `[SetUp]` method.

src/Castle.Core.Tests/DynamicProxy.Tests/PersistentProxyBuilderTestCase.cs
```diff
@@ -44,6 +53,107 @@ public void PersistentProxyBuilder_SavesSignedFile()
...
 
-#endif
\ No newline at end of file
+#elif NET9_0_OR_GREATER
+
+	[TestFixture]
+	[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
+	public class PersistentProxyBuilderTestCase
+	{
+		// The above fixture life cycle gives each test its own fresh fixture, such that each test.
```
stakx: `test.` &rarr; `test` (no full-stop)

src/Castle.Core/Castle.Core.csproj
stakx, line 1: Possibly extract the `net9.0` TFM addition into a separate PR. The `net9.0` TFM will be needed for by-ref-like support, too, so it might be nicer to introduce the TFM as a separate preliminary PR for both features, instead of being encased with this feature here.

src/Castle.Core/DynamicProxy/PersistentProxyBuilder.cs
```diff
@@ -12,12 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if FEATURE_ASSEMBLYBUILDER_SAVE
-
 #nullable enable
 
+using System.Diagnostics.CodeAnalysis;
+
```
stakx: * should be inside the `namespace` block
* likely superfluous anyway

src/Castle.Core.Tests/DynamicProxy.Tests/PersistentProxyBuilderTestCase.cs
```diff
@@ -44,6 +53,107 @@ public void PersistentProxyBuilder_SavesSignedFile()
...
+				File.Delete(assemblyPath);
+			}
+		}
+
+		[Test]
+		public void SavesOneAssemblyPerProxiedType()
+		{
+
+			var builder = new PersistentProxyBuilder();
+			builder.AssemblySaved += assemblyPaths.Add;
```
stakx: This (and the same code in the other test) could probably go into the `[SetUp]` method, too, since that one already has dealings with `assemblyPaths`.

src/Castle.Core/DynamicProxy/PersistentIsolatedTypesModuleScope.cs
```diff
@@ -54,6 +54,9 @@ public PersistentIsolatedTypesModuleScope(bool savePhysicalAssemblies = true,
 			usingStrongNamedModule = false;
 		}
 
+		// for testing purposes
+		internal event Action<string>? AssemblySaved;
```
stakx: I don't particularly like adding such stuff only for internal testing. Perhaps it would be worth thinking about doing this in a way that would make sense for public use, too?

PR #700, opened 2025-11-29
Drop AppVeyor in favor of GitHub Actions

Romfos:
Changes:
- Added release flow
- Drop appveyor
- Hotfix Explicit.NuGet.Versions
- Added Support Sourcelink and Deterministic Build (https://github.com/castleproject/Core/pull/669)
- Update some github actions steps versions

How to use:
image

If you want to publish directly to nuget **owner** of this repo should set NUGET_API_KEY to the repo secrets and check "publish to nuget" during release build

PR #722, opened 2026-01-02
Configure SourceLink, `.snupkg` symbols package format & update test projects

Romfos:
this PR is copy of #700 with removed appveyor changes

key points:

1. Test projects setup

- marks test projects as non-packable
- makes the tests in the Castle.Core.Tests.WeakNamed.csproj project runnable again (by adding the test framework & adapter  packages to the .csproj file)
- replaces NUnitLite with NUnit

1. Assembly output
- sets up SourceLink
- sets up deterministic builds

related [#700](https://github.com/castleproject/Core/pull/700#issuecomment-3700534209)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -86,4 +86,6 @@ for:
   # upload packages to AppVeyor
   artifacts:
   - path: build\*.nupkg
-    name: core
+    name: library
+  - path: build\*.snupkg
+    name: symbols


tools/Explicit.NuGet.Versions/Program.cs
```diff
@@ -39,6 +39,7 @@ private static void WriteNuspecToPackages(Dictionary<string, NuspecContentEntry>
 			{
 				using (var zipFile = ZipFile.Open(packageFile.Key, ZipArchiveMode.Update))
 				{
+					zipFile.GetEntry(packageFile.Value.EntryName).Delete();
```
stakx, line 42: I'm not sure I understand this change. What is the purpose of this change / why is it necessary?

buildscripts/common.props
```diff
@@ -24,14 +25,19 @@
 		<PackageIconUrl>http://www.castleproject.org/img/castle-logo.png</PackageIconUrl>
 		<PackageIcon>castle-logo.png</PackageIcon>
 		<PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
-		<PackageVersion>$(BuildVersion)</PackageVersion>
```
stakx, line 27: Can we really just remove `<PackageVersion>` without adding it back somewhere else?

src/Castle.Core/Castle.Core.csproj
```diff
@@ -8,8 +8,6 @@
 
 	<PropertyGroup>
 		<PackageId>Castle.Core</PackageId>
-		<GeneratePackageOnBuild>True</GeneratePackageOnBuild>
-		<PackageOutputPath>../../build/</PackageOutputPath>
```
stakx, line 12: When pushing a release to Nuget, AppVeyor expects the NuGet packages in the `build/` output directory. I suppose if you remove `<PackageOutputPath>` each project's packages will end up somewhere in the project's directory. If so, this change breaks NuGet publishing.

stakx (commented):
Not my final review, but a couple things I stumbled over.

buildscripts/common.props
Romfos: good catch. reverted untill  appveyor removing

src/Castle.Core/Castle.Core.csproj
Romfos: good catch. reverted untill appveyor removing

tools/Explicit.NuGet.Versions/Program.cs
Romfos: this is hotfix, without this nuget will have 2 files with the same name

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
```diff
@@ -7,21 +7,25 @@
...
 
 	<ItemGroup>
 	  <Compile Include="..\Castle.Core\DynamicProxy\Generators\Emitters\StrongNameUtil.cs" Link="DynamicProxy.Tests\StrongNameUtil.cs" />
 	</ItemGroup>
-
+	
+	<ItemGroup>
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
+		<PackageReference Include="NUnit3TestAdapter" Version="6.0.1" />
+	</ItemGroup>
```
stakx, line 27: While we're at it, I would suggest we add an explicit package reference for `NUnit` to this `<ItemGroup>`, instead of relying on it being available transitively (via the main test project reference below):

```diff
 <ItemGroup>
	<PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
	<PackageReference Include="NUnit3TestAdapter" Version="6.0.1" />
+	<PackageReference Include="NUnit" Version="3.14.0" />
 </ItemGroup>
```

buildscripts/common.props
```diff
@@ -32,6 +33,14 @@
 		<ContentTargetFolders>.</ContentTargetFolders>
 	</PropertyGroup>
 
+	<PropertyGroup Condition="'$(CI)' == 'True'">
+		<!--Deterministic Build and Source Link settings -->
+		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+		<EmbedUntrackedSources>true</EmbedUntrackedSources>
+		<IncludeSymbols>true</IncludeSymbols>
+		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
+	</PropertyGroup>
```
stakx, line 42: I'm noticing that on local builds, we still get (legacy?) `.symbols.nupkg` files (like before), while on CI, we get `.snupkg` files. I am not sure whether the former symbols package format makes particular sense for local builds... but I suspect not.

Perhaps it might be better to put both `<IncludeSymbols>` and `<SymbolPackageFormat>` in a non-conditional `<PropertyGroup>`, such that we always get the same symbols package format.

Or do you have a better recommendation?

buildscripts/common.props
```diff
@@ -32,6 +33,14 @@
 		<ContentTargetFolders>.</ContentTargetFolders>
 	</PropertyGroup>
 
+	<PropertyGroup Condition="'$(CI)' == 'True'">
+		<!--Deterministic Build and Source Link settings -->
+		<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+		<EmbedUntrackedSources>true</EmbedUntrackedSources>
+		<IncludeSymbols>true</IncludeSymbols>
+		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
stakx, line 41: I understand that you do not want to touch the AppVeyor configuration, but switching to `.snupkg` requires one small change over in `appveyor.yml`. At the very end of that file, please make the following change:

```diff
   # upload packages to AppVeyor
   artifacts:
   - path: build\*.nupkg
-    name: core
+    name: library
+  - path: build\*.snupkg
+    name: symbols
```

(This is so that the symbol packages are included in AppVeyor's NuGet package preview feed.)

stakx (changes requested):
This is looking mostly good, from what I can tell. Just a few details. After those have been taken care of, this should be ready for merging.

---

P.S. perhaps you could also add a brief description of your changes in the `CHANGELOG.md`? For exampe:

```diff
 Enhancements:
 ...
+Set up deterministic builds, SourceLink, and switched to `.snupkg` symbol packages (@Romfos, #722)
```

---

P.P.S: Are we sure that an explicit `<PackageReference Include="Microsoft.SourceLink.GitHub" Version="..." PrivateAssets="All" />` is no longer needed these days for SourceLink to work correctly? Has that since been integrated diretly into the .NET SDK? (https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects)

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
Romfos: I would recommend to not do it. From my point of view this is antipattern. I have following questions when I see this:
- Is it was made on purpose?
- What is the hidden logic behind this change?
- Can I remove it or not?

I think that adding such change you need to have a strong reason otherwise this is antipattern

buildscripts/common.props
Romfos: ok. added

buildscripts/common.props
Romfos: As for me - better solution is to implement it in a right way via `dotnet pack` and pass this `CI = true`

https://github.com/Romfos/Core/blob/47eceaf738b4de367a3469f5bc10f52e8eb82dc3/.github/workflows/release.yml#L29

some articles:
https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild

```
When set to true, this property enables settings that only apply to official builds as opposed to local builds on a developer machine. 
For example, stored file paths are normalized for official builds. 
But on a local development machine, the debugger isn't able to find local source files if file paths are normalized.
```

for now this change is `not in production` =)

src/Castle.Core.Tests.WeakNamed/Castle.Core.Tests.WeakNamed.csproj
stakx: How can it be an antipattern when a project references the dependency it needs? This is a project containing NUnit tests, why should it _not_ reference NUnit? I'm not sure I understand the reasoning here.

That said, I don't feel strongly about this. We can leave this unchanged for now, as long as things compile properly.

buildscripts/common.props
stakx: We're not quite there yet however.

Are you saying that non-CI builds should not emit any NuGet packages at all? In that case we'd still need to change _something_ here.

(Note that I wasn't suggesting taking the `<ContinuousIntegrationBuild>` out of the conditional group &ndash; that one clearly belongs there. I was referring only to the two last properties.)

Romfos: You can build nuget packages locally but probably you don't want to publish to nuget your local package symbols that could contains your local file paths, maybe other sensitive information

As I understand best practice here is to use it on CI

sorry, maybe I'm not filly understand this topic

PR #726, opened 2026-01-07
Accelerate DynamicProxy test execution

stakx:
Test execution has got much slower ever since I introduced a large number of test cases in `ParameterDefaultValuesTestCase`. The situation can easily be improved, since the slow-down is largely due to a higher number of `peverify.exe` invocations. Most of these aren't actually necessary, since it is the same proxy type getting generated for each test case.

We could either split the test fixture in two &ndash; with only one deriving from `BasePEVerifyTestCase` and getting the tests where PEVerify should be run &ndash; or we could come up with a way to skip `peverify.exe` for single tests. For now, I am opting for the latter solution.

I am aware that this is something of a quick fix. In the long run, it would probably make more sense to review all tests and decide which really need to be in a `BasePEVerifyTestCase`-derived fixture, and which ones don't really require any IL verification. Such a review will take much more time, I'm holding that off for the moment.

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/ParameterDefaultValuesTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/ParameterDefaultValuesTestCase.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2004-2021 Castle Project - http://www.castleproject.org/
+﻿// Copyright 2004-2026 Castle Project - http://www.castleproject.org/
 //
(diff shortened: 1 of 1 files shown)


stakx:
Interestingly, while these changes drastically reduce test execution time on my local machine (by nearly 1½ minutes), they hardly appear to affect our CI builds at all. So perhaps there's not much point going ahead with this, better to wait for the bigger "to `BasePEVerifyTestCase` or not to `BasePEVerifyTestCase`" review? I'll sleep on it.

PR #733, opened 2026-02-01
Preserve proxy type when cloning a record class proxy

### LINKED ISSUE #671: Support "with" for record proxies
Is it possible to have support for the `with` syntax for proxies?

I want to "enrich" my existing record with a mixin implementing a custom property, and make it persist across modifications of that record.

Unfortunately, it seems that when I use the `with` keyword, the proxy is gone and I end up with a completely new object.
At the same time, if I do the same thing manually, aka
<details>
    <summary>Example 1</summary>

```csharp
public record Base(int Value);

public interface IMixin
{
    string? MixedString { get; }
}

public record Inherited(int Value) : Base(Value), IMixin
{
    public string? MixedString { get; init; }
}

Base rec = new Inherited(1)
{
    MixedString = "Hello"
};
Console.WriteLine(rec);
var updated = rec with
{
    Value = 2
};
Console.WriteLine(updated);
```
</details>

Or even:
<details>
    <summary>Example 2</summary>

```csharp
public record Base(int Value);

public interface IMixin
{
    string? MixedString { get; }
}

public record Inherited(int Value) : Base(Value), IMixin
{
    string IMixin.MixedString => "Hello";
}

Base rec = new Inherited(1);
Console.WriteLine(rec);
Console.WriteLine((rec as IMixin).MixedString);
var updated = rec with
{
    Value = 2
};
Console.WriteLine(updated);
Console.WriteLine((updated as IMixin).MixedString);
```
</details>

The mixed in property is preserved when using `with` keyword, because the actual underlying class is preserved when using the `with` keyword.

Unfortunately, it seems not to be the case for proxies.

Is there a way to achieve something similar to that but completely in runtime with DynamicProxy?

stakx:
Closes #671.


stakx:
As this stands, this likely won't yet work correctly for class proxies with target. Not sure whether having record proxies with record targets is even a realistic / useful scenario, but it should perhaps be made to work nevertheless.

Issue #408, opened 2018-09-05
Future of Castle logging abstraction and adapters


jonorossi:
Extracted my comments from #404:

> The problem we've got right now is that we've discussed deprecating Castle's logging abstraction (which predates newer .NET logging, its from around 2004-2005) but haven't done anything about formalising that proposal to let people know our position and encourage users to use other libraries directly. We've briefly discussed how we could support a new logging facility for Windsor that isn't tied to the current Castle logging abstraction and would give users more freedom and not force all logging to a lowest common denominator interface (i.e. loosing structured logging when using Serilog). (https://github.com/castleproject/Core/pull/404#issuecomment-418238747)

> From my perspective, in Castle Project's prime (before 2010) users were building their applications with Windsor, MonoRail, ActiveRecord and other bits of the Castle stack so making use of the Castle logging abstraction wasn't a big thought as it just made sense; obviously most of these are now deprecated and other libraries have taken their place (i.e. ASP.NET MVC). We've been slimming down Windsor in the last year, and slowly deprecating bits dumped into what is today Castle Core. (https://github.com/castleproject/Core/pull/404#issuecomment-418246596)

> Oh, I didn't realise ASP.NET Boilerplate recommended using Castle's logging: https://aspnetboilerplate.com/Pages/Documents/Logging.

We need to work out a plan for how Castle libraries can log (e.g. DynamicProxy telling you about a cache miss), and how users of Windsor could possibly wire up an `ILogger`-like interface to become a new Windsor logging facility.

I'm using the term "abstraction" for Castle's `ILogger` and `ILoggerFactory`; "adapters" for the implementations of theses (e.g. log4net), and "facility" for the Windsor facility.

/cc @castleproject/committers

jonorossi:
I'm thinking a library like DynamicProxy doesn't need anything big like LibLog and we just add a hook to `ProxyGenerator` for it to report warnings, which I'm pretty sure is all it does.

stakx:
Originally posted the following regarding LibLog over in https://github.com/castleproject/Core/pull/404#issuecomment-418670071:

>  :+1: to @pi3k14 for mentioning LibLog. I originally discovered it via [Nicholas Blumhardt's blog article "Good citizenship - logging from .NET libraries"](https://nblumhardt.com/2017/07/library-logging/) and, without having it taken for a spin yet, some aspects of it make good sense to me:
>
[...]

PR #745, opened 2026-05-02
Ensure TraceLogger log level is maintained property

Youssef1313:
The added test would fail only under .NET Framework without the production change in this PR.

Under .NET Framework, if you execute the following:

```csharp
new TraceSource("Default", SourceLevels.Warning).Switch.Level
```

it will produce `-1` (corresponding to `SourceLevels.All`).

After `Initialize` gets called in `TraceLogger` constructor, the level is set as:

```csharp
Level = MapLoggerLevel(traceSource.Switch.Level);
```

which used to set it to Trace, even when Warning is explicitly requested.

The caching in `TraceLogger` generally looks wrong to me though. It caches TraceSource per Name, not taking into account the fact that multiple `TraceLogger` might be requested with the same name but different levels.

### DIFF
--- a/src/Castle.Core.Tests/DynamicProxy.Tests/LoggingTestCase.cs
+++ b/src/Castle.Core.Tests/DynamicProxy.Tests/LoggingTestCase.cs
@@ -109,6 +109,16 @@ namespace Castle.DynamicProxy.Tests
 				"Castle.DynamicProxy.Tests.LoggingTestCase+NonVirtualMethodClass because it cannot be intercepted."));
 		}
 
+		[Test]
+		public void LogsWarningOnlyByDefault()
+		{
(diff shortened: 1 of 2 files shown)


stakx:
To be honest, I'm not sure there's much sense in investing any more time and work in Castle logging facilities. Seems more likely to me that we're going to deprecate them soon (see the discussion in #408, plus Jono mentioned to me recently that Windsor hasn't seen much activity in recent times... so one fewer reason for keeping logging facilities going).

Youssef1313:
@stakx Any expected time when they will be deprecated? And when 6.x is expected to be released?

Meanwhile, this bug is causing perf issues for Moq under .NET Framework. I think the scope of the change is small enough and should be safe to get in until the deprecation happens?

